PR #31, opened 2019-11-13
Fix Dockerfile typo

danielhenrymantilla:


chfl4gs:
Thanks for the catch!

PR #34, opened 2019-11-16
Pr upstream 1

t14g0p:
Fix keystone pip3 install of the lib in the wrong dir. This caused an error when importing qiling on the docker container


chfl4gs:
Thanks for the PR.

PR #39, opened 2019-11-26
Set Up DllMain Args and Restructured import_symbols and import_address_table to Allow Ordinal Imports

jhumble:
Passed in dllmain args hinstDLL and fdwReason as described here: [https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain](https://docs.microsoft.com/en-us/windows/win32/dlls/dllmain) I did not include any handling for lpvReserved as in my experience it is rarely used. 

I also made some changes to support imports by ordinal and to avoid collisions where two different dlls have the same api name:

pe.import_symbols now maps addr to a dict with both name and ordinal. I updated the x86.py and x8664.py loaders to use this new structure in the api hooking. 

In pe.import_address_table I separated imports out by dll name (to avoid collisions) and added two mappings per import: api_name to addr and ordinal to addr. This allows for finding the correct address when loading a new import by name or by ordinal.  This structure also allows for easily applying hooks (by name) to imports by ordinal.

### DIFF
--- a/CREDITS.TXT
+++ b/CREDITS.TXT
@@ -23,6 +23,7 @@ Contributors (in no particular order)
 =====================================
 klks - The Code Refectorer, Advisor
 Kevin (chfl4gs) Foo - Travis, Dockerfile, Documentation and Website
+jhumble - Dll Ordinal Imports
 
 
 Alpha testers (in no particular order, named by github id)
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -47,14 +47,21 @@ class Process:
         # cache depends on address base
         fcache = path + ".%x.cache" % self.ql.DLL_LAST_ADDR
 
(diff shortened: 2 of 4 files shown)


xwings:
Hi,

Print out msg we will follow strace format, those additional information that required to print out you can use ql.drpint and only be viewed during output="debug"

i saw there are usage of globals()? we need to avoid that.

Can you please also add your name into CREDITS.TXT, Under "Contributors "

jhumble:
I made the changes you requested other than the globals() code.

That part already existed, and I'm not sure how I'd go about restructuring it. All I did was change the way it accesses the import since I changed the import_symbols structure a bit.

xwings:
my bad. global() was there.
Let me take a look. Thanks

Dliv3:
Thanks for the PR ! Merged! 👏

PR #44, opened 2019-12-04
hook for Sleep

re-fox:
Simple commit for hooking Sleep, in the case that there is malware that will check to make sure the Sleep command happened.  Uncomment the line to emulate the Windows Sleep in milliseconds.

### DIFF
--- a/qiling/os/windows/dlls/kernel32.py
+++ b/qiling/os/windows/dlls/kernel32.py
@@ -4,6 +4,7 @@
 # Built on top of Unicorn emulator (www.unicorn-engine.org) 
 
 import struct
+import time
 from qiling.os.windows.const import *
 from qiling.os.windows.fncc import *
 from qiling.os.windows.utils import *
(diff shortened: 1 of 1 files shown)


aquynh:
Cool, but we need a better way than this, as require to modify the code (uncommenting) is ugly.

re-fox:
I agree with you.  Not an optimal solution.

In my opinion, it should be something that should be enabled by default giving the user an option to disable/enable via configuration item or argument passed.  Do you have thoughts on qiling having some global configuration items?

aquynh:
Yes i would like to be able to optionally enable some features, like:

ql.function["sleep"] = true # false by default

Alternatively, we can also set our own customized functions, like:

ql.function["sleep"] = my_sleep

xwings:
We implemented ql.set_syscall and ql.set_winapi. I will merge this and close. Since sleep doesn't really needed in Qiling Framework.

xwings:
Thanks for the PR!

PR #45, opened 2019-12-04
Clipboard, TLS, GetLastError, and API Hook Work

jhumble:
Sorry up front for this pull request kind of being all over the place. I've been trying to write a basic unpacker and have been testing on one sample in particular (2583a5ed70dcec569898071601682277), and have just been fixing the things breaking emulation as I get further and further into it's execution. 

Many of these issues originated from code in Visual Studio's C Runtime startup code that is often compiled in and run prior to main, so it should hopefully help in getting a wide variety of programs emulating further.

**Thread Local  Storage**
Added hooks to allow for thread local storage interaction and built indexed storage into Qiling's Thread objects to actually support to storage/retrieval of data on a per thread basis. 

**GetLastError**
Most Windows APIs simply return success/fail as a return. Reason for failure is checked by calling GetLastError(). I implemented this hook and attached a variable under ql that api hooks can use to set error codes as needed.

**Clipboard**
Added a clipboard module and added it under the main ql obj in x86.py/x8664.py similar to how the registry is handled. It allows for simulation of the windows clipboard. Added hooks for the more common apis used to interact with the clipboard. 

**Various API hook additions and fixes**


Dliv3:
Great work! Could you add a sample file in examples/, to test this new code?

jhumble:
Not sure what's up with the conflict now. Is that because commit 6ea1cec2a867dceb19a0df9bad7797c479dd5d5b failed Travis?

Regardless, as of that last commit. Everything seems to be working. You can run each of the test binaries with windows_trace.py like this:
./examples$ python3 windows_trace.py ./rootfs/x8664_windows/bin/x8664_tls.exe

PR #47, opened 2019-12-05
Win clipboard tls hooks

jhumble:
Recreated due to conflicts I couldn't figure out how to fix in the previous pull request.

Sorry up front for this pull request kind of being all over the place. I've been trying to write a basic unpacker and have been testing on one sample in particular (2583a5ed70dcec569898071601682277), and have just been fixing the things breaking emulation as I get further and further into it's execution. 

Many of these issues originated from code in Visual Studio's C Runtime startup code that is often compiled in and run prior to main, so it should hopefully help in getting a wide variety of programs emulating further.

**Thread Local  Storage**
Added hooks to allow for thread local storage interaction and built indexed storage into Qiling's Thread objects to actually support to storage/retrieval of data on a per thread basis. 

**GetLastError**
Most Windows APIs simply return success/fail as a return. Reason for failure is checked by calling GetLastError(). I implemented this hook and attached a variable under ql that api hooks can use to set error codes as needed.

**Clipboard**
Added a clipboard module and added it under the main ql obj in x86.py/x8664.py similar to how the registry is handled. It allows for simulation of the windows clipboard. Added hooks for the more common apis used to interact with the clipboard. 

**Various API hook additions and fixes**

**Testing**
Added 3 test executables (2 x64, one 32-bit) that utilize the new functionality and hooks described above: thread local storage, clipboard, and getlasterror. You can run each of the test binaries with windows_trace.py like this:
./examples$ python3 windows_trace.py ./rootfs/x8664_windows/bin/x8664_tls.exe


jhumble:
I can't figure out why the Travis build is failing. Getting this error:
======================================================================
ERROR: test_elf_linux_arm64 (test_elf.ELFTest)
----------------------------------------------------------------------
  File "/home/jhumble/tools/forked_qiling/tests/test_elf.py", line 36, in test_elf_linux_arm64
    ql.run()
  File "/usr/lib/python3.7/site-packages/qiling-0.9-py3.7.egg/qiling/core.py", line 199, in run
    runner(self)
  File "/usr/lib/python3.7/site-packages/qiling-0.9-py3.7.egg/qiling/os/linux/arm64.py", line 113, in runner
    ql_arm64_enable_vfp(ql.uc)
  File "/usr/lib/python3.7/site-packages/qiling-0.9-py3.7.egg/qiling/os/linux/arm64.py", line 81, in ql_arm64_enable_vfp
    ARM64FP = uc.reg_read(UC_ARM64_REG_CPACR_EL1)
NameError: name 'UC_ARM64_REG_CPACR_EL1' is not defined

----------------------------------------------------------------------


But I haven't touched any of the Linux/ARM stuff at all.

qiling/os/windows/dlls/__init__.py
```diff
@@ -9,3 +9,4 @@
 from .wininet import *
 from .msi import *
 from .ntdll import *
+from .ucrtbased import *
```
Dliv3: It seems CI failed because there is no `ucrtbased` module

jhumble: That was it. Just forgot to push that file, which is why I wasn't seeing the error on my side. Sorry, haven't worked with Travis much and didn't see a way to dig into the results. 

Thanks!

Dliv3:
The two functions in `ucrtbased` have been implemented in `msvcrt.py`. Is there any problem using the implementation in `msvcrt.py`？

jhumble:
The samples I've been testing with imported those functions from ucrtbased rather than msvcrt. I didn't realize there were already (better) implementations in msvcrt. The two libraries should be pretty similar, so I just gutted ucrtbased and made it import the implementations from msvcrt.py [https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-library-features?view=vs-2019](https://docs.microsoft.com/en-us/cpp/c-runtime-library/crt-library-features?view=vs-2019) 

I did track down why I was having trouble getting the vfprintf working properly in my x64 test cases though. The get_params function in os/windows/utils.py was not taking into account the 32 byte home space reserved on the stack. I changed it to extract the stack parameters correctly. See:
[Stack Overflow Post](https://stackoverflow.com/questions/30190132/what-is-the-shadow-space-in-x64-assembly)
[MSDN Blog](https://blogs.msdn.microsoft.com/ntdebugging/2009/01/09/challenges-of-debugging-optimized-x64-code/)

Finally I added some error handling in the hook_WriteFile function to address cases where an invalid handle is passed in.

Dliv3:
Because CI does not include all windows test samples, in my tests, this PR will cause `crackme_x86_*.py` execution to fail. May be due to the `GetModuleHandleA`. Can you fix this?

Dliv3:
For example, in the output of `crackme_x86_windows_setcallback.py`, you can see that the execution result of `Easy_CrackMe.exe` is incorrect:
```
[+] Loading rootfs/x86_windows/bin/Easy_CrackMe.exe to 0x400000
[+] PE entry point at 0x401188
[+] Initiate stack address at 0xfffdd000
[+] TEB addr is 0x6000
[+] PEB addr is 0x6044
[+] Loading rootfs/x86_windows/dlls/user32.dll to 0x10000000
[+] Done with loading rootfs/x86_windows/dlls/user32.dll
[+] Loading rootfs/x86_windows/dlls/kernel32.dll to 0x100c9000
[+] Done with loading rootfs/x86_windows/dlls/kernel32.dll
[+] Done with loading rootfs/x86_windows/bin/Easy_CrackMe.exe
0x1010ba49: GetVersion() = 0x40004
0x10118074: HeapCreate(flOptions = 0x1, dwInitialSize = 0x1000, dwMaximumSize = 0x0) = 0x5017c30
0x10116e80: GetVersionExA(lpVersionInformation = 0xffffcee0) = 0x1
0x10116842: GetEnvironmentVariableA(lpName = "__MSVCRT_HEAP_SELECT", lpBuffer = 0xffffbd4c, nSize = 0x1090) = 0x0
```

jhumble:
You were right. Simple bug in hook_GetModudleHandleA. Should be better now. I checked, and both those crackmes complete properly now.

Dliv3:
Thanks for the PR ! Merged! 😁

PR #49, opened 2019-12-11
Hook error handling

jhumble:
Now emits two different types of errors. One in the case that the hook has not been defined/implemented and a separate error message along with a traceback in case an exception is raised during hook execution. 

This has made testing and debugging hook implementations a lot easier on my side.


xwings:
Hi,

Please look at the linux set.syscall, maybe you can also implement the same ? let user define their own API call without touching the core ?

xwings:
Hi,

I guess you need a new PR :)

xwings:
Hi,

We manage to find the problem that "kills the raise exception". So raise exception works now
This PR is just a work around, it will not stop the code and testing is still easy.

THanks for the idea and PR. it does really help us to notice the issue.

PR #50, opened 2019-12-11
Added fiber manager, supporting fiber apis and bunch of other api hooks

jhumble:
Added fiber manager, supporting fiber apis and bunch of other api hooks.


xwings:
Hi, 

As usual, do we have a test case ? maybe you can add it in test/test_pe_nu.py

jhumble:
This should all work fine, but it's not currently actually running the callback functions

As indicated in the big block comment in FiberManager.free(), I'm not sure how to actually have unicorn emulate the callback function. I can set up the callback parameters and return address, but as soon as the api hook for FlsFree finishes up and execution moves back to fncc, the wrapper code squashes any changes I've made to EIP/RIP by overwriting them with the return address from FlsFree.

Is there a clean way to actually accomplish this?

xwings:
Let the "PE Department" comment, @Dliv3

xwings:
@jhumble we still need some simple method to test the code. 

Once a wise man said, untested code is not a proven code. I quote from Dr. @aquynh

jhumble:
I added Fls.exe and a quick test (Similar to the others) in test_pe_nu.py. Did you want me to add something else also?

xwings:
Thanks !!!! we will test and merge ASAP.

Dliv3:
Thanks! Merged 😀

PR #54, opened 2019-12-16
add syscall _llseek

ucgJhe:
Try to emulate binary compiled from mipsel-linux-gnu-gcc in Ubuntu and complete the syscall.

P.S. basically it's a wrapper for lseek but for larger size.

### DIFF
--- a/qiling/os/linux/mips32el_syscall.py
+++ b/qiling/os/linux/mips32el_syscall.py
@@ -38,6 +38,7 @@ dict_mips32el_linux_syscall = {
     4116 : "ql_syscall_sysinfo",
     4122 : "ql_syscall_uname",
     4125 : "ql_syscall_mprotect",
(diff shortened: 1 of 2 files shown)


xwings:
Hi,

Please check the log and code seems to be broken.

Can you please fix it ?

ucgJhe:
yep sure.

xwings:
Hi,

a bit more minor change
white lines between functions is two rows not 3. python recomended way.

xwings:
hi, 

you also need to map the code to the arch you need.
if you test agiants mips, 

you need to ass the syscall to  qiling/os/linux/mips32el_syscall.py and match the syscall number

syscall can be refer from here:

https://w3challs.com/syscalls/?arch=mips_o32

xwings:
thanks !

Issue #53, opened 2019-12-15
fixed mem_read size in posix/syscall.py may cause UC_ERR_READ_UNMAPPED


ucgJhe:
There is a fixed mem_read size in posix/ql_syscall_open may cause UC_ERR_READ_UNMAPPED due to invalid memory access.

maybe try to shrink the size under certain conditions ?

xwings:
As far as i remember, we are following kernel. "TECHNICALLY" we are.

If we are not following kernel. (like how you describe the issue, i am sure we are not following now) We need to find a way to fix it.  Best way is to see how Linux kernel do it and we need to do the same.

xwings:
Just checked our code but i did not check the kernel code, yet.
We mapped at 0x100, which is 256 char. This is for a filename, not a complete full path character. If a folder + filename > 256 char. Which is totally possible, we are shit. very deep shit.

So, the bug you mention is valid. Do you have a solution for this ?

ucgJhe:
Can't compute the maximum size for mem_read cuz different architectures.
So I use a workaround that read byte-by-byte until null byte in my local repo, but it may slow the program down.

xwings:
Hi,

I had a discussion with @w1tcher  i guess that is the right solution.
Reading byte by byte till you hit nullbtye is the right way.

I do not think it will not huge impact perfomance. Really appriciate if you can perform a pull request.

ucgJhe:
yah, I will do.

xwings:
Thanks !

xwings:
Thanks for the PR and i will close this case.

PR #55, opened 2019-12-17
change fixed mem_read size to byte-by-byte mem_read

ucgJhe:
fix for #53 , try to stay pythonic and minimize performance impact here.


xwings:
hey,

thanks alot !

PR #57, opened 2019-12-17
PE loader now adds an entry for the main executable to LDR_DATA

jhumble:
I added a few lines of code to insert a LDR_MODULE entry for the main exe being run to match the structure on Windows. Prior to this fix, I was getting crashes on some packers/malware that were crawling the ldr data as a method of finding PE_IMAGE_BASE.

Also made a small fix to handle null parameters to SetClipboardData and removed a couple lines of debug code that made their way into windows_trace.py

### DIFF
--- a/qiling/os/windows/dlls/user32.py
+++ b/qiling/os/windows/dlls/user32.py
@@ -116,7 +116,7 @@ def hook_SetClipboardData(ql, address, params):
     try:
         data = bytes(params['hMem'], 'ascii')
     except:
-        data = ""
+        data = b""
     return ql.clipboard.set_data(params['uFormat'], data)
 
 #HANDLE GetClipboardData(
--- a/qiling/os/windows/dlls/user32.py
+++ b/qiling/os/windows/dlls/user32.py
@@ -113,7 +113,11 @@ def hook_CloseClipboard(ql, address, params):
     "hMem": STRING
 })
 def hook_SetClipboardData(ql, address, params):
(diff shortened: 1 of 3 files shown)


qiling/os/windows/dlls/user32.py
```diff
@@ -113,7 +113,11 @@ def hook_CloseClipboard(ql, address, params):
     "hMem": STRING
 })
 def hook_SetClipboardData(ql, address, params):
-    return ql.clipboard.set_data(params['uFormat'], bytes(params['hMem'], 'ascii'))
+    try:
+        data = bytes(params['hMem'], 'ascii')
```
cq674350529: The type of `data` is inconsistent. Maybe using `data=b""` instead?

qiling/os/windows/dlls/user32.py
```diff
@@ -113,7 +113,11 @@ def hook_CloseClipboard(ql, address, params):
...
+    except:
+        data = ""
```
cq674350529: The type of `data` is inconsistent. Maybe using `data=b""` instead?

xwings:
@jhumble if you agree, you can update accodringly ?

qiling/os/windows/dlls/user32.py
jhumble: You're right. I fixed it. Thank you.

Dliv3:
Thanks! Merged

PR #61, opened 2019-12-25
Ignore mode in open syscall if not O_CREAT

bambu:
Per both linux and macos man page, mode should only be used when the O_CREAT flag is passed. On macOS at least, the call to `ql_gfile.open()` will fail if the emulated program has data in the `mode` argument and it does not have the O_CREAT flag.

### DIFF
--- a/qiling/os/posix/filestruct.py
+++ b/qiling/os/posix/filestruct.py
@@ -16,6 +16,9 @@ class ql_file:
 
     @classmethod
     def open(self, open_path, open_flags, open_mode):
+        if (open_flags & os.O_CREAT) == 0:
+            open_mode = 0
+
         fd = os.open(open_path, open_flags, open_mode)
         return self(open_path, fd)
 
--- a/qiling/os/posix/syscall.py
+++ b/qiling/os/posix/syscall.py
@@ -232,9 +232,6 @@ def ql_syscall_open(ql, filename, flags, mode, null0, null1, null2):
                 mode = 0
 
             flags = open_flag_mapping(flags, ql)
-            if (flags & os.O_CREAT) == 0:
-                mode = 0
-
             ql.file_des[idx] = ql_file.open(real_path, flags, mode)
             regreturn = idx
         except:
@@ -267,9 +264,6 @@ def ql_syscall_openat(ql, openat_fd, openat_path, openat_flags, openat_mode, nul
             regreturn = -1
         else:
             openat_flags = open_flag_mapping(openat_flags, ql)
-            if (openat_flags & os.O_CREAT) == 0:
-                openat_mode = 0
-
             ql.file_des[idx] = ql_file.open(real_path, openat_flags, openat_mode)
             regreturn = (idx)
     ql.nprint("\nopenat(%d, %s, 0x%x, 0x%x) = %d" % (openat_fd, relative_path, openat_flags, openat_mode, regreturn))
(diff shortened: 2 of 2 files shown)


w1tcher:
Can you give a poc?

bambu:
Easiest PoC is just calling `ql_file.open` manually with the data that is being passed into it by `open` / `openat`.

```python
(Pdb) ql_file.open('/bin/ls', 0, 0x7ffff7ff8208)
*** OverflowError: signed integer is greater than maximum
```

Qiling PoC of shellcode that attempts to open "/bin/ls" as readonly and fails because rdx was populated previously with a stack pointer. This simulates the scenario that is  happening in a shared library when emulating a 64bit linux binary.

```python
from qiling import Qiling

"""
start:
    mov     rdx, 0x7ffff7fffff0 ; rdx has previously been set
    mov     rax, 0x2 ; open
    lea     rdi, [rel path] ; file to open
    mov     rsi, 0 ; O_RDONLY
    syscall
path:    db      /bin/ls, 0
"""
OPEN_ASM = "48bafffffff7ff7f000048c7c002000000488d3d0a00000048c7c6000000000f05cc"

asm = bytes.fromhex(OPEN_ASM) + b'/bin/ls\x00'
ql = Qiling(
    shellcoder=asm,
    rootfs="examples/rootfs/x8664_linux",
    archtype='x8664',
    ostype='linux'
)
ql.run()
```

Output:
```
open(/bin/ls, 0x0, 0x7ffff7ffffff) = -1
```

It may also be a cleaner solution to check for `O_CREAT` in `ql_file.open` instead of at `open` and `openat`

xwings:
maybe you can PR the cleaner solutions?

w1tcher:
I found the source code of os.open.
```
static PyObject *
posix_open(PyObject *self, PyObject *args)
{
    char *file = NULL;
    int flag;
    int mode = 0777;
    int fd;

#ifdef MS_WINDOWS
    Py_UNICODE *wpath;
    if (PyArg_ParseTuple(args, "ui|i:mkdir", &wpath, &flag, &mode)) {
        Py_BEGIN_ALLOW_THREADS
        fd = _wopen(wpath, flag, mode);
        Py_END_ALLOW_THREADS
        if (fd < 0)
            return posix_error();
        return PyInt_FromLong((long)fd);
    }
    /* Drop the argument parsing error as narrow strings
       are also valid. */
    PyErr_Clear();
#endif

    if (!PyArg_ParseTuple(args, "eti|i",
                          Py_FileSystemDefaultEncoding, &file,
                          &flag, &mode))
        return NULL;

    Py_BEGIN_ALLOW_THREADS
    fd = open(file, flag, mode);
    Py_END_ALLOW_THREADS
    if (fd < 0)
        return posix_error_with_allocated_filename(file);
    PyMem_Free(file);
    return PyInt_FromLong((long)fd);
}
```
I think the bug appeared when PyArg_ParseTuple, the program parsed mode as an int, but when passed in, it gave a value exceeding int, which caused OverflowError.

So I think when the bug appeared on a 64-bit system, a 64-bit register was used, and the upper 32 bits of the register may not be left blank. This will cause qiling to read as if it were read as an int64.

In O_CREAT, mode really doesn't make any sense, but I think this is something the kernel should judge, not what qiling should do.

bambu:
Thanks for taking the time to dig into the problem. I would agree that it does seem to be the cause of The problem. How about masking mode to be a signed 32 bit integer before passing it to os.open so that it does not cause the overflow error?

w1tcher:
I think we can make mode = mode & 0xffffffff when we get the value of mode.

bambu:
Sorry for the delay. 0xffffffff is unsigned and will cause it to throw the unsigned error. Closing this pull request and making another with the proper changes so as to not include the unrelated commit history.

PR #63, opened 2019-12-26
remove redundant ip converting

ucgJhe:
as title.

and does this confusing you guys ?

```
...
bind(3,:::8080,28) = -1
[+] syscall bind host: :: and port: 8080 sin_family: 10
socket(2, 1, 0) = 4
fcntl64(4, 2, 1) = 0
setsockopt
bind(4,0.0.0.0:8080,16) = 0
[+] syscall bind host: 0.0.0.0 and port: 8080 sin_family: 2
...
```
maybe ipv6 localhost should be printed in a more explicit way


xwings:
yes, i was debuggig some firmware and was thiking we should convert the bytes to a visible ip. so i put the dprint with conversion there. maybe u can add a v4 and v6 print there.

again, it will bind to localhost anyway. so it shold not make any issue, just for debug purposes.

ucgJhe:
maybe ::1 is better than :: ?

xwings:
sure and for sure. but :: is good enuf for v6 ? maybe add a place to indicate its v6 ?

ucgJhe:
use constant symbol like AF_INET, PF_INET6 instead of plain integers should be more clear and easy to debug.

xwings:
sure! can u do that

ucgJhe:
add socket_family, socket_types and socket_const for debug output

xwings:
for socket_type ,  i do remember mips is different from other. can you please verify ?

ucgJhe:
you are right, but seems separate the debug output from syscall is a better solution

xwings:
host = "::1" and host = IPv4 

bind is to localhost, but can we display the actual ipv6 address too ?

ucgJhe:
I am not quite sure what did you mean, ::1 is already an IPv6 ip address

xwings:
Hi,

what i am trying to say we did see cases port binding was to some funny ipv4 ip address. that is the reason why we had the ip conversion.

So, since we got a ipv6 now, can we have the same conversion such ash

host = ql_bin_to_ipv6(host)

ucgJhe:
actually the ipv4 conversion will cause a unsupported type error thats why i remove it

w1tcher:
Why do you want socket_protocol = socket_family_mapping (socket_protocol, ql.arch). I think it should be socket_domain here.

ucgJhe:
> Why do you want socket_protocol = socket_family_mapping (socket_protocol, ql.arch). I think it should be socket_domain here.

u r right, I misread the linux manual.

ucgJhe:
to be more specific 
https://github.com/qilingframework/qiling/blob/master/qiling/os/posix/syscall.py#L1384-L1403

host will be assigned to "::" when  sin_family == 10 is true and the error happen at line 1403

xwings:
Hi,

Happy New Year.

first, the the bind and debug print out should be different

`ql.file_des[bind_fd].bind(('::1', port))`

and

`ql.dprint ("[+] syscall bind host: %s and port: %i sin_family: %i" % (ql_bin_to_ipv4(host),port,sin_family ) )`

here,  syscall bind host: %s should show the IP (either v4 or v6) the program wanted to bind, but we still bind to localhost in emulation. Debug msg should show the original bind.

ucgJhe:
> Hi,
> 
> Happy New Year.
> 
> first, the the bind and debug print out should be different
> 
> `ql.file_des[bind_fd].bind(('::1', port))`
> 
> and
> 
> `ql.dprint ("[+] syscall bind host: %s and port: %i sin_family: %i" % (ql_bin_to_ipv4(host),port,sin_family ) )`
> 
> here, syscall bind host: %s should show the IP (either v4 or v6) the program wanted to bind, but we still bind to localhost in emulation. Debug msg should show the original bind.

and the conversion will fail when the program try to bind IPv6 , bcuz ql_bin_to_ipv4 is trying to compute integer from string type

xwings:
yes, that what i am referring to. 
we need a ql_bin_to_ipv4 or 

ql_bin_to_ip which will convert and check ipv4 and ipv6 automatically ?

ucgJhe:
I think the second one is better and let the library `ipaddress` do the job

PR #64, opened 2019-12-27
Implements 64-bit execve syscall

alfink:
On master, execve does not work with x64 yet. This patch fixes it for me


xwings:
will u be able to add a execve test into posix_syscall.c and also update Makefile for x64? so we can test it ?

xwings:
thanks and merged!

PR #66, opened 2020-01-03
Mask mode for os.open to avoid OverflowError

bambu:
History of the pull request is at #61. 

Avoid OverflowError for mode in os.open by masking mode with 0x7fffffff. Added test that checks failing condition.


xwings:
please add your name into CREDITS.TXT under contributor

xwings:
Hi,

you might wanna add test_posix.py to test_all.py

xwings:
thanks!

PR #74, opened 2020-01-08
Added missing fncc import. Fixed x64 param parsing

jhumble:
fncc.py was refactored recently and now the windows hook modules need to import both qiling.os.windows.fncc and qiling.os.fncc. I updated the dlls to import both as needed.

Additionally, I fixed _x8664_get_params_by_index to take into account the home space / shadow stack. I described this in a previous pull request: https://github.com/qilingframework/qiling/pull/47#issuecomment-562439953

### DIFF
--- a/qiling/os/windows/dlls/advapi32.py
+++ b/qiling/os/windows/dlls/advapi32.py
@@ -4,6 +4,7 @@
 # Built on top of Unicorn emulator (www.unicorn-engine.org) 
 import struct
 from qiling.os.windows.fncc import *
+from qiling.os.fncc import *
 from qiling.os.windows.utils import *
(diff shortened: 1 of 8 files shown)


jhumble:
I also included a test file that shows the fix. If you run python3 windows_trace.py rootfs/x8664_windows/bin/CreateFile.exe. Without the changes to x8664_get_params_by_index, the parameters are messed up. With this fix, they are all correct.

jhumble:
Also fixed some utf-16le string conversion bugs in some of the widestring versions of a couple windows api hooks.

xwings:
Hey Buddy,

Thanks !

PR #77, opened 2020-01-13
add test for syscall read, write in arch x86_linux

ucgJhe:
add test for syscall read, write in arch x86_linux.


xwings:
thanks !

PR #79, opened 2020-01-14
use log_dir instead of log_file

ucgJhe:
use log_dir instead of log_file in syscall vfork


xwings:
we got a issue talking about making these log using python logging, do you think you can help ?

ucgJhe:
sure  , how about merge this first and I will do it with another PR.

xwings:
Thanks !

Issue #30, opened 2019-11-01
Use python's logging module


klks:
We should move to python's logging module. the module is thread-safe and offers a number of features including log rotation and custom formatters.

xwings:
@w1tcher is taking over this case.

xwings:
i think we are more or less done with this

PR #80, opened 2020-01-18
try to use logging library for better logging

ucgJhe:
tested it on guest arch. x8664_linux, x86_linux (multithread included), mips32el on x8664_linux host, and partially rewrote print string to f-string.

may need more testing for other host arch.


xwings:
can you please check, still some bug need to be fix. as per travis

xwings:
Hi,

We can have make one of the test  in tests/test_elf with logs output. same goes with test_pe_nu

just pick and enable log_dir and log_split

ucgJhe:
from here use logging library for both stdout and logging to disk file.

ucgJhe:
@xwings gonna need more details for json and xml output (what should be printed, fields and value format etc.)to implement it.

xwings:
json and xml will be another feature. we will make it another PR.

xwings:
Thanks!

We will discuss in the Issue #30

PR #81, opened 2020-01-21
fix dprint for bytes message

ucgJhe:
handle bytes message in dprint

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -232,7 +232,9 @@ class Qiling:
         if self.output == QL_OUT_DEBUG:
             self.log_file_fd.debug(*args, **kw)
         elif self.output == QL_OUT_DUMP:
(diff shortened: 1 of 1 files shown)


aquynh:
This looks ugly. Can we have self.debug() instead?

ucgJhe:
sure, but how to tell which is nprint and dprint from context ?

xwings:
@ucgJhe self.debug == QL_OUT_DUMP and QL_OUT_DEBUG is dprint, as usual.

nprint is normal print without debug or dump.

ucgJhe:
you mean use self.debug to specify the level instead of self.output ?

I though he wants a self.debug function to do all the print out job

xwings:
@aquynh what do you think.

xwings:
i will take in this commit and we will plan in the next commit

PR #92, opened 2020-02-06
fix 'unhexlify' is not defined when use ql_asm2bytes

kabeor:
the function compile_instructions in ql_asm2bytes used unhexlify, but it haven‘t import unhexlify from binascii.

This will cause the following problems:
NameError: name 'unhexlify' is not defined


qiling/os/utils.py
```diff
@@ -25,6 +25,8 @@
 from qiling.exception import *
 from qiling.utils import *
 
+from binascii import unhexlify
+
```
aquynh: this blank line can be removed.

kabeor: finished

xwings:
Thanks !

PR #103, opened 2020-02-20
fix gdbserver issue

kabeor:


xwings:
update the update!

qiling/gdbserver/qldbg.py
```diff
@@ -77,7 +77,7 @@ def resume_emu(self, address=None, skip_bp=0):
 
         self.skip_bp_count = skip_bp
         if self.exit_point is not None:
-            self.ql.dprint('emu restart at:  ', hex(self.current_address))
+            self.ql.dprint('[+] emu restart at: 0x%x' % self.current_address)
```
cq674350529: `%#x` does the same like `0x%x`, just a hint :)

PR #107, opened 2020-02-25
add gdbserver write reg function

kabeor:


xwings:
now we left only multi arch

xwings:
thanks

PR #119, opened 2020-03-03
add gdbserver-arm support

kabeor:


aquynh:
great job!

however, maybe it is time to refactor gdblistener.py so that we do not mix all architecture dependent code into the same class, with so many if/then. maybe each architecture has 1 class, then call class method via dispatcher?

xwings:
thanks! merge for now and pending refactoring

PR #121, opened 2020-03-03
Use password-protected archive to protect shellcode.

quangnh89:
Due to antivirus softwares scan this shellcode and detect it as malware, we should use create password-protected archive to protect it.


xwings:
cool, can you do the same for wannacry  example too ? Thanks!

quangnh89:
> cool, can you do the same for wannacry example too ? Thanks!

No, I can't. After reviewing core component, I realize that that example loads PE file by using path. My example uses qiling to emulate shellcode from a buffer.

Qiling should accept a file-like object as parameter to load file.

xwings:
Ah, ok just realized that if for a string. 

thanks !

PR #124, opened 2020-03-06
add gdbserver-mips32 support

kabeor:


xwings:
Hi,

i think all the 

`if self.ql.arch == QL_MIPS32EL: ` 

can be

`if self.ql.arch in (QL_MIPS32EL, QL_MIPS32): `

can you please change it

xwings:
Thanks !!!!!

PR #125, opened 2020-03-06
fix for syscall nanosleep read size

ucgJhe:
according to the Linux manual the size of time_t is 8 bytes on x86_64 and 4 bytes on x86.


xwings:
thanks !!!!

PR #131, opened 2020-03-10
reveal open_flag_mapping and mode in debug message

ucgJhe:
now the syscall open and openat will look like `open(filename, O_CERAT | O_WRONLY, 0o644)` in debug meesage


xwings:
Hi,

Thanks !

PR #132, opened 2020-03-10
fix missing arch for open_flag_mapping, remove uneeded argument

ucgJhe:
as title.

### DIFF
--- a/qiling/os/posix/syscall.py
+++ b/qiling/os/posix/syscall.py
@@ -257,7 +257,7 @@ def ql_syscall_open(ql, filename, flags, mode, null0, null1, null2):
             regreturn = -1
 
     ql.nprint("open(%s, 0x%x, 0o%o) = %d" % (relative_path, flags, mode, regreturn))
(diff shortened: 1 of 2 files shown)


qiling/os/posix/syscall.py
```diff
@@ -257,7 +257,7 @@ def ql_syscall_open(ql, filename, flags, mode, null0, null1, null2):
             regreturn = -1
 
     ql.nprint("open(%s, 0x%x, 0o%o) = %d" % (relative_path, flags, mode, regreturn))
-    ql.dprint("open(%s, %s, 0o%o) = %d" % (relative_path, open_flags_mapping(flags), mode, regreturn))
+    ql.dprint("open(%s, %s, 0o%o) = %d" % (relative_path, open_flags_mapping(flags, ql.arch), mode, regreturn))
```
xwings: make it as drpint and [+] open so we know is a debug print.

ucgJhe: fix it for both syscall open, openat

xwings:
thanks !

PR #133, opened 2020-03-10
add syscall ql_syscall_lstat

ucgJhe:
as title.


xwings:
hi,

maybe just add it in for other arch too ?

xwings:
nice, thanks !

PR #142, opened 2020-03-13
Implementation for sample c43f1d61ccdc3ed32ba84d4aa6a47b34

0ssigeno:
I'm trying to fix every hook that is needed to run the sample c43f1d61ccdc3ed32ba84d4aa6a47b34.
Hope everything is good to go.


Dliv3:
It seems that the windows api in synchapi.py should belong to kernel32.dll, is it so?

There is the document of `EnterCriticalSection`, showing that the dll where `EnterCriticalSection` is located is kernel32.dll
https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-entercriticalsection

Dliv3:
By the way, can you provide the exe file of c43f1d61ccdc3ed32ba84d4aa6a47b34 ?  I want to do some tests

0ssigeno:
I splitted the kernel32 because, the class will probably explode in number of lines. I thought was better to subdivide the functions based on their header file. But if you prefer to have everything inside kernel32, i will update the PR.

Is possible to find the sample on [virustotal](https://www.virustotal.com/gui/file/7561a8acb46bbf3e5989b10f4d6c728e005830b24bc4720c77d98ef3e088509a/detection), but if you can not download it from there, here we go [DHL_FORM7PDF.zip](https://github.com/qilingframework/qiling/files/4332391/DHL_FORM7PDF.zip)

xwings:
hi,

my suggestion would be, if we split, we gonna 

1. split it now, split part and not splitting  another half is not healthy
2. if we need to split we gonna make it directory format, example
    - kernel32/abc.py
     kernel32/def.py

for now my suggestion is, we merge into one and we can start splitting in another PR ? and also, please add your name into credits.txt

0ssigeno:
Agree with that. I will add everything inside kernel32.py, I let you guys decide which is better, if let everything inside kernel32 or not.

xwings:
Thanks buddy, let @Dliv3 run a quick check and merge shall happen.

again, welcome to the group !

xwings:
Hi,

for lpMultiByteStr maybe we should print it in hex format so its easier for user to verify the content

Dliv3:
The wide char in windows is UTF-16le encoding, I don't think it should be modified to latin1 in hook_WideCharToMultiByte and hook_MultiByteToWideChar

0ssigeno:
You are right, the sample seems to work fine with utf-16le inside the 2 hooks. The problem was only inside read_wstring and read_cstring.
Agree on the ugly output for not ascii values, will try to change it (I'm still new to the qiling architecture, may take a while)

xwings:
Thanks and stay safe !

PR #146, opened 2020-03-17
display prot and flags constant under debug mode

ucgJhe:
as title.


xwings:
Nice, this is impressive !

PR #147, opened 2020-03-17
add syscall old_mmap for x86_32

ucgJhe:
according to the linux kernel this is only for ia32 compatibility.

### DIFF
--- a/qiling/os/linux/x86_syscall.py
+++ b/qiling/os/linux/x86_syscall.py
@@ -17,6 +17,7 @@ dict_x86_linux_syscall = {
     0x2D : "ql_syscall_brk",
     0x30 : "ql_syscall_signal",
     0x55 : "ql_syscall_readlink",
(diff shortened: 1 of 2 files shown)


xwings:
if i not mistaken old_mmap works the same with mmap. maybe we just stick to mmap ?

ucgJhe:
same implement but different interface, only happend when you run x86_32 binary calling syscall mmap directly from int 0x80 on arch x86_64.

xwings:
i thin mips32 big endian also using old_mmap. can you check and comfirm ?  if not i mistaken  i mapped it to mmap. can you please also change it. 

also the old_mmap ql.nprint please change to old_mmap instate to the current mmap,

xwings:
btw, i do keep a copy of mips32 EB binary in example/rootfs/mips32eb

ucgJhe:
after test both mips32eb and mips32el use mmap.

xwings:
cool. i will merge now

PR #148, opened 2020-03-17
Registry

0ssigeno:
Fixed some hooks and updated functions to retrieve registry.
Samples sent the string, for example "P\x00a\x00n\x00e\xxl" instead of "Panel" as a subkey, meaning that every compare would fail.


0ssigeno:
I fucked up with the sync of the fork, if there are problem in this aspect, please let me know and I will open another PR.

xwings:
the diff seems to be normal. let me try to merge

PR #149, opened 2020-03-17
display mprotect prot constant under debug mode

ucgJhe:
display mprotect prot constant under debug mode and some clean up.


xwings:
thanks !

PR #154, opened 2020-03-18
Hooks and split

0ssigeno:
Splitted kernel32.
Added more hooks.
I swear I will provide a test once a sample works entirely.


xwings:
Hi,

for all the examples, we leave the "import qiling" after the sys.path.append("..") is due to we can run the example without install.

I guess we should keep it there.

0ssigeno:
Reverted as asked!

xwings:
i think Quynh just updated the pe.py. can you fix it ?

0ssigeno:
Should be good on the conflict now

xwings:
i guess we need some time to review it. so far so good ! but still allow us to take some time

0ssigeno:
Let me know if there are any problems

examples/adcache_x86_windows_debug.py
```diff
@@ -4,15 +4,15 @@
 # Built on top of Unicorn emulator (www.unicorn-engine.org)
 
 import sys
-from zipfile import ZipFile
-
 sys.path.append("..")
+from zipfile import ZipFile
```
xwings: this should be on top of sys.path.append ?

examples/hello_mips32el_linux_debug.py
```diff
@@ -4,8 +4,8 @@
 # Built on top of Unicorn emulator (www.unicorn-engine.org)
 
 import sys, string, random
-sys.path.append("..")
 from qiling import *
+sys.path.append("..")
```
xwings: this should be ontop of from qiling import and not below

examples/hello_arm_linux_debug.py
```diff
@@ -3,8 +3,8 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 # Built on top of Unicorn emulator (www.unicorn-engine.org) 
 import sys
-sys.path.append("..")
 from qiling import *
+sys.path.append("..")
```
xwings: thiis one too

examples/hello_arm_linux_custom_syscall.py
```diff
@@ -4,8 +4,8 @@
 # Built on top of Unicorn emulator (www.unicorn-engine.org)
 
 import sys
-sys.path.append("..")
 from qiling import *
+sys.path.append("..")
```
xwings: same for append should be on to of qiling import

xwings:
Thankyou sir !

PR #155, opened 2020-03-19
fix mprotect logic and munmap logic

ucgJhe:
the behavior of mem_unmap from unicorn engine seems different from x86 syscall munmap, gonna need more testing


xwings:
please also fix 

```
qiling/qiling/os/posix/constant_mapping.py", line 83, in socket_type_mapping
    }.get(arch)(t)
TypeError: 'dict' object is not callable
```

PR #162, opened 2020-03-19
Hooks

0ssigeno:
Still having an error with sample DHL_FORMAT.
Another sample that I'm testing is 
[emotet_1.zip](https://github.com/qilingframework/qiling/files/4355919/emotet_1.zip), but having problem with a real implementation of (Find/Load)Resource.
Help is welcomed


xwings:
I will create an issue, thanks for the PR

PR #168, opened 2020-03-23
add function tell for class ql_file and allowed multiple mapping constant for socket type

ucgJhe:
add SOCK_CLOEXE, SOCK_NONBLOCK for linux socket type, allow multiple mapping for socket type, function tell for class ql_file


xwings:
thank you sir!

Issue #165, opened 2020-03-20
syscall number = 0x4e(78) not implement


bannsec:
When attempting to use qiling to run `/bin/ls` on my default bionic install, i get a notice that syscall 78 is not implemented, then qiling immediately hangs until i manually kill it.

xwings:
Hi,

This is not a bug, this is due to the /bin/ls needs a syscall which not yet implemented by us.

for example, syscall 78 or 0x4e in X86_64 should be pointing to getdent , according to https://syscalls.w3challs.com/?arch=x86_64

would you be abe to do a Pull Request ?

xwings:
Thanks @ucgJhe !
I will close this issue for now

PR #169, opened 2020-03-23
add syscall getdents for both linux x86 and x86_64

ucgJhe:
implement syscall getdents for #165 , but command "ls" still not working its more than just a simple syscall.

### DIFF
--- a/qiling/os/linux/x8664_syscall.py
+++ b/qiling/os/linux/x8664_syscall.py
@@ -33,6 +33,7 @@ dict_x8664_linux_syscall = {
     0x3C : "ql_syscall_exit",
     0x3F : "ql_syscall_uname",
     0x48 : "ql_syscall_fcntl",
+    0x4E : "ql_syscall_getdents",
     0x4F : "ql_syscall_getcwd",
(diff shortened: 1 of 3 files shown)


xwings:
NIce,

Can you add this to other arch as well ?

also. If we have these syscall in posix/syscall.py can we add it in into x8664 ?

```
[!] 0x7ffff488beb3: syscall number = 0xda(218) not implement
[!] 0x7ffff488bf15: syscall number = 0x111(273) not implement
[!] 0x7ffff488bff1: syscall number = 0xe(14) not implement
[!] 0x7ffff4134f9e: syscall number = 0x12e(302) not implement
[!] 0x7ffff412e975: syscall number = 0x89(137) not implement
[!] 0x7ffff412e975: syscall number = 0x89(137) not implement
```

if we dont have it, that is fine.

xwings:
Can you try to run /bin/ls again ? Seems like having issue now

ucgJhe:
after disabled syscall sigprocmask its working again

ucgJhe:
turns out its a typo issue , should be good now

xwings:
Nice work! Thanks !

PR #170, opened 2020-03-24
Batch sample analysis

0ssigeno:


0ssigeno:
This simple [bootkit](https://app.any.run/tasks/096b98d3-7214-47e0-971d-7f817be6bdc3/) seems to work fine now

xwings:
Hi,

Thanks. I do understand we got more fix to go. After all these cleanup, maybe  you can split PR according to PR "type" :)

qiling/os/windows/dlls/kernel32/tlhelp32.py
```diff
@@ -0,0 +1,69 @@
...
+#   LPPROCESSENTRY32W lppe
+# );
+@winapi(cc=STDCALL, params={
+    "hSnapshot": HANDLE,
+    "lppe": POINTER
+
+})
+def hook_Process32NextW(ql, address, params):
+    # Return True if more process, 0 else
+    global COUNTER_PROCESS32_NEXT
```
xwings: try to avoid global. we gonna bump into issue sooner or later.

qiling/os/windows/dlls/kernel32/fileapi.py
```diff
@@ -205,3 +217,19 @@ def hook_CreateFileA(ql, address, params):
...
+#   DWORD  nBufferLength,
+#   LPWSTR lpBuffer
+# );
+@winapi(cc=STDCALL, params={
+    "nBufferLength": DWORD,
+    "lpBuffer": POINTER
+})
+def hook_GetTempPathW(ql, address, params):
+    # TODO not sure if the string should end with \ and have a 00
+    temp = "C:\TEMP"
```
xwings: Will it be better C:\Windows\Temp ?

xwings: it should also create rootfs\Windows\Temp ?

qiling/os/windows/dlls/kernel32/tlhelp32.py
0ssigeno: Understood, but I thought that was a bad practice insert it inside the class Qiling.
Suggestions on how to implement this concept?

qiling/os/windows/dlls/kernel32/fileapi.py
0ssigeno: You are right, it should!

qiling/os/windows/dlls/kernel32/tlhelp32.py
xwings: qiling class could be better. depends on how ofen u use it. or some class among there few windows API

xwings: COUNTER_PROCESS32_NEXT use use in this function ? why do we need a global ?

0ssigeno: Meh, for now i think we can always return 1 safely.
I will just set a TODO to remember to change the function if we want a more complex implementation, but in this case we must implement correctly a snapshot handle (and the function createToolHelp32Snapshot) that i thought was not necessary.
If you agree with the simpler solution, I will change the functions and push, otherwise if we want to implement the handle, there is a lot of work to do.

xwings: Should be fine. My only issue is the global var. maybe you can move in the class.

xwings:
can you also provice this bookit example, and zip it with password infected. Add into travis too :)

qiling/os/windows/dlls/kernel32/tlhelp32.py
0ssigeno: The problem is that is only used in this function. If we do this with every variable that every hooks need, the class Qiling will explode with variables that have to use if not for a single hook. I don't think is a good practice either.

xwings: if only this function so we dont need a global ?

0ssigeno: I mean yeah, but there is another way to implement a counter of how many times a function has been called, without use a global?

xwings: we need to find a way or else if we decided to move to unittest, this global will screw things up. maybe we can implement a counter function with ql.counter['fuction_name'] ?

0ssigeno:
Have to remove the last push, my script pushed in the wrong fucking branch, sorry.

qiling/os/windows/dlls/kernel32/tlhelp32.py
0ssigeno: I think this is a better way to implement it. Let me know how you prefer, if you want to think about it, for now I can just return 0 .

xwings: make a function, i got a feeling we might need that for future for sure.

0ssigeno: 296eca3 should resolve it

xwings:
Hi,

Looks cool. Since this is Windows specific counter. We can rename it to winapi_counter. 

Also, we should have some print out with ql.dprint to print the count of how many times this api been called.

Will this be good for malware analysis ?

0ssigeno:
It will probably help if you want to mutate the behaviour of a syscall. An example that i have in mind is a sample that searches for other processes. If we always return "No, is not there", the sample will probably quit because it undestands to be in a sandbox.
About the ql.dprint(), how you think we need to procede? Print it when the function is called, after the parameters?

xwings:
agree, but functions call still a very weird name to me. maybe call it syscall_count those windows API are syscall according to microsoft documents.

ql.dprint should come in after the execution. once the process end, ql.drpint will come in and print it as a [list]. so malware analyst can collect it if needed. what do you think

0ssigeno:
Yeah, the name can be or sycall_count or winapi_counter, your call.
Nice idea about the print after the execution! Will implement it now

xwings:
syscall_count will be cool. we got a set_syscall, so now we got a syscall_count.

qiling/core.py
```diff
@@ -235,6 +239,9 @@ def run(self):
         runner = self.build_os_execution("runner")
         runner(self)
 
+        # Print syscalls and their call numbers.
+        self.dprint("Syscalls and number of invocations")
+        self.dprint(list(self.PE.syscall_count.items()))
```
xwings: I guess we need to move this print to os/windows/x86.py and os/windows/x8664.py or check for the var exits before print

0ssigeno: Yep, my bad!

xwings:
what is in winbase.py, those api belongs to the base ? 

if yes, can we just name it to just base.py

0ssigeno:
The header is called "winbase.h", i don't think si correct to change the name of the file, since all others have the corresponding header file name.
(https://docs.microsoft.com/en-us/windows/win32/api/winbase/)[https://docs.microsoft.com/en-us/windows/win32/api/winbase/]

xwings:
Ok, i agree. my bad.

qiling/os/windows/x86.py
```diff
@@ -175,3 +175,6 @@ def runner(ql):
 
     if ql.internal_exception is not None:
         raise ql.internal_exception
+
+    ql.dprint("Syscalls and number of invocations")
```
xwings: ql.print with prefix for the msg, [+] and the list with [-] so we know its a subset

xwings: maybe drpint move before the exception if else

qiling/os/windows/x8664.py
```diff
@@ -172,5 +172,5 @@ def runner(ql):
     if ql.internal_exception is not None:
         raise ql.internal_exception
 
-    ql.dprint("Syscalls and number of invocations")
-    ql.dprint(list(ql.PE.syscall_count.items()))
+    ql.dprint("[+] Syscalls and number of invocations")
```
xwings: sorry, my bad again, i guess we move to os/windows/utils.py since these are the same code and we dont habe to call twice.

0ssigeno: Ok, understood what you mean

xwings: call it post_report some something, something related to post process report and it should location before line 172, i think. looks cleaner

xwings:
hi, 

the reason i suggest to name it as post_report or something related. if we got more print out which similar to syscall_count print tout in the future we can just add those print out in that function.

0ssigeno:
I didnt' see the at all your comment for the name of the function, sorry.

xwings:
thanks alot and thanks for accomodating all the addtional request.

PR #176, opened 2020-03-27
GandCrab PR

0ssigeno:
Anyrun of sample: https://app.any.run/tasks/48406845-690f-47ee-b385-38534d75d00a/

What Qiling do:
Pass checks for OSVersion (is possible to change value inside variables.py)
Pass check for Administrator Privilege (variable inside SID)
Pass check for Keyboard Layout (depends of your NTUSER.DAT)
Pass check for System/User Default Language (needs a better solution here tbh)
Crashes on call SHGetSpecialFolderPathW, probably because my DLL doesn't have this function since its deprecated.


xwings:
Hi,

shellcode might behave differently. So, you might want to do a ql.shellcoder to check if that is a shellcode and do a different process.

docs/API.md
```diff
@@ -44,7 +44,7 @@ ql.nprint will not print anything when output="off"
 
 
 ```
-ql.dprint(0,"")
+ql.dprint("")
```
xwings: all the ql.drpint must come with a level now

qiling/os/windows/dlls/kernel32/tlhelp32.py
```diff
@@ -28,7 +28,7 @@ def hook_CreateToolhelp32Snapshot(ql, address, params):
     # TODO thinking about implementing an handler, gonna see if is really necessary
     flag = params["dwFlags"]
     if flag == TH32CS_SNAPPROCESS:
-        ql.dprint(0,"[.] We probably found a Ransomware!")
+       ql.dprint(0, "[!] We probably found a Ransomware!")
```
xwings: this could be verbose 1 or more now

qiling/os/windows/variables.py
```diff
@@ -0,0 +1,22 @@
+from qiling.os.windows.const import *
+
```
xwings: this file should provide default value, but if user defined value should be in rootfs or "c:\" in the sandbox. if user we might want to all a ql.windows_var = filename.py  to allow user to add in the config without changing the core

qiling/os/windows/variables.py
```diff
@@ -0,0 +1,22 @@
...
+                   VER_SERVICEPACKMAJOR: 0,
+                   VER_SERVICEPACKMINOR: 0xADD,
+                   VER_SUITENAME: 0xADD,
+                   VER_PRODUCT_TYPE: 0xADD,
+                   "wReserved": 0xADD,
+                   "OS": "1001"
+                   }
+
+# Environment
+Environment = {"windir": "C:\\Windows",
```
xwings: sounds like a malware sandbox IDable name, but again if we reach malware can ID us, we are very famous by then

qiling/os/windows/dlls/msvcrt.py
```diff
@@ -185,43 +185,6 @@ def hook__get_initial_narrow_environment(ql, address, params):
     return ret
 
 
-def printf(ql, address, fmt, params_addr, name):
```
xwings: you might wanna keep this part for if ql.shellcoder == True and normal Windows goes to your version

qiling/os/windows/dlls/kernel32/winnls.py
```diff
@@ -158,3 +158,25 @@ def hook_LCMapStringA(ql, address, params):
...
+    return lang
+
+
+# LANGID GetSystemDefaultUILanguage();
+@winapi(cc=STDCALL, params={
+})
+def hook_GetSystemDefaultUILanguage(ql, address, params):
+    # TODO find better documentation
+    # https://docs.microsoft.com/it-it/windows/win32/intl/language-identifiers
+    ql.dprint(0, "[!] Sample is checking system language!")
```
xwings: my suggestion is you might want to choose different symbols for this dprint, example [$] or [=] just a example. the reason is

1. easier to capture for malware analysis. 
2. if we gonna make a sandbox based on qiling, we just need to capture a specific symbols
3.  [!] we keep it for error, or raise

qiling/os/windows/dlls/kernel32/tlhelp32.py
xwings: this is maybe a "decision-able"  statement, so a brand new symbol [.] or something can be fine. why a new symbol, i do mention in another comment

qiling/os/windows/dlls/kernel32/winnls.py
0ssigeno: Agree on the new symbol, will change it.
Agree on the level, forgot to change it

qiling/os/windows/dlls/msvcrt.py
0ssigeno: I don't understand why, the function was moved to utils since, it should not create a problem.

qiling/os/windows/variables.py
0ssigeno: Yeah, this was on point that I needed to discuss. We need something for variables, like User, environment, temp path and so on.
There are many ways to implement it, or a config file that we parse, or something else.
My implementation was "the easy way" for the programmer

qiling/os/windows/dlls/kernel32/winnls.py
0ssigeno: f4cf13f should resolve the problem

qiling/os/windows/dlls/kernel32/tlhelp32.py
0ssigeno: f4cf13f should resolve the problem

0ssigeno: f4cf13f should resolve the problem

xwings:
seems like ueslessdisk failing now, i will check later why

0ssigeno:
Because python encoding is shit. If we use "backslashreplace" and then we encode the string with .encode("utf-16le"), that should double the size (every char is followed by a \x00), is instead quadrupled. Why? Because the "backslashreplace" characters are considered 4 different chars, instead of just one.

qiling/os/windows/dlls/msvcrt.py
xwings: i think quynh move it. for what reason i got no idea. maybe @aquynh can answer

qiling/os/windows/variables.py
xwings: like i said, we need to have a option just like ql.gdb, we need to have a ql.win_env = file so user can copy the file point to the file. 

easiest method is, user copy this file, move to location and redefine it as
```
ql.win_env = config.py
```

if nothing is being defined. this file will be default.

xwings:
Cool. how about the env var ?

0ssigeno:
Your call on the implementation.
We can create a config file that the class Qiling parse and store the values in a dict object, but is just one of the possible solutions

xwings:
my idea is. if user did not define  ql.win_env it will goto qiling/os/windows/variables.py of user define 
```
ql.win_env = filename.py
```
then it will point to rootfs/filename.py which user can copy from qiling/os/windows/variables.py edit and just enable ql.win_env

xwings:
can you please also include grandcrab and update travis and the test/test_pe_nu.py ?

0ssigeno:
Will do everything that you asked on monday.

xwings:
Sure, thanks !

xwings:
Hi,

I just tested and it seems there are one (could be more)  API not being implemented.

```
[+] Done with loading rootfs/x86_windows/bin/GandCrab.bin
[!] CreateToolhelp32Snapshot is not implemented
```

0ssigeno:
Are you sure you are using this branch? Because the files are all there.
https://github.com/qilingframework/qiling/blob/5937f0141d0116c9c45c4e4424103e28bf1470e5/qiling/os/windows/dlls/kernel32/tlhelp32.py

xwings:
I am testing  the code you PR'ed so it should be https://github.com/certego/qiling.git

xwings:
I did a merge, on my local so we are good. 

btw is this expected ? I remmember you told me some missing Windows API and we need to force it outself ?
```
winapi_name = ql.PE.import_symbols[address]['name'].decode()
AttributeError: 'NoneType' object has no attribute 'decode'
```

** you don't have to rush to answer

0ssigeno:
I will have to see how the parse works to give a valid answer. The syscall in question is called ShGetSpecialFolderPathW. I think the parser can't find it because it is not more supported. If we try an old shell32.dll (Windows vista for example), should work fine.
https://docs.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetspecialfolderpathw

xwings:
it seems like i have the file. somewhere must be wrong.
```
(7)$ strings shell32.dll | grep -i ShGetSpecialFolderPathW
SHGetSpecialFolderPathW
SHGetSpecialFolderPathW
```

could be the parser or some unknown place

xwings:
I tried the oldest english version and it seems like not working

https://www.dll-files.com/shell32.dll.html

0ssigeno:
Have no idea on why it doesn't parse it.
I think another PR will be needed for that

xwings:
I guess something more fun came out after some debug.

```
175
[!] 175 is not implemented

306
[!] SHGetFolderPathW is not implemented
```

Looks like ordinal 175 is not  SHGetFolderPathW

0ssigeno:
Mmmh, i followed the binary syscall tree, on ida and ghidra, and the call is indeed ShGetSpecialFolderPathW ( you are talking about SHGetFolderPathW, they are different).

xwings:
Sorry, I look at at the wrong shit. let me update it

xwings:
Sorry, i was messing with some really old DLLs and the ordinal could be wrong.

I did a quick test:
```
        try:
            winapi_name = ql.PE.import_symbols[address]['name'].decode()
        except:
            winapi_name = ql.PE.import_symbols[address]['ordinal']
        
        if winapi_name and type(winapi_name) == int:
            winapi_name = str(winapi_name)
```

out come:

```
175
[!] 175 is not implemented
```
My guess is, the parser is not wrong but it did not parse enuf. I did not dig further but i suspect ordinal is somehow a alias-ed Windows API.

So my suggestion is, we might need a mapper, if any winapi_name exits and its a int (due to no name), it will need to goto the mapper and search for its mapping name.

add a mapper in os/windows/const.py call 
```
shell32_dll { 
    175, "ShGetSpecialFolderPathW " 
}
```

Something like that

0ssigeno:
Yeah, ordinal is another way to call a function inside Windows.
Your idea is good in my opinion, but I think this project needs another PR. In this one I will add the binary and the test files, stopping the emulation at the point where the error raise.
After that, we can fix the mapper and continue with GandCrab. Is it ok for you?

xwings:
no Issue, i tested the code with this. Just add on more line after
```
def hook_SHGetSpecialFolderPathW(ql, address, params):
    directory_id = params["csidl"]
    dst = params["pszPath"]
    if directory_id == CSIDL_COMMON_APPDATA:
        path = Environment["appdata"]
        # We always create the directory
        dir = path.split("C:\\")[1].replace("\\", "/")
        path_emulated = os.path.join(ql.rootfs, dir)
        ql.dprint(0, path_emulated)
        ql.uc.mem_write(dst, (path + "\x00").encode("utf-16le"))
        if not os.path.exists(path_emulated):
            os.makedirs(path_emulated, 0o755)
    else:
        raise QlErrorNotImplemented("[!] API not implemented")
    return 1
```

There you go:
```
hook_175 = hook_SHGetSpecialFolderPathW
```

it works, but the only issue we have too many ordinal 175. so hook_shell32_175 will be a better option

xwings:
Also

https://github.com/wine-mirror/wine/blob/master/dlls/shell32/shell32.spec

there is a -noname in ordinal 175. I guess that might be the reason.

xwings:
This PR is ok, we dont have to rush. you can spend sometime fix it will we are good to PR. 

Or if you wish to PR now, I guess we got more to fix. at least in your PR, hook_SHGetSpecialFolderPathW ql.dprint still the old d.print and without symbol.

Let me know your plan.

0ssigeno:
Ok, for me is good, I just thought was better to differentiate goals in PR. Every DLL should have its own ordinal, so there are many number 175. I'm not sure, but is possible that the ordinal is inside the header, so for example, we can have many number 175 all inside kernel32.dll, so I agree we shoul start differentiate numbers from now. 
Will fix hook_SHGetSpecialFolderPathW, forgot to remove the print, my bad.

xwings:
Cool, if you can also fix the conflict for qiling/arch/filetype.py 

I move and split into qiling/const.py and qiling/utils.py. Having a filetype.py in arch folder is ugly.

0ssigeno:
Yeah sure, on monday will do everything that we talked about!

xwings:
Actually i am not sure will mapping be good or just have a alias name. Another method we can do is

in qiling/os/windows/x86.py

```
    if address in ql.PE.import_symbols:
        try:
            winapi_name = ql.PE.import_symbols[address]['name'].decode()
        except:
            winapi_name = ql.PE.import_symbols[address]['dlls'].split('.dll')[0] + "_" + str(ql.PE.import_symbols[address]['ordinal'])
```

in this case we will have do

```
hook_shell32_175 = hook_SHGetSpecialFolderPathW
```
Finally, in qiling/loader/pe.py  around line 74
```
            for entry in dll.DIRECTORY_ENTRY_EXPORT.symbols:
                self.import_symbols[self.ql.DLL_LAST_ADDR + entry.address] = {'name': entry.name,
                                                                              'ordinal': entry.ordinal, 
                                                                              'dlls': dll_name}
```

I this we cannot avoid the dll_name. Or either mapping or alias method.  I am ok for both.

0ssigeno:
Agree, that solution seems better in the long run

0ssigeno:
Check if the solution implemented is fine, is a fusion of your 2 solutions. If you prefer your choice, just let me know and I will change.
Working on the tests now.

xwings:
@chfl4gs  will help on how to locate the file

xwings:
So this dll is not a valid PE ? wtf

@0ssigeno maybe you can try gandcrab which is not 5.0. 

https://www.kashifali.ca/tag/cybercrime/page/2/

qiling/core.py
```diff
@@ -97,7 +97,8 @@ def __init__(
             stack_address=0,
             stack_size=0,
             interp_base=0,
-            automatize_input=False
+            automatize_input=False,
+            windows_config=None
```
xwings: automatized and config remove from here. this two var i assume happens during ql.run so we can just call ql.xxx before ql.run()

xwings:
For examples/rootfs/x86_windows/configuration.cfg  we need to have another copy in qiling/os/windows/configuration.cfg 

if user did not define, we need to use this copy. I guess u need to update MANIFEST.in to add in the cfg too

qiling/core.py
0ssigeno: Yes, i can remove them from here, but how can the user decide if he wants to automatize the input and the config file otherwise?

qiling/core.py
```diff
@@ -125,8 +123,8 @@ def __init__(
         self.global_thread_id = 0
         self.gdb = None
         self.gdbsession = None
-        self.automatize_input = automatize_input
-        self.config = config
+        self.automatize_input = False
+        self.config = os.path.dirname(os.path.abspath(__file__)) + "/os/windows/configuration.cfg"
```
xwings: you might need a os.path.join here. to avoid "\" for windows and "/" for linux

0ssigeno: You are absolutely right.

xwings:
lets merge for now ! thanks !

PR #179, opened 2020-03-29
Add pread64 syscall on x86_64

balayette:
Adds `ssize_t pread(int fd, void *buf, size_t count, off_t offset)` on x86_64, which seems to be used by libc 2.31 on Arch Linux.


xwings:
Can you please provide a sample for this? So we can test

xwings:
did a quick check and seems ok. thanks

PR #182, opened 2020-03-30
verbose level for different debug output

ucgJhe:
merge this for different debug output
0 for normal debug
1 for display constant symbol , read/write/send/recv content


xwings:
Hi,

You might want to take look for Windows too ?

ucgJhe:
will see what i can do

xwings:
THANKS!

but let me merge the big PR first. and we'll see how we move.

xwings:
ok, as expected. we got some conflict.

xwings:
Prefect ! Thanks !

PR #186, opened 2020-03-31
ql_syscall_mkdir arm support added

waveburst:
- ql_syscall_mkdir arm support added 
- umask and getcwd syscall numbers add for arm.


xwings:
cool ! thanks !

PR #187, opened 2020-03-31
ql_syscall_setpriority + ql_syscall_clock_gettime linux arm support

waveburst:
- ql_syscall_setpriority + ql_syscall_clock_gettime linux arm support
- ql_syscall_futex syscall number added for arm


qiling/os/posix/syscall.py
```diff
@@ -239,7 +239,17 @@ def ql_syscall_setuid(ql, null0, null1, null2, null3, null4, null5):
         UID = 1000
     ql.nprint("setuid(%i)" % UID)
     regreturn = UID
-    ql_definesyscall_return(ql, regreturn)     
+    ql_definesyscall_return(ql, regreturn)
+
+
+def ql_syscall_setpriority(ql, which, who, prio, null3, null4, null5):
```
xwings, line 245: i am not certain about this, we might dont even to call os.piority. 

do you have a test case for this ?

qiling/os/posix/syscall.py
```diff
@@ -239,7 +239,17 @@ def ql_syscall_setuid(ql, null0, null1, null2, null3, null4, null5):
...
+    ql.nprint("setpriority(%i %i %i)" % (which, who, prio))
+    try:
+        os.setpriority(which, who, prio)
```
xwings, line 248: Dont have to set anything. return a 0 should be all fine

waveburst: Yeah you right, its not really necessary to call the actual syscall. 
I was trying to emulate Ruckus R510 and of course set-priority is not really matter.

xwings: cool, we split the syscall file finally. so i guess you need to update the syscall. thanks !

xwings:
hi, any update on this. thanks !

qiling/os/posix/syscall.py
waveburst: I'm not sure after the split where this change should be done.
./build/lib/qiling/os/posix/syscall.py or ./qiling/os/posix/syscall/unistd.py

chenhuitao: We split the posix/syscall.py file into multi-files in posix/syscall dir. The syscall should be in the file whose name is same with the header file which define the syscall fucntion. This case, setpriority is defined in resource.h,  ql_syscall_setpriority should be in syscall/resource.py. ql_syscall_clock_gettime should be in syscall/time.py.

xwings: @waveburst maybe u can move this ql_syscall_setpriority to qiling/os/posix/syscall/resource.py ?

PR #190, opened 2020-04-03
keyword end for ql.print functions

ucgJhe:
merger this for support keyword "end" in ql.print functions

### DIFF
--- a/qiling/utils.py
+++ b/qiling/utils.py
@@ -312,8 +312,12 @@ def ql_setup_logging_stream(ql, logger=None):
     ql_mode = ql.output
 
     # setup StreamHandler for logging to stdout
-    #ch = logging.StreamHandler()
-    ch = logging.getLogger()
+    if ql.log_console == True:
+        ch = logging.StreamHandler()
+    else:
+        # not print out to stdout by using NullHandler
+        ch = logging.NullHandler()
+
     ch.setLevel(logging.DEBUG)
     
     """
@@ -321,9 +325,9 @@ def ql_setup_logging_stream(ql, logger=None):
     with print ("", end ="") with logger, 
     fix it with concat, will keep it for now
     """
-    # if ql_mode in (QL_OUT_DISASM, QL_OUT_DUMP):
-    #     # use empty string for newline if disasm or dump mode was enabled
-    #     ch.terminator = ""
+    if ql_mode in (QL_OUT_DISASM, QL_OUT_DUMP):
+        # use empty string for newline if disasm or dump mode was enabled
+        ch.terminator = ""
 
     if logger is None:
         logger = ql_setup_logger()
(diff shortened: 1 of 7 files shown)


qiling/utils.py
```diff
@@ -312,18 +312,22 @@ def ql_setup_logging_stream(ql, logger=None):
...
     guess this was added due to the incompatablity
     with print ("", end ="") with logger, 
     fix it with concat, will keep it for now
     """
-    # if ql_mode in (QL_OUT_DISASM, QL_OUT_DUMP):
-    #     # use empty string for newline if disasm or dump mode was enabled
-    #     ch.terminator = ""
+    if ql_mode in (QL_OUT_DISASM, QL_OUT_DUMP):
+        # use empty string for newline if disasm or dump mode was enabled
+        ch.terminator = ""
```
xwings, line 330: i think this we gonna have issue in dprint. it will take out all the default \n ?

ucgJhe: if we use it with format string like print("%s" % var) will not cause any trouble, and i test these two modes with hello_world example its good so far.

xwings: Any specific reason why we need this ? Since we already fix the issue in nprint

ucgJhe: since we are using the python logging module, its not supporting keyword "end"  like normal print did, and this is the most easy way to do it, set all stream with empty string terminated and print it out with append newline character or the specific one assigned by keyword "end" we need.

xwings: i mean only in QL_OUT_DISASM, QL_OUT_DUMP why ?

ucgJhe: its special needs for `ql_hook_code_disasm`, or I can move it to nprint function, change the terminator while using keyword end , and huge context switch will be expected due to these two modes use `ql_hook_code_disasm` a lot.
So I think keep it empty default and add a newline character when we need it at nprint will be more efficiency.

xwings:
I guess u need another test. I did a quick test and i agree on the "" termininator, but not only on QL_OUT_DUMP or DISASM only 

These are the few command u can test. it seems like in nomal mode we still need the "" terminator

```
python ../qltool run -f rootfs/x8664_linux/bin/x8664_hello --rootfs rootfs/x8664_linux --debug
```

```
python ../qltool run -f rootfs/x8664_linux/bin/x8664_hello --rootfs rootfs/x8664_linux --dump
```

```
python ../qltool run -f rootfs/x8664_linux/bin/x8664_hello --rootfs rootfs/x8664_linux
```

ucgJhe:
you are right , looks extra newline on debug mode, by the way any reason that create file and directories in the function ?
https://github.com/qilingframework/qiling/blob/master/qiling/utils.py#L335-#L337

FileHandler should do job

xwings:
I did not write 

https://github.com/qilingframework/qiling/blob/master/qiling/utils.py#L335-#L337

feel free to delete it. use those 4 command and test (i added one more).   I guess that is all the test cases we need.

ucgJhe:
okay looks all good now

xwings:
Cool. Thanks !

xwings:
tested and all good !

PR #195, opened 2020-04-04
try to tabify logging output

ucgJhe:
as title.


xwings:
cool, thanks. I will merge and give it a try !

PR #199, opened 2020-04-06
 add syscall filter support under default(strace) mode

ucgJhe:
add syscall filter for qiling, for better debugging and easy to use with qltool

image


qiling/core.py
```diff
@@ -50,6 +50,7 @@ def __init__(
             stdout=0,
             stderr=0,
             output=None,
+            syscall_filter=None,
```
xwings, line 53: since this is during execution, we don't have to add in the init, just self.syscall_filter = None and at line 64 will be good

qltool
```diff
@@ -235,7 +238,7 @@ if __name__ == '__main__':
 
         if options.filename != '' and options.run_args == []:
             ql = Qiling(filename=[options.filename] + options.args, rootfs=options.rootfs, 
-                        output=options.output, log_console = options.console, log_dir=options.log_dir)
+                        output=options.output, log_console = options.console, log_dir=options.log_dir, syscall_filter=options.syscall_filter)
```
xwings, line 241: same here

xwings:
i guess there is a conflict

qiling/core.py
```diff
@@ -202,10 +205,15 @@ def __init__(
...
         QL_ENDIAN_EL = Little Endian
         big endian is define during ql_elf_check_archtype
+
+        since there is no elf for shellcoder mode
+        judge archendian by whether bigendian set or not
         """
         self.archbit = ql_get_arch_bits(self.arch)
         if self.arch not in (QL_ENDINABLE):
             self.archendian = QL_ENDIAN_EL
+        else:
```
xwings, line 215: i left a self.bigendian and i totally forgotten, i guess we can do this 
```
        if self.arch not in (QL_ENDINABLE):
            self.archendian = QL_ENDIAN_EL
        if self.shellcoder and self.bigendian == True:
            self.archendian = QL_ENDIAN_EB
        elif self.shellcoder:
            self.archendian = QL_ENDIAN_EL  
```

qltool
```diff
@@ -200,6 +206,7 @@ if __name__ == '__main__':
     shellcode_parser.add_argument('--bin', action='store_true', default=True, dest='bin', help='input file format, -bin')
     shellcode_parser.add_argument('--output', required=False, default='default', help='output mode, options are off, debug , disasm, dump')
     shellcode_parser.add_argument('--strace', action='store_true', default=False, dest='strace', help='Run in strace mode')
+    shellcode_parser.add_argument('-e', '--syscall-filter', required=False, dest="syscall_filter", default=None, help="Only work with strace mode, you can choose what to be printout, multi-syscalls should be separated by comma")
```
xwings, line 209: syscall_filter or we should name it strace filter

qiling/core.py
```diff
@@ -267,6 +275,20 @@ def shellcode(self):
         loader_shellcode(self)
 
     def run(self):
+
+        # setup syscall filter for logger
+        if self.syscall_filter != None and self.output == QL_OUT_DEFAULT:
```
xwings, line 280: we need to move this to somewhere more proper, core.py try to me as os/arch friendly as possible. syscall is too *nix

xwings:
Sorry, man. you need to do more changes. thanks for the help.

ucgJhe:
@xwings can you merge this after passing Travis ?
will make another PR for macos

xwings:
prefect! Thanks !

PR #204, opened 2020-04-09
Use ==/!= to compare str, bytes, and int literals

cclauss:
Identity is not the same thing as equality in Python so use ==/!= to compare str, bytes, and int literals. In Python >= 3.8, these instances will raise SyntaxWarnings so it is best to fix them now. https://docs.python.org/3.8/whatsnew/3.8.html#porting-to-python-3-8

% python3.8
```
>>> 2 is 2
<stdin>:1: SyntaxWarning: "is" with a literal. Did you mean "=="?
True
```


xwings:
Hi! 

welcome to the project. Just wondering will you be able to do a quick check and see if we can more this kind of mistake so you can help us to fix it ?

THANKS !

aquynh:
merged, thanks!

PR #205, opened 2020-04-10
Running performance test on Qiling code

klks:
This will allow us to track where Qiling framework is spending most of its time processing, will allow us to also look for performance improvements of code.


xwings:
welcome back sir ! thanks !

PR #206, opened 2020-04-10
More accurate page perms while mapping

ucgJhe:
try to make whole page perms thing looks like in gdb, also map perms and handle add_mapinfo in mem.map.


xwings:
hey buddy, 

try leave reasons/comments why you comment out the code? so we will know why and do we need to fix it in the future
if the code really useless, delete it

qiling/os/memory.py
```diff
@@ -302,16 +315,17 @@ def map(self, addr, size, perms=UC_PROT_ALL, ptr = None):
         '''
         if ptr == None:
             if self.is_mapped(addr, size) == False:
-               self.add_mapinfo(addr, addr + size, 'rw-', "[mapped]")
-               self.ql.uc.mem_map(addr, size)
+               self.ql.uc.mem_map(addr, size, perms)
+               self.add_mapinfo(addr, addr+size, perms, info if info else "[mapped]")
```
xwings, line 319: just to make sure, so now all the mapinfo will be handle by mem.map and not syscall, we need to make sure all these are os free.

ucgJhe: all add_mapinfo has been deleted except one in macos

ucgJhe: just removed it since we don't really use it right now

xwings: maybe can take a quick look at macos ?

ucgJhe: i did , check this commit b86a5f6.
we are not using it right now, and its probably not goona work because there is no prot_mapping for macos, so I removed it.

xwings: ok, let me check one more round later and commit

xwings:
sorry buddy, i update dprint and here comes a conflict

ucgJhe:
will redo this till new version released

PR #212, opened 2020-04-12
Split core to sub-files

klks:


xwings:
First commit from dev !

PR #215, opened 2020-04-12
Argument change from log_file to log_dir in example

kxynos:
log_file doesn't exist as an argument after changes to core.py


xwings:
hi, 

welcome, thanks for pointing it out

PR #217, opened 2020-04-12
NoneType checking of ql.thread.management

kxynos:
Avoid this error: 
  File "/usr/local/lib/python3.6/dist-packages/qiling-0.9-py3.6.egg/qiling/os/posix/syscall/sched.py", line 71, in ql_syscall_clone
    f_th = ql.thread_management.cur_thread	
AttributeError: 'NoneType' object has no attribute 'cur_thread'


xwings:
I don't think will help us if the binary designed as multi thread. 

by default multithread is off by default, you need to turn it on by using ql.multithread = True

PR #227, opened 2020-04-14
Updates to posix/syscall/unistd.py and os/utils.py 

kxynos:


xwings:
Hi,

I suggest you can merge read and write string in to one function and move to os/memory.py

so it will be ql.mem.string()

ql.mem.string(address) means read
ql.mem.string(address, value) means write

xwings:
Let me know once you update the code !

kxynos:
Code is ready. Do you want me to replace old ql_read_string too ?

xwings:
what i am trying to say it. merge  your ql_writing_string into qiling/os/memory.py and make it 

ql.mem.string(address) means read
ql.mem.string(address, value) means write

xwings:
i made changes to qiling/os/memory.py and u just need to integrate your code

kxynos:
I am testing the code I wrote for qiling/os/memory.py and will upload it soon.

PR #228, opened 2020-04-15
try to map accurately

ucgJhe:
merge this for more accurate mapping. All add_mapinfo and del_mapinfo now handled by mem.map and mem.unmap


w1tcher:
Thank you.

PR #232, opened 2020-04-15
Updated os/memory.py and syscall/unistd.py

kxynos:


xwings:
Hi, 

Again, welcome to the project. 

This is good now. Thanks !

Issue #231, opened 2020-04-15
Add AT_SECURE auxiliary vector for Android native binaries


justfoxing:
I've been messing around with running Android native binaries in Qiling. I've had to tweak some syscalls a little, but thanks to your API that's been super easy to work with :D

However, I did run into one place where I needed to modify Qiling code - currently, as bionic libc initializes in an Android binary, it checks for the existence of AT_SECURE in the auxiliary vectors, and errors out if it's not present (https://cs.android.com/android/platform/superproject/+/master:bionic/libc/bionic/libc_init_common.cpp;bpv=1;bpt=1;l=303?q=AT_SECURE).

I managed to work around this by adding `elf_table += self.NEW_AUX_ENT(AT_SECURE, 0, ql)` immediately before `elf_table += self.NEW_AUX_ENT(AT_NULL, 0, ql)` in qiling/loader/elf.py. Any chance you could add this, or expose a flag to allow users to specify if they want AT_SECURE?

Cheers!

xwings:
hi,

will you be able to make a pull request and we work it out from there ?

justfoxing:
Sure, just added one.

xwings:
thanks. will wait for your sample test and we are good to go

PR #233, opened 2020-04-16
Add AT_SECURE auxiliary value to loader/elf.py

justfoxing:
Helps support Android native binaries (bionic libc will abort during init if AT_SECURE isn't present in the auxiliary vector).

At this stage, just sets 0 (non-secure) for the value.

Issue #231


aquynh:
Nice work! Can you please add a testcase into examples/ as well?

justfoxing:
Sure, I'll put something together.

justfoxing:
Testcase added for Android on arm64 (arm32 is something I'm still trying to solve).

tests/test_android.py
```diff
@@ -0,0 +1,82 @@
...
+    except:
+        regreturn = -1
+
+    ql.nprint("getrandom(0x%x, 0x%x, 0x%x) = %d" %
+              (buf, buflen, flags, regreturn))
+
+    if data:
+        ql.dprint(0, "[+] getrandom() CONTENT:")
+        ql.dprint(0, str(data))
+    ql_definesyscall_return(ql, regreturn)
```
xwings, line 33: hi, we did a change so you can call
```
ql.os.definesyscall_return
```

can you update and try it. thanks

justfoxing: Hi, I'm not seeing that change in either dev or master branches yet (I'm getting "'QlOsLinux' object has no attribute 'definesyscall_return'" when I run). Has it been merged in yet?

justfoxing: Haha, yes it has, I was just failing to see it, my bad. Changed over to ql.os.definesyscall_return.

xwings:
perfect ! i will add it in to the test

Issue #222, opened 2020-04-13
failed to resolve multiple symbolic links


cq674350529:
When running a binary of Netgear `R6400` firmware using `qiling` framework, it failed to resolve multiple symbolic links. Take the following as an example.

 ```shell
~/squashfs-root$ ls -l ./lib/libbdbroker.so 
lrwxrwxrwx 1 root root 48 Jul 29  2019 ./lib/libbdbroker.so -> /opt/bitdefender/patches/base/lib/libbdbroker.so

~/squashfs-root$ ls -l ./opt/bitdefender 
lrwxrwxrwx 1 root root 23 Jul 29  2019 ./opt/bitdefender -> /media/nand/bitdefender
```
As to `/lib/libbdbroker.so`, the `ql_transform_to_real_path()` will return `~/squashfs-root/opt/bitdefender/patches/base/lib/libbdbroker.so`. And this path doesn't exist, thus causing fail to load the `libbdbroker.so`. Actually, it should resolve the symbolic link `/opt/bitdefender` again.

> PS: all symbolic links work in `chroot` env.

Thanks!

xwings:
two method,

1. remove the symlink and replace with a real file. which you need to do that in IoT analysis most of the time.
2. write a patch ? 

i did not try symblock link, maybe we should consider that,  

chroot will work, but chroot did not give you the power like what Qiling can give you.

cq674350529:
Yes, I can fix it manually. But as you said, with the powerful `qiling` framework, I prefer to be "lazy" and let the framework handle it. I'll have a try to see if I can work out a patch for it.

xwings:
thanks !

xwings:
thanks for the PR !

PR #240, opened 2020-04-18
support multilevel symbolic link

cq674350529:
try to fix https://github.com/qilingframework/qiling/issues/222

Currently, if the result of `ql_transform_to_real_path()` contains directory symbolic links, it will fail to resolve it.


xwings:
thanks !

PR #242, opened 2020-04-20
Al khaser fixes

0ssigeno:
Is now possible to execute arbitrary code from inside a hook.
Fixed the test to print the result of alkhaser.


xwings:
One Big Step. Thanks !

PR #244, opened 2020-04-22
Fix exception in registry

Moloson:
By default, when running the wannacry_x86_windows_hookaddress.py or any script that uses registry.py, exception is raised due to the fact that **QlPrintException** is not an exception object.

This commit fix-it by raising a **QlErrorFileNotFound**.


xwings:
welcome to Qiling,

Thanks !

PR #245, opened 2020-04-23
Al khaser2

0ssigeno:


qiling/core.py
```diff
@@ -88,7 +87,8 @@ def __init__(
         self.internal_exception = None
         self.platform = platform.system()
         self.debugger = None
-        self.automatize_input = False 
+        self.automatize_input = False
+        self.profile = profile
```
xwings, line 91: leave self.profile in line 72. its comes with Qiling() so there will be clear.

xwings:
Nice. Thanks!

PR #246, opened 2020-04-24
More API Hooks and Minor Windows Clipboard Module Fixes

jhumble:
Added some new API hooks and made some fixes to the clipboard module which was broken when last_error was moved from ql to ql.os.

### DIFF
--- a/qiling/os/windows/dlls/kernel32/debugapi.py
+++ b/qiling/os/windows/dlls/kernel32/debugapi.py
@@ -47,3 +47,13 @@ def hook_OutputDebugStringW(ql, address, params):
     string = params["lpOutputString"]
     ql.os.stdout.write(string.encode())
     return 0
+
+#void OutputDebugStringA(
+#  LPCSTR lpOutputString
+#);
+@winapi(cc=STDCALL, params={
+    "lpOutputString": STRING
+})
+def hook_OutputDebugStringA(ql, address, params):
+    ql.os.stdout.write('OutputDebugStringA: "%s"' % (params['lpOutputString']))
+    return 0
--- a/qiling/os/windows/dlls/kernel32/debugapi.py
+++ b/qiling/os/windows/dlls/kernel32/debugapi.py
@@ -45,7 +45,7 @@ def hook_CheckRemoteDebuggerPresent(ql, address, params):
 })
 def hook_OutputDebugStringW(ql, address, params):
     string = params["lpOutputString"]
-    ql.os.stdout.write(string.encode())
+    ql.nprint('OutputDebugStringW: "%s"' % (string.encode()))
     return 0
 
 #void OutputDebugStringA(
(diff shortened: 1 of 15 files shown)


qiling/os/windows/dlls/wsock32.py
```diff
@@ -72,3 +72,29 @@ def hook_connect(ql, address, params):
...
+#  char  **h_aliases;
+#  short h_addrtype;
+#  short h_length;
+#  char  **h_addr_list;
+#} HOSTENT, *PHOSTENT, *LPHOSTENT;
+@winapi(cc=STDCALL, params={
+    "name": STRING
+})
+def hook_gethostbyname(ql, address, params):
+    dummy_ip = b"\x40\x30\x20\x10" #10.20.30.40
```
xwings, line 90: we should store this in profile? its in qiling/profile/windows.ql

jhumble: Thanks! Do you want me to just add a [Network] section?

xwings: sure, any appropriate name will work

qiling/os/windows/dlls/kernel32/debugapi.py
```diff
@@ -47,3 +47,13 @@ def hook_OutputDebugStringW(ql, address, params):
...
     return 0
+
+#void OutputDebugStringA(
+#  LPCSTR lpOutputString
+#);
+@winapi(cc=STDCALL, params={
+    "lpOutputString": STRING
+})
+def hook_OutputDebugStringA(ql, address, params):
+    ql.os.stdout.write('OutputDebugStringA: "%s"' % (params['lpOutputString']))
```
xwings, line 58: will it be better to make this two lines of ql.os.stdout.write into ql.drpint ?

jhumble: I'm not sure. I initially did have this implemented only as ql.dprint, but noticed that the existing OutputDebugStringW implementation only used ql.os.stdout.write, so I just changed mine to match it.

That doesn't seem correct though, since OutputDebugString will not actually emit any console input. I'll change both of them to use ql.dprint.

xwings: the reson why ql.drpint or ql.nprint we are in control. we can choose to print or not to print. in this case should be nprint and not dprint.

qiling/os/windows/dlls/wsock32.py
jhumble: Added as [NETWORK]: dns_response_ip

qiling/os/windows/dlls/kernel32/debugapi.py
jhumble: Changed to nprint only

jhumble:
Cleaned up clipboard.py in accordance with the FIXME notes that were in it. No need for it to have a reference to ql, only to ql.os. 

Added profile.ql option to set the ip returned on dns requests (gethostbyname). 

Fixed a bug in the Mutex implementation in os/windows/structs.py (a variable and method were both named "lock"). 

Changed hook_GetLocaleInfoA and _CreateFile to both set the appropriate windows error on failure and return rather than crashing the emulator

Fixed other hooks that were throwing API Implementation Errors when running some bulk random executables through.

xwings:
Once again, good to have you back!

Thanks for the PR.

PR #247, opened 2020-04-24
Bug fixes

jhumble:
I managed to forget to include the changes to wsock32.py and fileapi.py in commit: https://github.com/qilingframework/qiling/commit/bff5fd4139f70ee868b11815683732d88821202c

I added those, and also included a fix in the pe loader that occurs when trying to load a dll with no imports (ntdll.dll for example)

I additionally added some code to parse exports for the main module being loaded. This allows execution of various exported functions for a dll similarly to rundll32. 

Some executables (especially malware) will have exported functions which are found and called by calling LoadLibrary against themselves, then using that handle to find the addresses of exported funtions with GetProcAddress.
Previously this would crash, but now works. I updated the GetProcAddress hook to check if hMod == self, and if so to search through the newly available exports.

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -239,6 +239,8 @@ class Process(QlLoader):
         self.ldr_list.append(ldr_table_entry)
 
     def init_exports(self):
+        if self.ql.shellcoder:
+            return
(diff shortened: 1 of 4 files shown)


xwings:
We need a fix. travis is not happy :)

xwings:
Travis is happy now!

thanks for the PR

Issue #254, opened 2020-04-27
small bugs with gdb debugserver


D3Smind:
On my ubuntu the installer does not copy the qiling/qiling/debugger/gdbserver/xml/ folder when installing and such the gdb debugserver fails.

When debugging PE files with qiling the Server fails at line 128 in the /qiling/blob/master/qiling/debugger/gdbserver/gdbserver.py
`if self.gdb.bp_list in ([self.ql.loader.elf_entry], [self.ql.loader.entry_point]):`
Due to the PE Loader not having a elf_entry.

xwings:
i remember PE works, let me check again.

D3Smind:
I tried again on a clean install. The problem of the missing xmls persists.
gdb> Xml file not found: /usr/local/lib/python3.6/dist-packages/qiling-1.0-py3.6.egg/qiling/debugger/gdbserver/xml/x8664/target.xml

When copy the files manually elf binaries work. The x8664_hello.exe does run but the debugger disconnects and it prints:

> Traceback (most recent call last):
>   File "gdb.py", line 30, in <module>
>     test_gdb(["rootfs/x8664_windows/bin/x8664_hello.exe"], "rootfs/x8664_windows")
>   File "gdb.py", line 26, in test_gdb
>     ql.run()  
>   File "/usr/local/lib/python3.6/dist-packages/qiling-1.0-py3.6.egg/qiling/core.py", line 203, in run
>     self.remotedebugsession.run()
>   File "/usr/local/lib/python3.6/dist-packages/qiling-1.0-py3.6.egg/qiling/debugger/gdbserver/gdbserver.py", line 708, in run
>     commands[cmd](subcmd)
>   File "/usr/local/lib/python3.6/dist-packages/qiling-1.0-py3.6.egg/qiling/debugger/gdbserver/gdbserver.py", line 128, in handle_c
>     if self.gdb.bp_list in ([self.ql.loader.elf_entry], [self.ql.loader.entry_point]):
> AttributeError: 'QlLoaderPE' object has no attribute 'elf_entry'

xwings:
Hi,

Can i know which version of gdb and qiling are u running on. I got a crash and its crash on gdb. not QIling.

xwings:
i add a quick fix in https://github.com/qilingframework/qiling/commit/743cfad06056142e30f78bb984dacae2667361b1

u can clone dev branch and try

D3Smind:
Hi, 

[...]

PR #255, opened 2020-04-28
fixes for gdb debugserver

D3Smind:
fixes for the problems described in #254

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,2 +1,2 @@
-recursive-include qiling/gdbserver/xml *
+recursive-include qiling/debugger/gdbserver/xml *
 recursive-include qiling/profiles *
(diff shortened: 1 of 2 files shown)


xwings:
The problem is,

i tried with gdb and ida pro it seems to be working fine withe the exe. can u do a pull and check again ?

xwings:
But you are right with the Manifest :)

xwings:
image

should be working

xwings:
i make the entry_point standardized in gdb so we dont have to be confuse. can you try and and update the conflict. 

i will merge this PR after u solve the conflict

D3Smind:
that is the far better solution :+1: 

thx for the help and the great emulator

xwings:
Once again,

Welcome to the group. Hope to see more PR !

PR #256, opened 2020-04-28
fix mips archtype in shellcode_run.py

cq674350529:
change archtype to `mips32`, since there is no `mips32el` anymore.
https://github.com/qilingframework/qiling/blob/617bcee20c77c8940dcd9aaf867e9b27a4054812/qiling/utils.py#L87-L98

Also, the archendian is `LE` in default.


xwings:
Hi, 

you got a good eye!

PR #257, opened 2020-04-28
Is now possible to retrieve each syscall that qiling emulated, withou…

0ssigeno:
…t having to parse the result file


xwings:
nice work!

PR #261, opened 2020-05-01
New and fixed API hooks

jhumble:
Added some new api hooks, fixed some others. 

Decorated winapi with https://github.com/micheles/decorator.
Most Windows APIs come in two versions, one with a W suffix for widestrings and one with A suffix for ascii. Due to this, many of the api hooks are just duplicates or in some cases only one version exists. With this decorator, it's possible to access the undecorated version of the hook, so rather than copy pasting the entire function for the second version, the second version can often be implemented like this:

def hook_CreateEventW(ql, address, params):                                                                                                                                                                                                                                                                                                                                                                        
    return hook_CreateEventA.__wrapped__(ql, address, params)

In my opinion this is a lot cleaner and allows for less unanticipated divergences between the two versions. This just happened to be the cleanest way I could find to do this, but if there is a better way, let me know. 

At some point the parameter list for ql.hook_mem_unmapped was changed, but ql_x86_windows_hook_mem_error in qiling/os/windows/utils.py was not updated accordingly. I added the new access parameter so it would stop throwing errors.

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -257,7 +257,7 @@ class Process(QlLoader):
                 self.import_address_table[dll_name][entry.ordinal] = self.PE_IMAGE_BASE + entry.address
         except:
             import traceback
-            self.ql.nprint('Failed to load exports for %s:\n%s' % (self.name, traceback.format_exc()))
+            self.ql.nprint('Failed to load exports for %s:\n%s' % (self.ql.filename, traceback.format_exc()))
 
 
 class QlLoaderPE(Process, QlLoader):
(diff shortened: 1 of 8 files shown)


aquynh:
this is nice! but can you please split this into 5-6 pull reqs, each fixes a separate issue, so it is easier to track in commit history?

thanks.

xwings:
Hi,

i still did not  see where did you use decorator>=4.2.1

jhumble:
> 
> 
> Hi,
> 
> i still did not see where did you use decorator>=4.2.1

My bad, I was using the decorator module , but ended up replacing with 

> from functools import wraps

 in qiling/os/windows/fncc.py since it's part of the standard lib and does basically the same thing. I'll remove from decorator from requirements. 

I'll take a stab at splitting up the pull request tonight. I'm pretty terrible with git, so it might take a while.

xwings:
For now I guess we are good.

Previously I just worry about the decorators.

I will merge after the test pass.

xwings:
cool. 

thanks!

xwings:
Hey man, try to dig your email address and seems like i got no luck. Will you be able to drop me an email. kj@qiling.io

PR #262, opened 2020-05-02
Experimental feature to expose registers to ql

klks:
Experimental feature to expose registers to QL, starting with X86
this will allow us to do things like this

```
if ql.registers.eax == 0xDEADBEEF:
    ql.registers.eax = 0xBEEFDEAD
    ql.registers.bx = 0xFFFF
```


xwings:
lets try!

PR #269, opened 2020-05-04
Post Report: We want every string

0ssigeno:
Is possible to retrieve both the CSTRING or WSTRING and their addres thank to the parameter in the decorator STRING_ADDR and WSTRING_ADDR.

We are now saving every string that is used inside an api, that is moved in the memory, in ql.os.appeared_strings.
Is not necessary anymore to parse the result file, is possible to use qiling as a framework, as intended

### DIFF
--- a/qiling/os/utils.py
+++ b/qiling/os/utils.py
@@ -212,20 +212,20 @@ class QLOsUtils:
         return relative_path
 
     def post_report(self):
-        self.ql.dprint(D_INFO, "[+] Syscalls called")
+        self.ql.dprint(D_RPRT, "[+] Syscalls called")
         for key, values in self.ql.os.syscalls.items():
-            self.ql.dprint(D_INFO, "[-] %s:" % key)
+            self.ql.dprint(D_RPRT, "[-] %s:" % key)
             for value in values:
-                self.ql.dprint(D_INFO, "[-] %s " % str(dumps(value)))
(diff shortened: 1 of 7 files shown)


qiling/os/utils.py
```diff
@@ -222,6 +222,11 @@ def post_report(self):
             self.ql.dprint(D_INFO, "[-] %s:" % key)
             for value in values:
                 self.ql.dprint(D_INFO, "[-] %s " % str(dumps(value)))
+        self.ql.dprint(D_INFO, "[+] Strings")
+        for key, values in self.ql.os.appeared_strings.items():
```
xwings, line 226: all the post report should change to D_RPRT ? i named that after report

qiling/os/windows/fncc.py
```diff
@@ -32,7 +33,6 @@ def _x86_get_params_by_index(ql, index):
 
 
 def _x8664_get_params_by_index(ql, index):
-    #self.ql = ql
     reg_list = [UC_X86_REG_RCX, UC_X86_REG_RDX, UC_X86_REG_R8, UC_X86_REG_R9]
```
xwings, line 36: u might be able to use  "rcx" and things here. so can reduce the import of unicorn everywhere

xwings:
cool. thanks !

PR #272, opened 2020-05-05
Registries and the Crab

0ssigeno:
Fixed registries. Now i possible to select a precise key (like KEY_LOCAL_MACHINE\HARDWARE\DESCRIPTION\System\CentralProcessor\0\ProcessorNameString ) where the last part is the value_name.

Fixed gandcrab to work without hardcode shit in the code (we change just some registries, still have to work on why it crashes).


tests/profiles/windows_gandcrab.ql
```diff
@@ -0,0 +1,63 @@
...
+heap_address = 0x500000000
+heap_size = 0x5000000
+stack_address = 0x7ffffffde000
+stack_size = 0x40000
+image_address  = 0x400000
+dll_address = 0x7ffff0000000
+entry_point = 0x140000000
+
+
+[OS32]
```
xwings, line 11: dont have to be a complete profile. u can pick and choose you wanted.

PR #273, opened 2020-05-06
Is possible to have partial_set_api.

0ssigeno:
If the parameters matches, the function is called instead of the default syscall.
Works only for windows :(

### DIFF
--- a/qiling/core_hooks.py
+++ b/qiling/core_hooks.py
@@ -312,6 +312,13 @@ class QLCoreHooks(object):
         else:
             self.os.add_function_hook(api_name, my_func)
 
+    def set_partial_api(self, api_name, params: dict, my_func):
+        if self.ostype == QL_OS.WINDOWS:
+            params["func"] = my_func
+            self.os.user_partial_defined_api[api_name] = params
+        else:
+            raise QlErrorCoreHook("Only windows is supported for now :(")
+
     # ql.func_arg - get syscall for all posix series
     @property
     def func_arg(self):
--- a/qiling/core_hooks.py
+++ b/qiling/core_hooks.py
@@ -315,7 +315,8 @@ class QLCoreHooks(object):
     def set_partial_api(self, api_name, params: dict, my_func):
         if self.ostype == QL_OS.WINDOWS:
             params["func"] = my_func
-            self.os.user_partial_defined_api[api_name] = params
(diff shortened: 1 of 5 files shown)


qiling/core_hooks.py
```diff
@@ -312,6 +312,13 @@ def set_api(self, api_name, my_func):
         else:
             self.os.add_function_hook(api_name, my_func)
 
+    def set_partial_api(self, api_name, params: dict, my_func):
```
xwings, line 315: the name is kind of bad. anyway we can merge into set api

btw, if we can do it for windows. we should be able to do it with UEFI at least ?

qiling/os/windows/utils.py
```diff
@@ -6,6 +6,7 @@
 import uuid
 from qiling.const import *
 from qiling.os.const import *
+from qiling.os.windows.const import *
```
xwings, line 9: only .const ?

xwings:
before i forget, example and travis too. thanks

qiling/core_hooks.py
0ssigeno: We can't merge them together because in windows,  the function hook_winapi() that is the center of set_api(),  does not have the knowledge of the parameters

0ssigeno: I can give a look about uefi, but first I have to understand how things work there

xwings: trow in a example and i will see what can i name

klks: my 2 cents
retain set_api, but expand a 3rd param and check if its a dict
```
def set_api(self, api_name, my_func, params=None):
    if isinstance(params, dict):
        if self.ostype == QL_OS.WINDOWS:
            params["func"] = my_func
            self.os.user_partial_defined_api[api_name] = params
        else:
            raise QlErrorCoreHook("Only windows is supported for now :(")
    else:
        if self.ostype == QL_OS.WINDOWS:
            self.os.user_defined_api[api_name] = my_func
        else:
            self.os.add_function_hook(api_name, my_func)
```

aquynh:
This looks like an adhoc hack, and can be easily implemented in your application. Sorry that I dont see this as an API to be supported for many usecase.

0ssigeno:
Ok, closing the PR if is not interesting

PR #285, opened 2020-05-15
Try to bring back syscall filter and StreamHandler

ucgJhe:
as title. other OS API filter will be next

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -12,7 +12,7 @@ from .const import QL_ENDINABLE, QL_ENDIAN, QL_POSIX, QL_OS_ALL, QL_OUTPUT, QL_O
 from .exception import QlErrorFileNotFound, QlErrorArch, QlErrorOsType, QlErrorOutput
 from .utils import arch_convert, ostype_convert, output_convert
 from .utils import ql_is_valid_arch, ql_get_arch_bits
-from .utils import ql_setup_logging_env
+from .utils import ql_setup_logging_env, ql_setup_logging_stream
 from .core_struct import QLCoreStructs
 from .core_hooks import QLCoreHooks
 from .core_utils import QLCoreUtils
@@ -133,8 +133,11 @@ class Qiling(QLCoreStructs, QLCoreHooks, QLCoreUtils):
         if self.log_dir != "" and type(self.log_dir) == str:
             _logger = ql_setup_logging_env(self)    
             self.log_file_fd = _logger
-        else:
-            self.log_dir = None     
+        elif self.console == True: # only print out in console
+            _logger = ql_setup_logging_stream(self)
+            self.log_file_fd = _logger
+            # self.log_dir = None
+
         
         # qiling output method conversion
         if self.output and type(self.output) == str:
@@ -198,6 +201,7 @@ class Qiling(QLCoreStructs, QLCoreHooks, QLCoreUtils):
         # resume with debugger
         if self.debugger is not None:
             self.remotedebugsession.run()
+
 
     # patch @code to memory address @addr
     def patch(self, addr, code, file_name=b''):
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -136,7 +136,6 @@ class Qiling(QLCoreStructs, QLCoreHooks, QLCoreUtils):
         elif self.console == True:
             _logger = ql_setup_logging_stream(self)
             self.log_file_fd = _logger
-            # self.log_dir = None
 
         # qiling output method conversion
         if self.output and type(self.output) == str:
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -133,12 +133,11 @@ class Qiling(QLCoreStructs, QLCoreHooks, QLCoreUtils):
         if self.log_dir != "" and type(self.log_dir) == str:
             _logger = ql_setup_logging_env(self)    
             self.log_file_fd = _logger
-        elif self.console == True: # only print out in console
+        elif self.console == True:
             _logger = ql_setup_logging_stream(self)
             self.log_file_fd = _logger
             # self.log_dir = None
 
-        
         # qiling output method conversion
         if self.output and type(self.output) == str:
             # setter / getter for output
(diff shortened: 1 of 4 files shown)


qiling/core.py
```diff
@@ -133,9 +133,11 @@ def __init__(
         if self.log_dir != "" and type(self.log_dir) == str:
             _logger = ql_setup_logging_env(self)    
             self.log_file_fd = _logger
-        else:
-            self.log_dir = None     
-        
+        elif self.console == True:
```
xwings, line 136: this only work during startup? what if we want to turn on or off console during execution or within a hook?

qiling/core_utils.py
```diff
@@ -46,9 +44,9 @@ def nprint(self, *args, **kw):
 
             # setup filter for logger
             # FIXME: only works for logging due to we might need runtime disable nprint, it should be a global filter not only syscall
-            # if self.strace != None and self.output == QL_OUTPUT.DEFAULT:
-            #     from .utils import Ql_filter
-            #     self.log_file_fd.addFilter(ql_filter(self.filter))
+            if self.filter != None and self.output == QL_OUTPUT.DEFAULT:
+                from .utils import Ql_filter
+                self.log_file_fd.addFilter(Ql_filter(self.filter))
```
xwings, line 49: QL filter is a bad name, i admin i choose the name but we need a better one

qiling/core.py
ucgJhe: add a simple console checking will let you do it, turn off or on console logging by `ql.console = True` or `ql.console = False` , is this enough or we need more granular ?

ucgJhe: like this
```python
import sys
sys.path.append("..")
from qiling.os.posix.syscall.fcntl import ql_syscall_openat
from qiling import *

def my_openat(ql, *args):
    ql.console = False
    ql_syscall_openat(ql, *args)
    ql.console = True


if __name__ == "__main__":
    ql = Qiling(["examples/rootfs/x8664_linux/bin/x8664_hello"], "examples/rootfs/x8664_linux", console=True)
    ql.set_syscall("openat", my_openat)
    ql.run()

```

xwings: can you test it this way ?
```
def my_openat(ql, *args):
    ql.console = False
    ql,nprint("Shows nothing") 
    ql.console = True
    ql,nprint("Shows something")
```

ucgJhe: image

xwings: Nice

qiling/core_utils.py
xwings: Now we need to figure this out. maybe just call it filter

ucgJhe: how about that we made Ql_filter as a setup function for now, maybe the name will be change after other filters is implemented in the future

qiling/core.py
```diff
@@ -133,9 +133,11 @@ def __init__(
...
+            _logger = ql_setup_logging_stream(self)
+            self.log_file_fd = _logger
+            # self.log_dir = None
```
xwings, line 139: this need to be  out.

qiling/core_utils.py
```diff
@@ -46,13 +44,14 @@ def nprint(self, *args, **kw):
...
-            #     self.log_file_fd.addFilter(ql_filter(self.filter))
+            if self.filter != None and self.output == QL_OUTPUT.DEFAULT:
+                from .utils import ql_setup_filter
+                self.log_file_fd.addFilter(ql_setup_filter(self.filter))
 
             msg = args[0]
             msg += kw["end"] if kw.get("end", None) != None else os.linesep
-            fd.info(msg)
+            if self.console == True:
+                fd.info(msg)
```
xwings, line 54: One more question. 

if ql.console == False will it still write to the log ?

qiling/core.py
ucgJhe: you mean the commented part or whole section ?

xwings: yes comment one. unless there is a specific reason. if year please write it down

qiling/core_utils.py
xwings: I guess this is the last reqeust

Once we confirm this we are good to go.

ucgJhe: done it by add a FalseFilter for all stream handler when console=False, at the same time still write to disk logging file

ucgJhe: testing code:
```
#!/usr/bin/env python3
# 
import sys
sys.path.append("..")
from qiling.os.posix.syscall.fcntl import ql_syscall_openat
from qiling import *

def my_openat(ql, *args):
    ql.console = True
    ql.nprint("Shows something")
    ql.console = False
    ql.nprint("Shows nothing") 
    ql_syscall_openat(ql, *args)

if __name__ == "__main__":
    ql = Qiling(["examples/rootfs/x8664_linux/bin/x8664_hello"], "examples/rootfs/x8664_linux", console=False, log_dir='qlog')
    ql.set_syscall("openat", my_openat)
    ql.run()
```

ucgJhe: result from console:
image

ucgJhe: result from log:
image

xwings:
Nice, thanks for the work.

qiling/utils.py
```diff
@@ -194,4 +209,21 @@ def ql_setup_logging_file(ql_mode, log_file_path, logger=None):
...
+            super().__init__()
+            # accept list or string func_names so you can use it in qltool and programming
+            self.filter_list = func_names.strip().split(",") if isinstance(func_names, str) else func_names
+
+        def filter(self, record):
+            return any((record.getMessage().startswith(each) for each in self.filter_list))
+
+    return _filter(func_names)
+
+class _FalseFilter(logging.Filter):
```
xwings, line 225: the naming is not consistent. can we have a better naming ?

ucgJhe: how about merge it to ql_setup_filter ?

xwings: i guess that is a good idea

xwings:
Thanks man. Lets merge

PR #286, opened 2020-05-15
Hooks, structs and pep8

0ssigeno:


qiling/os/windows/dlls/kernel32/tlhelp32.py
```diff
@@ -27,7 +28,7 @@ def hook_CreateToolhelp32Snapshot(ql, address, params):
     # TODO thinking about implementing an handler, gonna see if is really necessary
     flag = params["dwFlags"]
     if flag == TH32CS_SNAPPROCESS:
-       ql.dprint(D_RPRT, "[=] The sample is checking every process!")
+        ql.dprint(D_RPRT, "[=] The sample is checking every process!")
```
xwings, line 31: Binary .. targeted process or somthing. the word sample is very generic to malware

qiling/os/windows/const.py
```diff
@@ -588,3 +587,37 @@
 # ...
 ThreadHideFromDebugger = 0x11
 # ...
+
+
+# https://docs.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messageboxw
+
+MB_ABORTRETRYIGNORE = 0x000000020
+MB_CANCELTRYCONTINUE = 0x000000060
```
xwings, line 595: does all these valiue works from 7 till 10 " maybe we need some additional comments if there are some version specific Windows

qiling/os/windows/dlls/kernel32/fileapi.py
```diff
@@ -303,7 +305,7 @@ def hook_GetVolumeInformationW(ql, address, params):
         pt_volume_name = params["lpVolumeNameBuffer"]
         if pt_volume_name != 0:
             # TODO implement
-            volume_name = ("AAAABBBB"+"\x00").encode("utf-16le")
+            volume_name = ("AAAABBBB" + "\x00").encode("utf-16le")
```
xwings, line 308: do we need this in config so user can change it ?

qiling/os/windows/dlls/kernel32/fileapi.py
```diff
@@ -392,10 +394,11 @@ def hook_CreateDirectoryA(ql, address, params):
...
         target_dir = os.path.join(ql.rootfs, lpPathName.replace("\\", os.sep))
         print('TARGET_DIR = %s' % target_dir)
-        ql.os.transform_to_real_path(lpPathName) 
+        ql.os.transform_to_real_path(lpPathName)
         # Verify the directory is in ql.rootfs to ensure no path traversal has taken place
-        if os.path.commonprefix(target_dir, ql.rootfs) != ql.rootfs:
-            ql.dprint(D_INFO, 'CreateDirectoryA attempted to create dir outside of rootfs: "%s". Not creating.' % (target_dir))
+        if os.path.commonprefix([target_dir, ql.rootfs]) != ql.rootfs:
+            ql.dprint(D_INFO,
+                      'CreateDirectoryA attempted to create dir outside of rootfs: "%s". Not creating.' % target_dir)
```
xwings, line 401: this technically never exits. we dont have to print. rootfs is the "OS" it should not escape anyway.

qiling/os/windows/dlls/user32.py
```diff
@@ -341,12 +404,13 @@ def hook_GetOEMCP(ql, address, params):
 def hook_LoadStringA(ql, address, params):
     dst = params["lpBuffer"]
     max_len = params["cchBufferMax"]
+    # TODO fix the string
     string = "AAAABBBBCCCCDDDD" + "\x00"
```
xwings, line 408: hardcoded by windows or we should allow user to change

xwings:
i did a refactor for fncc. i guess you need to do the same update.

qiling/os/windows/dlls/kernel32/tlhelp32.py
0ssigeno: c9e25f5

qiling/os/windows/dlls/kernel32/fileapi.py
0ssigeno: 7de3cd5

qiling/os/windows/dlls/kernel32/fileapi.py
0ssigeno: 15db38b

qiling/os/windows/dlls/user32.py
0ssigeno: Technically no, should not be hardcoded nor the user should change it. Every string is described from a uID, but i'cant find the documentation of which string has which id

qiling/os/windows/const.py
0ssigeno: I can be wrong, but i think every windows version support this

qiling/os/windows/dlls/user32.py
xwings: maybe we profile it. that will be a better way

xwings:
Thank and i shall press merge

PR #287, opened 2020-05-16
Fix UEFI GDB and get/set variable.

liba2k:


qiling/os/uefi/uefi.py
```diff
@@ -16,6 +16,7 @@ def __init__(self, ql):
         super(QlOsUefi, self).__init__(ql)
         self.ql = ql
         self.entry_point = 0
+        self.pid = 0
```
xwings, line 19: i think we got a OS wide pid. can u use that ?

xwings: my bad it was for posix. i will move it to QlOS so u can use it

xwings: Done moving, will be be able change it ?

liba2k: Done. I've added a small change in pe_uefi.

xwings:
cool, thanks !

PR #292, opened 2020-05-21
fix abnormal nprint behavior

ucgJhe:
as title.
now you can use nrpint() simply to print a newline but not nprint("")


xwings:
thanks !

PR #293, opened 2020-05-21
replace all nprint("\n") to nprint()

ucgJhe:
as title.


xwings:
we should fix nprint.

either ql.nprint() or ql.nprint("\n") should work the same

ucgJhe:
its a bit weird if they work as the same, since print('\n') and print('') is different

xwings:
perfect!

PR #294, opened 2020-05-21
Improve the check for existance of address mapping

0xPhoeniX:


xwings:
is the another PR by you  too ?

0xPhoeniX:
No, it is not. Mine addresses the implementation of mmap/mmap2 syscall.
The other one an issues in elf loader.

xwings:
my bad. so there are 3 mmap. mmap2 mmap and mmap old. i guess all of those also need the same check ?

xwings:
and u r in the wrong base. i changed to dev. so, can u fix the conflict ?

0xPhoeniX:
Will do.

0xPhoeniX:
https://github.com/qilingframework/qiling/pull/324
Includes all the changes.

PR #295, opened 2020-05-21
Fix incorrect check

3vilWind:


xwings:
which pr i should keep ?

xwings:
sorry i am blind, let me check this pr

xwings:
thanks !

PR #297, opened 2020-05-24
Some EFI_HANDLE related fixes for the UEFI emulation environment.

assafcarlsbad:
- Dereference the `Handle` parameter passed to `Install(Multiple)ProtocolInterface(s)`.
- Pass a valid `EFI_HANDLE` to the module's initialization routine.
- Make sure no module will occupy the NULL page.
- Correctly handle the case where the `EFI_HANDLE` passed to Install(Multiple)ProtocolInterface(s) is NULL.


xwings:
Hi,

Thanks for the update.

PR #300, opened 2020-05-25
Fix improper calls to read_cstring and argument check in _LCMapString

L1nkZ:
It seems some calls to `read_cstring` are made incorrectly.
I also pushed a fix for some invalid memory writes that could happen in `_LCMapString`.


xwings:
Howdy,

welcome to the community. thanks !

PR #303, opened 2020-05-26
Fix a broken link in README.md

abcSup:
I noticed that https://docs.qiling.io/en/latest/getting_started/ is broken and should be replaced with https://docs.qiling.io/en/latest/howto/


xwings:
Lovely, thanks !

PR #308, opened 2020-05-27
Now al-khaser is able to pass 2 more checks: interrupt 0x3 and 0x2d

0ssigeno:


qiling/os/windows/utils.py
```diff
@@ -46,3 +46,17 @@ def is_file_library(string):
 
 def string_to_hex(string):
     return ":".join("{:02x}".format(ord(c)) for c in string)
+
+
+def find_size_function(ql, func_addr):
+    # We have to retrieve the return address position
+    code = ql.mem.read(func_addr, 0x100)
+    if b"\xc3" in code:
```
xwings, line 54: loop or something ?since its all the same

0ssigeno: 3dd389f should resolve the issue

qiling/os/windows/utils.py
```diff
@@ -46,3 +46,11 @@ def is_file_library(string):
 
 def string_to_hex(string):
     return ":".join("{:02x}".format(ord(c)) for c in string)
+
+
+def find_size_function(ql, func_addr):
+    # We have to retrieve the return address position
+    code = ql.mem.read(func_addr, 0x100)
+    opcodes = [b"\xc3", b"\xc2", b"\xcb", b"\xca"]
```
xwings, line 54: opcode sounds funny. these are  interrupt handler ? or any proper name ?

0ssigeno: https://c9x.me/x86/html/file_module_x86_id_280.html
I think the name is correct

xwings: opcode is a type. but these opcode design for a purpose.

0ssigeno: I mean, what i'm trying to do with this function is to find the return opcode.
Maybe return_opcodes is a better name? Or as you wish, i don't really mind

xwings: Return from Procedure. we need a shorter name

xwings: from wikipedia
```
In computing, an opcode (abbreviated from operation code, also known as instruction machine code, instruction code, instruction syllable, instruction parcel or opstring) is the portion of a machine language instruction that specifies the operation to be performed
```

qiling/os/windows/dlls/kernel32/errhandlingapi.py
```diff
@@ -110,14 +101,45 @@ def hook_RaiseException(ql, address, params):
...
+        ql.mem.write(double_pointer, pointer.to_bytes(4, "little"))
+        # arg
+        ql.stack_push(double_pointer)
+        # ret
+        ql.stack_push(ret)
+        # func
+        ql.stack_push(start)
+
+    def exec_standard_into(ql, into, user_data):
+        # fixme probably this works only with al-khaser.
```
xwings, line 125: "FIXME:" not fixme

xwings:
So,

He did not comment. I will silently commit

PR #315, opened 2020-06-01
Align PE image size on a 4KiB boundary

L1nkZ:
`self.pe.OPTIONAL_HEADER.SizeOfImage` might not be a multiple of the page size (which is 4096 bytes on Windows).  
This commit fixes failing calls to `uc_mem_map` that would return `UC_ERR_ARG` for some executables.  
I'm not sure that this commit has side effects but  I'll wait for the reviews.


Dliv3:
Thanks for the PR, Merged!

PR #319, opened 2020-06-03
Updating al-khaser

0ssigeno:
- ntdll.dll was not loaded at the start. Every PE requires ntdll.dll to be loaded, even if is not in the import table. Thanks to that, is possible to modify GetModuleHandle to not load libraries anymore (that was a workaround, the syscall is not supposed to load libraries)
- Removed WSTRING_ADDR, STRING_ADDR, since is possible to update the parameter inside the syscall (and their use was already removed)
- Updated PEB with a new attribute (many more still require implementation)
- Added syscalls, structs, and constants
- Added the [HARDWARE] and [PROCESSES] configuration inside the Windows profile
- Updating end point of al-khaser, allowing to run every debug check. (and some sandbox checks)


xwings:
Hey,

I assume these are pretty al-khaser related API and we are testing it.  Should be all tested. 

Thanks !

0ssigeno:
Yeah, every API/struct implemented was tested inside al-khaser.

PR #330, opened 2020-06-10
Fixed structs.write() when the argument was a byte array.

0ssigeno:
Added debug print for struct, just for clarification.
Fixed lstrlenW behaviour when the argument were the argument was a pointer to bytes.


xwings:
I guess some sort of test always healthy.

qiling/os/windows/structs.py
```diff
@@ -293,6 +293,7 @@ def read(self, addr):
         raise NotImplementedError
 
     def generic_write(self, addr: int, attributes: list):
+        self.ql.dprint(D_INFO, "[+] Writing Windows object " + self.__class__.__name__)
```
xwings, line 296: all the content writing debug print we should move it to D_D_RPRT

0ssigeno:
Now gandcrab is executed with three different environment: 
administrator
user
administrator with russian keyboard

If the number of syscalls for each test is different, than we should be good, or at least, we have not broke something that is very important. For test each structure independently, i will make another PR

xwings:
cool. thanks

i will merge for now

PR #332, opened 2020-06-11
hook_mem_invalid --> hook_mem_valid.

assafcarlsbad:


xwings:
We must be blind, again thanks !

PR #334, opened 2020-06-13
export winsdk function info

kabeor:
usage: just from filename import filename_data


klks:
instead of exporting all of this as python code/dict, maybe a text/json file is more appropriate?
then the python code can read the files in and process them, can also have a lookup function.
all of this definitions can then come in as an extension to qiling and loaded as needed

xwings:
i created qiling/extensions maybe all these code can be in qiling/extensions/windows_sdk

kabeor:
@klks I'll convert all files to json later. Let's see will it work well.

qiling/os/freebsd/map_syscall.py
```diff
@@ -14,22 +14,476 @@ def map_syscall(ql, syscall_num):
...
+    "thr_set_name": (464),
+    "aio_fsync": (465),
+    "rtprio_thread": (466),
+    "sctp_peeloff": (471),
+    "sctp_generic_sendmsg": (472),
+    "sctp_generic_sendmsg_iov": (473),
+    "sctp_generic_recvmsg": (474),
+    "pread": (475),
+    "pwrite": (476),
+    "*mmap": (477),
```
xwings: this should be "mmap2"

qiling/os/freebsd/map_syscall.py
```diff
@@ -14,22 +14,476 @@ def map_syscall(ql, syscall_num):
...
+    "aio_suspend": (315),
+    "aio_cancel": (316),
+    "aio_error": (317),
+    "aio_read": (318),
+    "aio_write": (319),
+    "lio_listio": (320),
+    "yield": (321),
+    "mlockall": (324),
+    "munlockall": (325),
+    "__getcwd": (326),
```
xwings: this should be "getcwd"

qiling/os/freebsd/map_syscall.py
```diff
@@ -14,22 +14,476 @@ def map_syscall(ql, syscall_num):
...
-    "bind": (0x68),
-    "listen": (0x6A),
-    "sysarch": (0xA5),
-    "_sysctl": (0xCA),
-    "clock_gettime": (0xE8),
-    "issetugid": (0xFD),
-    "getcwd": (0x0146),
-    "mmap2": (0x01DD),
+    "nosys": (0),
+    "sys_exit": (1),
```
xwings: this should be "exit"

PR #336, opened 2020-06-13
Introduce sanitized heap.

assafcarlsbad:
Inspired by the sanitized heap implementation from BaseSAFE: https://github.com/fgsect/BaseSAFE/blob/master/unicornafl_bindings/src/utils.rs.

Clients can enable the sanitized heap using the following snippet:
```
ql.loader.heap = qiling.os.memory.QlSanitizedMemoryHeap(ql, ql.loader.heap)
ql.loader.heap.oob_handler = my_oob_handler
ql.loader.heap.bo_handler = my_bo_handler
ql.loader.heap.bad_free_handler = my_bad_free_handler
ql.loader.heap.uaf_handler = my_uaf_handler
```


xwings:
Hi,

Will you be able to help on updating docs.qiling.io  too ?

aquynh:
another approach is to integrate this santinizer class into the existing Heap class, and user can choose to turn santinizer on if needed.

assafcarlsbad:
@xwings I built a vulnerable UEFI driver which has all sorts of memory corruption bugs in it. Do you want me to share it with you?

@aquynh That's indeed a possibility, but I thought that separating the sanitized heap from the standard heap is the correct way to go. If you think otherwise let me know and I'll modify the code accordingly.

aquynh:
By default the santinizer is off, so it is ok to me ;-)

xwings:
i think we build into heap class? and not a independent class on its own ?

xwings:
Hi, 

any update ? :)

assafcarlsbad:
Moved the sanitized heap under 'extensions' and added an example script.

xwings:
Hey, thanks !

PR #341, opened 2020-06-15
fix print out for multithread qiling instance

ucgJhe:
as title.


qiling/os/linux/thread.py
```diff
@@ -47,19 +47,13 @@ def __init__(self, ql, thread_management = None, start_address = 0, context = No
         self.log_file_fd = None
 
         # FIXME: @ucgJhe for Temporay FIX
```
xwings, line 49: if this is fixed, please remove this line

xwings:
Thanks !

PR #343, opened 2020-06-17
Add MacOS and IOS Syscall

kabeor:


xwings:
Thanks, i will do some file moving and update the table

PR #344, opened 2020-06-17
Add windows_sdk json file

kabeor:
{
   " function_name": [
         {
              "name": xxxx,
              "type": "int"
         }
    ]
}


klks:
can the json files be put into a folder, maybe call it definitions of defs. the parser script for this can then live in the windows_sdk folder. we should also branch this into a feature and once its complete, merge it back into dev.

kabeor:
Totally understand. I'm trying to make json file more complete and let fncc.py use it.

xwings:
new base now

xwings:
i will merge now and lets work on the CC

PR #345, opened 2020-06-18
Adding support for dynamic load of modules.

liba2k:
Also adds a small fix for configuration table.


xwings:
Cool, thanks !

liba2k:
That was quick, Thank you!

PR #346, opened 2020-06-20
add new feature: fs_mapper hijacking

ucgJhe:
add new feature fs_mapper hijacking and corresponding example

### DIFF
--- a/examples/hello_x86_linux_fake_urandom.py
+++ b/examples/hello_x86_linux_fake_urandom.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python3
+# 
+# Cross Platform and Multi Architecture Advanced Binary Emulation Framework
+# Built on top of Unicorn emulator (www.unicorn-engine.org) 
+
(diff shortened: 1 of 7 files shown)


w1tcher:
This is a good idea. But I think add_fs_mapper should pass in a class, not an instance. 
https://github.com/qilingframework/qiling/pull/346/commits/1e500000df025b4f232917db59ae35057f04906a#diff-24f7844cea3bdc52061cf711e422fedbR31
We can change this here.
```python
        elif not isinstance(open_path, str):
            return open_path()
```

ucgJhe:
I did have a nvram example for this feature, and it does need object to handle read/write operation and so on , thats why I use instance but not class here.

ucgJhe:
I did have a example for nvram , by that case read/write operation will be handled and I need instance to keep the result.

w1tcher:
OK

w1tcher:
OK, Thanks.

PR #347, opened 2020-06-21
convert ql.heap to ql.os.heap

lopqto:
**gethostbyname** and **RtlAllocateHeap** were still using `ql.heap` and I changed it to `ql.os.heap`


xwings:
Cool. thanks !

PR #349, opened 2020-06-23
use space for nprint seperator as default option

ucgJhe:
as title.
so we can use ql.nprint(1, 2, 3, 4) now.


xwings:
what is  ql.nprint(1, 2, 3, 4) ?

ucgJhe:
> what is ql.nprint(1, 2, 3, 4) ?

`ql.nprint(1, 2, 3, 4)` you'll get "1 2 3 4" instead of "1234"

before this PR you can not use comma to separate input message

xwings:
understand. thanks

PR #350, opened 2020-06-23
add vsyscall table in elf runtime memory

ucgJhe:
as title. so we can run hello world golang binary now.
if golang binary exmaple should be included in this PR please let me know.

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -628,11 +628,11 @@ class QlLoaderELF(QlLoader, ELFParse):
                 self.ql.mem.map(_vsyscall_addr, _vsyscall_size, info="[vsyscall]")
                 self.ql.mem.write(_vsyscall_addr, _vsyscall_size * b'\xcc')
 
-                _vsyscall_entry_asm = [0x60c0c748, # gettimeofday
-                                       0xc9c0c748, # time
+                _vsyscall_entry_asm = [0x60c0c748, # mov rax, 0x60; syscall gettimeofday
+                                       0xc9c0c748, # mov rax, 0xc9; syscall time
                                        ]
 
                 for idx, val in enumerate(_vsyscall_entry_asm):
-                    self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+0) * 4, self.ql.pack32(val))        # entry
+                    self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+0) * 4, self.ql.pack32(val))        # insert entry
                     self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+1) * 4, self.ql.pack32(0x0f000000)) # syscall
-                    self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+2) * 4, self.ql.pack32(0xccccc305)) # return
+                    self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+2) * 4, self.ql.pack32(0xccccc305)) # ret
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -618,20 +618,21 @@ class QlLoaderELF(QlLoader, ELFParse):
         self.ql.os.function_hook = FunctionHook(self.ql, self.elf_phdr + mem_start, self.elf_phnum, self.elf_phent, load_address, load_address + mem_end)
 
         # map vsyscall section for some specific needs
-        _vsyscall_addr = 0xffffffffff600000
-        _vsyscall_size = 0x1000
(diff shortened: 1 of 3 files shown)


qiling/loader/elf.py
```diff
@@ -615,4 +615,23 @@ def load_with_ld(self, stack_addr, load_address = -1, argv = [], env = {}):
         self.new_stack = new_stack
         self.load_address = load_address
         self.images.append(self.coverage_image(load_address, load_address + mem_end, self.path))
-        self.ql.os.function_hook = FunctionHook(self.ql, self.elf_phdr + mem_start, self.elf_phnum, self.elf_phent, load_address, load_address + mem_end)
\ No newline at end of file
+        self.ql.os.function_hook = FunctionHook(self.ql, self.elf_phdr + mem_start, self.elf_phnum, self.elf_phent, load_address, load_address + mem_end)
+
+        # map vsyscall section for some specific needs
+        _vsyscall_addr = 0xffffffffff600000
```
xwings, line 621: this address is very "x8664 ?"

qiling/loader/elf.py
```diff
@@ -615,4 +615,23 @@ def load_with_ld(self, stack_addr, load_address = -1, argv = [], env = {}):
...
+            self.ql.mem.map(_vsyscall_addr, _vsyscall_size, info="[vsyscall]")
+            self.ql.mem.write(_vsyscall_addr, _vsyscall_size * b'\xcc')
+
+            _vsyscall_entry_asm = [0x60c0c748, # gettimeofday
+                                   0xc9c0c748, # time
+                                   ]
+
+            for idx, val in enumerate(_vsyscall_entry_asm):
+                self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+0) * 4, self.ql.pack32(val))        # entry
+                self.ql.mem.write(_vsyscall_addr + idx * 0x400 + (idx+1) * 4, self.ql.pack64(0x0f000000)) # syscall
```
xwings, line 636: any reason to fix all these address ?

ucgJhe: vsyscall table should be looks like this in memory
image

aquynh:
The question is why hardcode them in the loader?

qiling/loader/elf.py
ucgJhe: you are right , its only for x8664 elf binary

ucgJhe:
> The question is why hardcode them in the loader?

so we can run golang binary without extra setup

any advice is welcome

aquynh:
I mean these addresses never change?

ucgJhe:
> I mean these addresses never change?

I check  it here ref: http://lxr.linux.no/linux-bk+v2.5.5/include/asm-x86_64/vsyscall.h#L10
depends on how linux kernel do it , i will check other version make sure its fixed

aquynh:
At least use profile for the hardcode values, as always.

ucgJhe:
> At least use profile for the hardcode values, as always.

got it, ill move it to profile 

thanks

qiling/loader/elf.py
```diff
@@ -615,4 +615,24 @@ def load_with_ld(self, stack_addr, load_address = -1, argv = [], env = {}):
...
+            _vsyscall_addr = int(self.ql.os.profile.get("OS64", "vsyscall_address"), 16)
+            _vsyscall_size = int(self.ql.os.profile.get("OS64", "vsyscall_size"), 16)
+
+            if not self.ql.mem.is_mapped(_vsyscall_addr, _vsyscall_size):
+                # initialize with \xcc then insert syscall entry
+                # each syscall should be 1KiB(0x400 bytes) away
+                self.ql.mem.map(_vsyscall_addr, _vsyscall_size, info="[vsyscall]")
+                self.ql.mem.write(_vsyscall_addr, _vsyscall_size * b'\xcc')
+
+                _vsyscall_entry_asm = [0x60c0c748, # gettimeofday
```
xwings, line 631: i assume this is opcode and can we use ql.compile here ?

ucgJhe: its doable , but I found that the result of ql.compile need to be adjusted to fit the size, will make it more complicated

xwings: try if possible. so code will be more readable. else stick to this is fine.

ucgJhe: i did try , doesnt look good to me. I recommend stick to this , and i will comment the asm after each one, is that acceptable ?

ucgJhe: just figured out how to use ql.compile correctly, it should be more readable now.

xwings:
ok, now we support golang binary. cool!

PR #367, opened 2020-07-05
RFC: Add experimental support for 8086 and dos

wtdcode:
To have a better understanding of the whole project, I try to add the support for 8086 and DOS. 

Current status:
- [x] arch
    - [x] 8086
- [ ] loader
    - [x] COM support
    - [ ] DOS executable support (reading dosbox source...)
    - [ ] mapping support if any
- [ ] os
     - [ ] Full syscall support
- [ ] examples
    - [x] COM
    - [ ] DOS executable

Here I get some questions requesting for comments.

1. I add a new `arch` named `A8086` but I'm not sure whether it should have been included in `x86`. if so, many codes in qiling need to be refactored. Note that `UC_MODE_16` is used with `UC_ARCH_X86`.
2. I notice that some syscalls in 8086 are related to the cursor. Shall we create a special terminal e.g utilizing `ncurses` to support such syscalls?


xwings:
Hey, 

this is cool and indeed very cool.

I do remember Quynh wrote one something on Unicorn thats runs dos. You might want to take a look.

https://github.com/aquynh/unidos

xwings:
move to new base

wtdcode:
See #439

PR #368, opened 2020-07-07
Support for running MIPS on MacOS & fixed a issue

kabeor:
Support for running MIPS on MacOS
fixed issue "UnboundLocalError: local variable 'f' referenced before assignment"


qiling/os/posix/const_mapping.py
```diff
@@ -117,10 +120,13 @@ def flag_mapping(flags, mapping_name, mapping_from, mapping_to):
             f = mac_open_flags
             t = linux_open_flags
 
-    elif ql.archtype== QL_ARCH.MIPS and ql.platform == QL_OS.LINUX:
-        f = mips_open_flags
-        t = linux_open_flags
-
+    else:
```
xwings, line 123: we need to make sure its QL_OS.LINUX and QL_ARCH.MIPS to avoid potential problem

xwings:
cool, thanks

PR #369, opened 2020-07-07
Support EFI_FIRMWARE_VOLUME2_PROTOCOL.

assafcarlsbad:
Some UEFI drivers such as `PcdDxe` on my system uses the `EFI_FIRMWARE_VOLUME2_PROTOCOL` in order to read specific sections from the BIOS image. This PR is a preliminary attempt to implement this protocol on top of the excellent [uefi-firmware-parser](https://github.com/theopolis/uefi-firmware-parser) library.


xwings:
looks like this PR  affect code coverage ? will you be able to fix it ?

assafcarlsbad:
On it.

xwings:
If we did not use too many code from uefi_firmware , can we write the code on our own ?

assafcarlsbad:
I'm afraid `uefi_firmware` does a lot of low-level parsing under the hood. We only need it to support fuzzing of entire firmware volumes rather than standalone modules. If this use case is too specialized (and I can agree with such statement), than we can leave it outside Qiling altogether and make it a part of the fuzzer instead.

xwings:
Ok.

There are few reasons. Mainly i want to keep installation module as less as possible unless is required for the core.

Any more options ?

PR #370, opened 2020-07-08
support for architecture-specific IoT program

aaaddress1:
some malware binary built for uncommon IoT operation system (ARM+Linux) but osabi (ident[7]) might be 64 ~ 255. I face a Mirai varieties with osabi = 0x97. I think detect magic of ELF header is stable enough to assert it's a ELF file.

ref: https://refspecs.linuxfoundation.org/elf/gabi4+/ch4.eheader.html#elfid

### DIFF
--- a/qiling/loader/utils.py
+++ b/qiling/loader/utils.py
@@ -43,12 +43,11 @@ def ql_elf_check_archtype(path):
         osabi = ident[0x7]
         e_machine = ident[0x12:0x14]
 
-        if osabi == 0x11 or osabi == 0x03 or osabi == 0x0:
-            ostype = QL_OS.LINUX
-        elif osabi == 0x09:
+        if osabi == 0x09:
             ostype = QL_OS.FREEBSD
         else:
-            ostype = None
+        #if osabi == 0x11 or osabi == 0x03 or osabi == 0x0:
+            ostype = QL_OS.LINUX
 
         if e_machine == b"\x03\x00":
             archendian = QL_ENDIAN.EL


qiling/loader/utils.py
```diff
@@ -43,12 +43,11 @@ def getident():
...
         e_machine = ident[0x12:0x14]
 
-        if osabi == 0x11 or osabi == 0x03 or osabi == 0x0:
-            ostype = QL_OS.LINUX
-        elif osabi == 0x09:
+        if osabi == 0x09:
             ostype = QL_OS.FREEBSD
         else:
-            ostype = None
+        #if osabi == 0x11 or osabi == 0x03 or osabi == 0x0:
```
xwings, line 49: what is the osabi value did you see ? maybe we can add into const or somewhere

aaaddress1: I got a osabi value 0x78 from a malware binary (Mirai), and I found this happends on multiple binary samples of ARM based Dlink Linux.

aaaddress1: I could send you guys those specific smaples if needed.

xwings: Will you be able to put a additional info on the comment out section by adding 0x97 is needed.

xwings: Or another option is make it osabi in () i guess its better.

xwings: I will merge and comment

xwings:
and Thanks !

PR #372, opened 2020-07-09
Add test option to skip malware

learn-more:


xwings:
This could be trivia, the reason why these malware is there is due to these malware comes with those Windows API we need to test. So skipping it will make these test kind of "work" but we will lost lots of test cases.

learn-more:
> This could be trivia, the reason why these malware is there is due to these malware comes with those Windows API we need to test. So skipping it will make these test kind of "work" but we will lost lots of test cases.

I understand that they are important for regression testing, but I don't feel like unpacking those locally and testing with them.
This is why I added a new option where you can disable these tests with a new environment variable.
This is just to make locally testing this easier, without having to modify code for it.

tests/test_pe.py
```diff
@@ -6,6 +6,7 @@
 import sys, random, unittest
 import string as st
 from binascii import unhexlify
+import os
```
xwings, line 9: import os can join in to the upper part import. alphabetical order of coz

tests/test_pe.py
```diff
@@ -33,13 +34,17 @@ def test_pe_win_x86_hello(self):
 
 
     def test_pe_win_x86_uselessdisk(self):
+        if 'QILING_SKIP_MALWARE' in os.environ:
```
xwings, line 37: maybe a different name? QL_FAST_TEST or something.

tests/test_pe.py
learn-more: Nothing in the first section is alphabetically ordered?
Where do you want the os to go exactly?

xwings: import os, random, sys, unittest

should be like that.

learn-more: Thanks.
I tried to follow the code around it, without making any unrelated changes, hence why I did not consider reordering them all :)

xwings:
Cool. Thanks !

PR #374, opened 2020-07-09
Move the creation of a disassembler to a new function

learn-more:
This makes it easier when using qiling to create a disassembler for the correct architecture,
without the need of having to copy that code into your own project.


xwings:
Cool. 

Thanks !

PR #376, opened 2020-07-12
Print offending module name + offset in emu_error().

assafcarlsbad:
Yields a more friendly representation of the crashing PC, e.g: `[+] PC = 0x2071e (/tmp/rom2/PcdDxe/section0.pe+0x71e)`.


xwings:
Perfect. Thanks !

PR #378, opened 2020-07-15
fix MIPS BE and packing function in ql

ucgJhe:
now pack64 and unpack64 will done its job correctly, and fix MIPS big endian pagesz


xwings:
Cool. Now we got working EB !

PR #379, opened 2020-07-15
When available, print the module name as part of show_mapinfo().

assafcarlsbad:
Yields a more user-friendly output, e.g.:
`[+] 01540000 - 01543000 - rwx    [mapped] (/tmp/rom2/PciOutOfResourceSetupPage/section0.pe)`


xwings:
thanks !

PR #384, opened 2020-07-20
Adding DLL images for PE coverage logging.

jershmagersh:
We also want to track DLLs when tracing Portable Executables. If you need anything else added to this please let me know!


xwings:
Cool. Thanks !

PR #386, opened 2020-07-22
make syscall munmap sync changes to disk file if its MAP_SHARED

ucgJhe:
try do decrypt a firmware and encounter this missing part, now syscall mnumap will sync changes to disk file


xwings:
Hey,

thanks !

PR #395, opened 2020-07-24
Add sprintf call from msvcrt.dll

W3ndige:


xwings:
Hi,

First of all welcome to the community.

there is another hook_sprintf in user32.py will you be able to check and see is this a repeating Windows API ?

W3ndige:
Yes! Both user32.dll and msvcrt.dll have sprintf function with the main difference of the first one being Windows function and the second one is from the C standard library of C++ runtime.

xwings:
cool. do u have any binary so we can test both ?

W3ndige:
It is actually being used by the wannacry binary. To get a call to this function you have just to firstly create some quick hooks (returning some bogus information) of services function (CreateService, Start Service, OpenSCManager, etc), which by the way I plan to implement in the future.

After that there will be a call to sprintf from msvcrt.dll library.

xwings:
Cool. thanks. I will merge it now

PR #397, opened 2020-07-25
Ret main

learn-more:
Currently, qiling raises errors when main returns (on windows PE binaries),
however on windows this gracefully exits the current thread.

To emulate this behavior, a sentinel is allocated, and the address to this is stored on the stack when emulating a windows executable.
Since threading does not seem complete yet, for now emulation just halts when the sentinel is hit.

Additionally, the sentinel address is stored in the `loader`, so that it can be used for other exit points as well

32 and 64 bit testcases are included.


aquynh:
we need this for all platforms/architectures, not just windows/x86.

can you make this more generic?

learn-more:
I can try to add a few more targets, but at least macho / android is not gonna be possible,
and I am not sure about arm.

learn-more:
Trying to compile a binary like this on linux causes a segfault when running it,
so I'd rather leave the other implementations to someone more familiar with that platform.

learn-more:
http://www.muppetlabs.com/~breadbox/software/tiny/teensy.html also describes this behavior, so whilst this is valid behavior on windows,
it does not seem to be valid on linux.

aquynh:
We may emulate from anywhere, not just from entry point. This is why we
want this for not only Windows, but also other platforms.

learn-more:
In that case it would probably make sense to stop emulating if the stack pointer passes the stack pointer when emulation starts?

qiling/loader/pe.py
```diff
@@ -325,6 +332,11 @@ def load(self):
         self.ql.nprint("[+] Initiate stack address at 0x%x " % self.stack_address)
         self.ql.mem.map(self.stack_address, self.stack_size, info="[stack]")
 
+        self.EXIT_TRAP_ADDR = self.ql.os.heap.alloc(8)
+        # write nop to THREAD_RET_ADDR
+        self.ql.mem.write(self.EXIT_TRAP_ADDR, b"\x90"*8)
```
aquynh, line 337: i think you dont need to write anything to the trap address, as callback make emulation stop without exec anything there.

learn-more: Indeed.
The address should be valid (since the `ret` is evaluated), but the contents of the address do not matter.

aquynh:
even in the current implementation, this trap mechanism is not really Windows specific.

maybe you can create a new API for Qiling class, so whoever need it can call it.

aquynh:
since you mentioned ARM, actually you can implement this to be arch-independent, too (since writing NOPs to trap memory is not needed).

learn-more:
I have added 2 possible implementations:
* One is using `hook_code` - This checks the stack pointer on each instruction, and when it goes negative, it aborts.
* One is using `hook_address` - This checks if the sentinel address is hit, and aborts when it is.

Which method do you prefer?
I'll remove the other one (and cleanup leftover debug commits)

aquynh:
you can hook_address at the trap address, and stop emulation in the callback. hook_code and check stack pointer would have many false positives.

aquynh:
actually these 2 techniques can complement each other. i am thinking about keeping both, but unsure what is the best way to use either, or combine them.

qiling/core.py
```diff
@@ -177,15 +178,45 @@ def __init__(
         # Run the loader
         self.loader.run()
 
+        # Check for negative stack pointer:
+        self.hook_code(self._check_sp)
+        self.EXIT_TRAP_ADDR = self.mem.map_anywhere(0x1000)
```
aquynh, line 183: why 0x1000 here?

learn-more: no reason

learn-more:
> actually these 2 techniques can complement each other. i am thinking about keeping both, but unsure what is the best way to use either, or combine them.

Could add an option so it can be configured to use either, both or none?

learn-more:
`Linux Simple Crackme Brute Force, This Will Take Some Time ...` seems to fail now

xwings:
> > actually these 2 techniques can complement each other. i am thinking about keeping both, but unsure what is the best way to use either, or combine them.
> 
> Could add an option so it can be configured to use either, both or none?

u can ql.profile it.

learn-more:
> > > actually these 2 techniques can complement each other. i am thinking about keeping both, but unsure what is the best way to use either, or combine them.
> > 
> > 
> > Could add an option so it can be configured to use either, both or none?
> 
> u can ql.profile it.

Which section would you recommend?

aquynh:
This should belong to API, but not profile

xwings:
@learn-more will you be able to fix the conflict and the additional request. i guess we are good to merge it ?

learn-more:
I am working on a better implementation that is configurable

xwings:
cool!

learn-more:
I have updated the PR (finally!)
Now there are testcases for PE and ELF files, both x86 and x8664, and testing both options.

Some special workaround for the ELF loader was needed, since the bootstrap did do some funky stuff with the stack.
I left this in a separate commit so it is clear what is the actual functionality, and what is the ELF workaround.

wtdcode:
xwings assigned me to this PR. I will leave a review these days.

qiling/core.py
```diff
@@ -670,7 +688,46 @@ def enable_lib_patch(self):
...
-    
+
+    def _init_stop_guard(self):
+        if not self.stop_options.any:
+            return
+
+        # Allocate a guard page, we need this in both cases
+        # On a negative stack pointer, we still need a return address (otherwise we end up at 0)
+        # Make sure it is not close to the heap (PE), otherwise the heap cannot grow
+        self.EXIT_TRAP_ADDR = self.mem.find_free_space(0x1000, min_addr=0x9000000, alignment=0x10)
```
wtdcode, line 699: This variable doesn't conform to our naming convention. Is it from Win32 API?

qiling/core.py
```diff
@@ -670,7 +688,46 @@ def enable_lib_patch(self):
                 self.mem.write(self.mem.get_lib_base(filename) + addr, code)
             except:
                 raise RuntimeError("Fail to patch %s at address 0x%x" % (filename, addr))
-    
+
+    def _init_stop_guard(self):
```
wtdcode, line 692: Why not put this function to utils.py?

qiling/loader/loader.py
```diff
@@ -14,6 +14,7 @@ def __init__(self, ql):
         self.argv   = self.ql.argv
         self.images = []
         self.coverage_image = namedtuple('Image', 'base end path')
+        self.skip_exit_check = False
```
wtdcode, line 17: Could we have a property for this?

wtdcode: I know some other members are not properties since the refactor is still going on.

tests/test_pe.py
```diff
@@ -182,6 +182,46 @@ def test_pe_win_x8664_fls(self):
         del ql
 
 
+    def test_pe_win_x86_return_from_main_stackpointer(self):
+        """
+            Sample binary compiled with tcc: `tcc return_main.c -nostdlib -o return_main.exe`
+return_main.c:
+    int _start()
```
wtdcode, line 189: Sorry I'm still not so clear whether this program behaves differently on win32 and posix systems?

tests/test_pe.py
```diff
@@ -182,6 +182,46 @@ def test_pe_win_x8664_fls(self):
         del ql
 
 
+    def test_pe_win_x86_return_from_main_stackpointer(self):
+        """
+            Sample binary compiled with tcc: `tcc return_main.c -nostdlib -o return_main.exe`
+return_main.c:
```
wtdcode, line 188: You need to add this file to `examples/src`.

qiling/core.py
```diff
@@ -670,7 +688,46 @@ def enable_lib_patch(self):
...
+
+        # Allocate a guard page, we need this in both cases
+        # On a negative stack pointer, we still need a return address (otherwise we end up at 0)
+        # Make sure it is not close to the heap (PE), otherwise the heap cannot grow
+        self.EXIT_TRAP_ADDR = self.mem.find_free_space(0x1000, min_addr=0x9000000, alignment=0x10)
+        self.mem.map(self.EXIT_TRAP_ADDR, 0x1000, info='[Stop guard]')
+
+        # Stop on a negative stack pointer
+        if self.stop_options.stackpointer:
+            def _check_sp(ql, address, size):
```
wtdcode, line 704: Why this would happen? Qiling fail to detect programs have existed on Windows for the case like `return_main.c`? Am I right?

qiling/core.py
wtdcode: Another question: is there any way to avoid hard-coded value? I'm not sure.

qiling/core.py
```diff
@@ -670,7 +688,46 @@ def enable_lib_patch(self):
...
+                if not ql.loader.skip_exit_check:
+                    sp = ql._initial_sp - ql.reg.arch_sp
+                    if sp < 0:
+                        logging.info('Process returned from entrypoint (stackpointer)!')
+                        ql.emu_stop()
+
+            self.hook_code(_check_sp)
+
+        # Stop when running to exit trap address
```
wtdcode, line 713: Why programs may run to this address?

qiling/core.py
```diff
@@ -670,7 +688,46 @@ def enable_lib_patch(self):
...
+        if self.stop_options.exit_trap:
+            def _exit_trap(ql):
+                logging.info('Process returned from entrypoint (exit_trap)!')
+                ql.emu_stop()
+
+            self.hook_address(_exit_trap, self.EXIT_TRAP_ADDR)
+
+    def write_exit_trap(self):
+        self._initial_sp = self.reg.arch_sp
```
wtdcode, line 722: Maybe this member should be put to `ql.arch` or `ql.reg`?

qiling/core.py
learn-more: another way would be to make `find_free_space` work differently.
In the current implementation it allocates stuff right next to each other, and in the case of a PE file that means allocating stuff right next to the heap.
This means that the heap cannot grow.

This name can be changed to anything, it is not related to winapi.
I'll change it to lowercase.

qiling/core.py
learn-more: no reason, do you want me to move it there?

qiling/loader/loader.py
learn-more: Sure, I assumed this was only for user-facing properties.
(Did you guys measure the performance impact of the annotated functions vs just setting variables?)

qiling/core.py
learn-more: Yes, on windows it is perfectly valid for an application to return from main.
This means that the main thread will be terminated by the OS.

In Qiling what happens is an invalid memory access, since there is nothing (or garbage) on the stack where main will return to.

qiling/core.py
learn-more: See my other comment.

qiling/core.py
learn-more: It is not really a register or anything, it is more of a sentinel we need to keep track of to abort execution.
I could move it if you really want to, but in my opinion where it is now is a better place.

qiling/core.py
learn-more: To clarify: with 'main' I did not mean the function usually called by the C runtime, but the actual entry point of the application (pointed to by the field AddressOfEntrypoint)

qiling/core.py
wtdcode: okay, I think `_exit_trap_addr` shall work.

qiling/core.py
wtdcode: Since it's also one of the core functions (start/stop etc..), it's okay to leave it here.

qiling/loader/loader.py
wtdcode: Okay, let me do that when I refactor this part.

wtdcode:
Checklists:

- [ ] Update ChangeLog.
- [ ] Move example sources to `examples/src`.
- [ ] `ql._exit_trap_check`

wtdcode:
Merged finally. Thanks!

PR #399, opened 2020-07-26
fix syscall hook

ucgJhe:
fix syscall hook, now we can hook first and last syscall correctly, and remove redundant parentheses under dump mode


xwings:
thanks buddy

PR #403, opened 2020-07-28
Add a workaround for running linux binaries on windows

learn-more:
The 'resource' library does not exist for windows,
so this adds a dummy shim.


xwings:
not only this. we might need to do fcntl too

learn-more:
I did not get that far, this one bombed out trying to import so I made a quick workaround.

xwings:
cool, one step at a time.

thanks !

PR #408, opened 2020-07-30
New feature: saving and restoring file descriptor capability

ucgJhe:
merge this for capability of saving and restoring file descriptors during runtime


ucgJhe:
ql.save will save all(memory, register, fds) as default

xwings:
one more request.

can u please add ql.save and ql.restore to test, 

there is a partial_execution test. maybe we can replace that with the latest ql.save and ql.restore ?

xwings:
and .. i mess up some of the code. maybe you need t fix it too

xwings:
cool.
now we can test file des too

thanks!

PR #409, opened 2020-07-30
makes ql.save to disk file and also ql.restore from disk file

ucgJhe:
as title.


ucgJhe:
close for now , until we solve some issue

xwings:
nice, i think we still have some issue with cpu_context when saving to a file. but it can be tune later.

PR #410, opened 2020-07-30
Fix ql_syscall__newselect: correctly handle NULL as timeout

alfink:
> If timeout is specified as NULL, select() blocks indefinitely waiting for a file descriptor to become ready.  

previously, qiling dereferenced the null pointer. Example:

```
ql_syscall__newselect(4, 7ff3cb00, 0, 0, 0)
[!] Syscall ERROR: ql_syscall__newselect DEBUG: Invalid memory read (UC_ERR_READ_UNMAPPED)
```


xwings:
cool, thanks !

PR #412, opened 2020-07-31
Fix interpreter path in mapping

wtdcode:
Show the full path of the interpreter.

e.g.: We should show `{rootfs}/lib/ld.so`, not `/lib/ld.so`.


xwings:
welcome and thank you !

PR #413, opened 2020-07-31
Fix for idapython

wtdcode:
Fix environment hack by IDAPython.

See the comments for details.


xwings:
cool, now lets move to the real magic

PR #418, opened 2020-08-03
Put two plugins together

wtdcode:
To be integrated.


xwings:
merge for now. need more refactoring due to lots of repeating function

PR #420, opened 2020-08-04
Replace ql.os.file_des with ql.os.fd in ql_syscall_getsockname

alfink:


xwings:
ok, this is our fault.

thanks !

PR #421, opened 2020-08-04
Fix: ql_syscall_wait4 wstatus may be null

alfink:
man wait4:

> Similarly, the following wait4() call:
> 
>        wait4(pid, wstatus, options, rusage);
> 
> is equivalent to:
> 
>        waitpid(pid, wstatus, options);

man waitpid:
> If wstatus is not NULL, wait() and waitpid() store status information in the int to which it points.


xwings:
cool, thanks

PR #423, opened 2020-08-05
Fix: return the actual number of bytes sent in send()

alfink:
ql_syscall_send should return the number of bytes that have actually been sent, which can be less than send_len.


xwings:
hey, 

thanks !

PR #428, opened 2020-08-07
fix debug_print_stack method parameter inconsistency

newthis:
The original local variable i is not correct in debug_print_stack_method.


xwings:
Hi,

Welcome to Qiling and Thanks !

PR #429, opened 2020-08-07
Let Linux run with Windows!

wtdcode:
Two bugs are fixed in order to make Linux run on Windows.

1. The transformation between emulated path and real path is a mess

2. open() flags mapping left out Windows

Some refactors need to be done for `transform_to_real_path` and `transform_to_link_path`.


wtdcode:
Besides, I'd like to propose the elimination of `except:` and `except Exception`, which makes debugging painful and reveals real bugs.

xwings:
thanks

PR #432, opened 2020-08-09
Feature: add Qdb as built-in debugger for Qiling Framework

ucgJhe:
a gdb-like built-in debugger for Qiling Framework, support step-by-step execution so you can do manual check the runtime state of you target easily
its only MIPS supported for new, the rest were on the way.


ucgJhe:
attach Qdb during the runtime
image

aquynh:
What are differences between this and the current gdbserver?

ucgJhe:
> What are differences between this and the current gdbserver?

no more tricky setup between server and client and less effort to make it work as a simple debugger, also somehow provide better UX than `output="debug"`

aquynh:
Please have more details in README

xwings:
Nice Job!

PR #437, opened 2020-08-12
keep up qdb with Qdb and move README.md to docs.qiling.io

ucgJhe:
as title.


aquynh:
Later this may be moved to the core, not in extensions/

ucgJhe:
sure, any suggestion for that ? i can move it  within this PR.

aquynh:
i am not sure. in my opinion, debugger/ should not be in extensions/, either.

aquynh:
but it makes more sense if qdb/ is under debugger/ directory.

xwings:
now my question is. we have ql.qdb == True and ql.debugger == True.

We should make ql.debugger = True (this will be qdb by default) , else gdb will be ql.debugger = gdb:127.0.0.1:9999

xwings:
> but it makes more sense if qdb/ is under debugger/ directory.

and yes, to extentensions/debugger for now until we decide what to do

aquynh:
> 
> 
> now my question is. we have ql.qdb == True and ql.debugger == True
> 
> We should make ql.debugger = True (this will be qdb by default) ,

if you set qdb to be the default debugger, this would break compatibility (since gdb used to be the default debugger)

xwings:
i move qiling/extentions/debugger to qling/debugger, @ucgJhe u can move qdb there.

xwings:
We should mak eql.debugger = "qdb" (this will be qdb, as suggested by quynh) ,

aquynh:
you may change the filename examples/gdb.py to  examples/debugger.py

xwings:
Looks Good and thanks!

PR #438, opened 2020-08-12
Fix path transformation

wtdcode:
1. Refactor the transformation for path (`real_path`, `link_path` and `relative_path`) and it should be more robust now.
2. FS mapper can work with Windows now, see tests for details.


xwings:
You just fixed a historical bug.

PR #439, opened 2020-08-13
Add DOS support

wtdcode:
as title.


wtdcode:
Waiting for @chfl4gs to update Dockerfile and see why TravisCI gets segment fault on Windows.

.travis.yml
```diff
@@ -115,7 +115,7 @@ matrix:
         - unzip -Pinfected al-khaser.bin.zip
         - cd $TRAVIS_BUILD_DIR/tests
       script:
-        - cmd.exe //C '.\test_pe.bat'
```
xwings, line 118: We need to change this  back ?

qiling/os/dos/dos.py
```diff
@@ -0,0 +1,152 @@
...
+            self.exit_point = self.ql.exit_point
+
+        if  self.ql.entry_point is not None:
+            self.ql.loader.elf_entry = self.ql.entry_point
+        else:
+            self.ql.entry_point = self.ql.loader.start_address
+        if not self.ql.shellcoder:
+            try:
+                self.ql.reg.write("ds", self.ql.loader.cs)
+                self.ql.reg.write("es", self.ql.loader.cs)
```
xwings, line 143: all the setup code should not be in run.
most of our setup code in loader

tests/test_pe.py
```diff
@@ -17,6 +17,12 @@
 from unicorn.x86_const import *
 
 class PETest(unittest.TestCase):
+
+    def test_pe_dos_8086_hello(self):
+        ql = Qiling(["../examples/rootfs/8086_dos/HI.COM"], "../examples/rootfs/8086_dos")
```
xwings, line 22: I suggest we  can make it test_dos.py so we can run the same test in mac and linux

setup.py
```diff
@@ -2,19 +2,28 @@
...
-VERSION = ql_version
+VERSION = gb['__version__']
 
 with open('requirements.txt') as f:
     required = f.read().splitlines()
 
 with open("README.md", "r", encoding="utf-8") as ld:
     long_description = ld.read()
 
+if "linux"  in sys.platform:
```
xwings, line 21: can we move all the requirement here ? and not setup.py and requirement.txt

qiling/loader/utils.py
```diff
@@ -111,6 +112,14 @@ def getident():
     return arch, ostype, archendian
 
 def ql_pe_check_archtype(path):
+
+    ftype = magic.from_file(path)
```
xwings, line 116: since we are using libmagic, can we just use libmagic for all file type check ?

.travis.yml
chfl4gs: Timeout increased for crush debugging. Revert to orig.

wtdcode: See commit c5bd30a.

qiling/os/dos/dos.py
wtdcode: See commit 80f568e.

tests/test_pe.py
wtdcode: See commit 2cc0303.

setup.py
wtdcode: See commits 4818261, 3c6b61b and 81fca85.

qiling/loader/utils.py
wtdcode: See commit 4795da3.

wtdcode:
@chfl4gs I have updated `Dockerfile` and please rebuild a new image after PR being merged.

In addition, building capstone (4.0.2) from source doesn't work with MacOS. `python3 setup.py install` will always prefer building from sources (since setuptools doesn't support wheels) so I replace it with `pip3 install .` in `.travis.yml` which prefers wheels.

.travis.yml
```diff
@@ -54,7 +52,7 @@ matrix:
 #              sudo pip3 install $HOME/dist/*.whl;
 #          fi
         - cd $TRAVIS_BUILD_DIR 
-        - python3 setup.py install 
+        - pip3 install . # Workaround since building from capstone sources doesn't work on Macos.
```
xwings, line 55: capstone or keystone

wtdcode: keystone.

Fixed in dbb78be.

xwings:
One more things, please add this feature to ChangeLog

PR #445, opened 2020-08-14
 Qiling IDA Plugin Edit Register Support && Address Sync

kabeor:
Qiling IDA Plugin New features:

- Edit Register Support
- IDA instruction View && Register View && Stack View Address refresh synchronously
- Register View support both IDA Address(Rebase to 0x0) and Qiling Address(Real Address)


xwings:
NIce !

PR #448, opened 2020-08-16
fix missing rootfs when handling link path

ucgJhe:
before this commit, link path can be mistreated from `squahsf-root/lib/libc.so.0`  to `/lib/libc.so.0`


xwings:
symlink fixed.  now we need to figure out the fs_mapper.

Issue #446, opened 2020-08-14
transform_to_real_path broken


alfink:
There are a bunch of bugs in transform_to_real_path on the dev branch that break all my scripts

# symlinks
This is the code on the dev branch:
```python
        if isinstance(real_path, os.PathLike) and os.path.islink(real_path):
            link_path = os.readlink(real_path)
            if link_path.is_absolute():
                real_path = Path(link_path)
            else:
                real_path = Path(os.path.join(os.path.dirname(path), link_path))
```
- os.readlink returns a string, so link_path.is_absolute() will throw an error. Replacing `link_path.is_absolute()` with `Path(link_path).is_absolute()` works for me
- `real_path = Path(os.path.join(os.path.dirname(path), link_path))` returns the wrong path, as `path` is the user-provided relative path. you want to use `real_path` instead. So the line should be `real_path = Path(os.path.join(os.path.dirname(real_path), link_path))`

# fs mapping

I wanted to do a pull request for the fixes above, but the fs mapping is still broken and I don't want to do the PR before testing if my fixes work. None of my fs_mappings work anymore with the current dev branch. I think this piece of code is not really working
```python
        # TODO: A better design for fs mapping.
        for fm, to in self.ql.fs_mapper:
            if to == path and not isinstance(fm, str):
                real_path = fm
                return real_path
            elif isinstance(fm, str):
                try:
                    remains = relative_path.relative_to(Path(fm))
                    real_path = Path(to) / remains
                except ValueError:
                    continue
```
realtive_path.relative_to fails whenever relative_path is not a subpath of fm, which makes it always fail for me. Also, I'm not sure if the if-branch is correct, because the function wouldn't return a string then.

xwings:
I guess i need to ping @wtdcode , the reason why we update this code is to fix the common path conversion between linux and windows.

xwings:
Ok, you are right. It breaks all the fs_mapper. 

Iwill  need to add into CI test and make sure we dont break that again.

wtdcode:
Thanks for pointing out the two bugs in symlinks. For fs mapping, as comments suggest, I think we should have a better implementation here. I will draft a PR these days.

xwings:
@alfink either way the fs_mapper is broken for now.

Will you be able to submit a PR the two fix u mention and we will work on fixing the fs_map ?

xwings:
i think the fs_mapper works now.

@alfink will you be able to try ?

xwings:
close for  now

[...]

PR #450, opened 2020-08-17
Refactor fs mapping

wtdcode:
## Motivation

In the past, fs mapping is implemented as:

1. When path conversion between emulated systems and hosts happens, we check whether the destination is in our mapper.
2. If the mapped destination is a string, we re-construct the destination path. 
3. If the mapped destination is a user-implemented object, we return the object directly.
4. `ql_file.open` will handle the new destination according to its type.

Here we have several problems:

1. As #446 points out, we return either a string or a user-defined object before, which is rather hacky and ugly.
2. When defining a self-defined mapped object, we never know the exact methods we should implement.
3. The fs mapping doesn't support Windows since it doesn't use `ql_file` internally.

All these problems call for a new, transparent, and universal implementation for fs mapping.

## Implementation

1. Add a `FsMappedObject` so it is clear which interface is required to map a path.
2. Implement a `Fsmapper` to handle all paths we are going to open so we have a chance to check user-defined mappings.
3. Provide a similar trick for Windows `CreateFile`.
4. Remove the code handling fs mapping in `ql_file` since `ql_file` shouldn't be aware of fs mapping details.

The implementation is somewhat experimental since I'm not sure whether it will cause other side-effects.

### DIFF
--- a/qiling/os/mapper.py
+++ b/qiling/os/mapper.py
@@ -48,7 +48,7 @@ class QlFsMappedObject:
     def name(self):
         raise NotImplementedError("QlFsMappedObject property not implemented: name")
 
-class FsMapper:
+class QlFsMapper:
     
     def __init__(self, ql):
         self._mapping = {}
--- a/qiling/os/mapper.py
+++ b/qiling/os/mapper.py
@@ -58,22 +58,22 @@ class QlFsMapper:
     def ql(self):
         return self._ql
 
-    def _open_mapping_ql_file(self, fm, openflags, openmode):
-        to = self._mapping[fm]
-        if not isinstance(to, str): # We have opened this mapping or it is implemented by user.
-            return to
+    def _open_mapping_ql_file(self, ql_path, openflags, openmode):
+        real_dest = self._mapping[ql_path]
+        if not isinstance(real_dest, str): # We have opened this mapping or it is implemented by user.
+            return real_dest
         else:
-            qlfile = ql_file.open(to, openflags, openmode) # open the path and replace the destination now.
-            self._mapping[fm] = qlfile
+            qlfile = ql_file.open(real_dest, openflags, openmode) # open the path and replace the destination now.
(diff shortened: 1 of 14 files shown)


qiling/os/mapper.py
```diff
@@ -0,0 +1,107 @@
...
+# All mapped object should inherit this class.
+# Note this object is compatible with ql_file
+# Q: Why not derive from ql_file directly?
+# A: ql_file assumes that it holds a path with corresponding fd, but
+#    a FsMappedObejct doesn't have to be associated with a path or fd
+#    and thus the default implementation may cause unexpected behavior.
+#    Simply let it crash if the method is not implemented.
+#
+#    A quick way to create a FsMappedObject is `ql_file.open` or `open`.
+class FsMappedObject:
```
xwings, line 13: follow our naming ? QlFsMappedObject ?

qiling/os/mapper.py
```diff
@@ -0,0 +1,107 @@
...
+        raise NotImplementedError("FsMappedObject method not implemented: dup")
+    
+    def readline(self, end = b'\n'):
+        raise NotImplementedError("FsMappedObject method not implemented: readline")
+
+    @property
+    def name(self):
+        raise NotImplementedError("FsMappedObject property not implemented: name")
+
+class FsMapper:
```
xwings, line 51: QlFsMapper ?

qiling/os/mapper.py
```diff
@@ -0,0 +1,107 @@
...
+    
+    def __init__(self, ql):
+        self._mapping = {}
+        self._ql = ql
+    
+    @property
+    def ql(self):
+        return self._ql
+
+    def _open_mapping_ql_file(self, fm, openflags, openmode):
```
xwings, line 61: we need remove all the "fm" naming. we renamed to ql_path and real_dest？

xwings:
wow,

Travis loves you this time.

PR #455, opened 2020-08-18
Fix mapper

wtdcode:
This PR follows #450 and #446.

1. Support passing a class for mapping.
2. The open result will be cached only if user passes an object, which is compatible with previous versions.
3. Add a new test for opening a mapping multiple times.


xwings:
thanks !

PR #458, opened 2020-08-19
porting new Qdb for ARM and THUMB

ucgJhe:
as title.

### DIFF
--- a/qiling/debugger/qdb/core.py
+++ b/qiling/debugger/qdb/core.py
@@ -1,10 +1,13 @@
+#!/usr/bin/env python3
+
 import cmd
+from functools import partial
 
 from qiling import *
 from qiling.const import *
-from functools import partial
+
 from .frontend import context_printer, context_reg, context_asm, examine_mem, color
-from .utils import parse_int, handle_bnj
+from .utils import parse_int, handle_bnj, is_thumb, PROGRAM_EXITED
 
 
 
@@ -98,15 +101,6 @@ class Qldbg(cmd.Cmd):
         self.do_context()
         self._ql.emu_stop()
 
-    def run(self, address=None):
-        """
-        handle qiling instance launching
-        """
-        if not address:
-            address = self._ql.loader.entry_point
-
-        self._ql.emu_start(address, 0)
-
     def do_context(self, *args):
         """
         show context information for current location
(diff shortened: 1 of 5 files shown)


qiling/debugger/qdb/utils.py
```diff
@@ -97,6 +347,9 @@ def _read_reg(regs, _reg):
     ret_addr = cur_addr + MIPS_INST_SIZE
 
     if line.mnemonic.startswith('j') or line.mnemonic.startswith('b'):
+        if line.mnemonic == "break": # indicates program extied
+            return PROGRAM_EXITED
```
xwings, line 351: PROGRAM_EXITED maybe IS_EXIT ?

qiling/debugger/qdb/utils.py
```diff
@@ -1,27 +1,15 @@
 #!/usr/bin/env python3
-from qiling.const import *
+
 from functools import partial
 
+from qiling.const import *
 
+PROGRAM_EXITED = "DONE"
```
xwings, line 7: DONE ? why not true or false ?

qiling/debugger/qdb/core.py
```diff
@@ -203,6 +211,21 @@ def do_examine(self, args):
...
+        context_reg(self._ql, self._saved_states)
+        context_asm(self._ql, self._ql.reg.arch_pc, 4)
+
+    def do_show(self, *args):
+        """
+        show some runtime informations
+        """
+        self._ql.mem.show_mapinfo()
+        print("Qdb:", [(hex(idx), val) for idx, val in self.breakpoints.items()])
+        print("internal:", [(hex(idx), val) for idx, val in self._ql._addr_hook.items()])
```
xwings, line 227: i am not too sure. do we need print or we can use nprint. just in case

ucgJhe: this information is for debugging qdb and qiling itself mostly , and use native print for not messing up log file

qiling/debugger/qdb/utils.py
ucgJhe: this value will be returned as a address, so true or false will be converted to 0x1 or 0x0 by python automatically, will mess up the breakpoint thing , but sure i can change it

xwings:
sweet! Thanks

PR #463, opened 2020-08-20
takeaway LSB for running arm thumb mode binary with qdb in qltool

ucgJhe:
as title.

### DIFF
--- a/qltool
+++ b/qltool
@@ -7,6 +7,7 @@
 import argparse, os, string, sys, ast
 from binascii import unhexlify
 from qiling import *
+from qiling.const import QL_ARCH
 from qiling.__version__ import __version__ as ql_version
 from qiling.extensions.coverage import utils as cov_utils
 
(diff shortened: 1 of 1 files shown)


qltool
```diff
@@ -234,7 +234,7 @@ if __name__ == '__main__':
         # attach Qdb at entry point
         if options.qdb == True:
             from qiling.debugger.qdb import Qdb
-            ql.hook_address(Qdb.attach, ql.loader.entry_point)
+            ql.hook_address(Qdb.attach, ql.loader.entry_point & 0xfffffffe)
```
xwings, line 237: for arm / thumb only or all arch ?

ucgJhe: qdb is now working with MIPS and ARM, since MIPS is aligned by 4 byte its okay to both arch.

xwings: make it properly and be ready to x86 too :)

xwings:
thanks !

PR #480, opened 2020-08-27
add new feature record and replay in qdb

ucgJhe:
as title.


aquynh:
Please also update docs + example with this new feature

ucgJhe:
sure, as always

xwings:
How to enable rr in Qiling? not qltool

xwings:
Thanks!

PR #482, opened 2020-08-28
Add implementation for Nt/ZwSetInformationProcess 

lucroida:
See title.


xwings:
Cool, indeed very cool.

Nice one, Thanks !

PR #483, opened 2020-08-29
Change 'file_des' to 'fd' in the example code

hooki:
The 'file_des' member does not exist in the QlOsPosix class. The 'fd' member is incorrectly typed as 'file_des' in example codes.


xwings:
Hi,

Thanks alot !

PR #492, opened 2020-09-02
Bugfix/newselect: the timeout parameter is not properly dereferenced

yjGT:
Hello,

I found the newselect syscall emulation misses dereferencing the last parameter. The address of the stack is passed to ql_syscall_newselect, while the timeout should be dereferenced from the pointer stored on the stack rather than from the stack directly. The PR should fix it. Please review. Thanks.

### DIFF
--- a/qiling/os/posix/syscall/select.py
+++ b/qiling/os/posix/syscall/select.py
@@ -42,12 +42,15 @@ def ql_syscall__newselect(ql, _newselect_nfds, _newselect_readfds, _newselect_wr
     tmp_e_fd, tmp_e_map = parse_fd_set(ql, _newselect_nfds, _newselect_exceptfds)
 
     if _newselect_timeout != 0:
-        timeout = ql.unpack32(ql.mem.read(_newselect_timeout, 4))
+        timeout_p = ql.unpack(ql.mem.read(_newselect_timeout, 4))
+        sec = ql.unpack(ql.mem.read(timeout_p, 4))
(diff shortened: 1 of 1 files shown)


qiling/os/posix/syscall/select.py
```diff
@@ -41,13 +41,18 @@ def set_fd_set(buf, idx):
     tmp_w_fd, tmp_w_map = parse_fd_set(ql, _newselect_nfds, _newselect_writefds)
     tmp_e_fd, tmp_e_map = parse_fd_set(ql, _newselect_nfds, _newselect_exceptfds)
 
+    n = ql.archbit // 8 # 4 for 32-bit , 8 for 64-bit
+
     if _newselect_timeout != 0:
-        timeout = ql.unpack32(ql.mem.read(_newselect_timeout, 4))
+        timeout_p = ql.unpack(ql.mem.read(_newselect_timeout, n))
```
xwings, line 47: maybe you should rename timeout_p  to a more proper name too ?

yjGT: timeout_ptr? What do you suggest?I am quite open to names :)

xwings: i did not check yet. is that a pointer ? if that is a pointer then ptr is ok.

yjGT: updated. yes, it is the pointer of struct timeval *.

xwings:
Cool, thanks!

PR #493, opened 2020-09-03
complete 16 bits related packing and unpacking function also fix example hello_8086_dos.py

ucgJhe:
as title.


xwings:
the latest version i should fix the dos path.

please update docs too. the struct part

ucgJhe:
sure

xwings:
i am not sure why there is no conflict. i will merge now and check later.

thanks

PR #496, opened 2020-09-05
IDA plugin fixed some issues when restart

kabeor:
- 1. Memory View fix address
- 2. save snapshot
- 3. fixed breakpoints not work after "Restart"


xwings:
bravo

PR #499, opened 2020-09-06
fix typo

hooki:


xwings:
this is valid and how can we make such a silly mistake.

thanks !

PR #501, opened 2020-09-08
Doogie example

wtdcode:
Doogie example with some fix for real mode emulation.

Run doogie_8086_crack.py for details.


xwings:
cool, thanks !

PR #504, opened 2020-09-09
Doogie fix crlf

wtdcode:
1. Fix CRLF problem when curses get input.
2. Map the MBR file by default so we can run it with qltool directly.
3. Slight modify doogie showcase for clarity.


xwings:
ok sir

PR #509, opened 2020-09-10
ollvm de-flatten(deflat) implementation with IDA plugin && some minor fix

wtdcode:
1. The idaplugin doesn't work if it is put into `plugins/` directory directly.  Calling `run` in `__init__` could fix it. However, the tooltip still doesn't show. Note that using the plugin as a standalone script always works so I guess it's highly likely that it is an IDA bug (or may be related to Python3.8, see [this](https://www.hex-rays.com/blog/ida-7-4-and-python-3-8/)).
2. ollvm de-flatten(deflat) implementation. Some screenshots below and I will write docs later.

Before deflat:

Green: Real blocks.
Blue: Dispatcher.
Gray: Fake blocks.
Pink: Return blocks.
Yellow: The first block.

Snipaste_2020-09-11_02-03-54

Snipaste_2020-09-11_02-04-29

After deflat:

Only real blocks are reserved and their relations are fixed by symbolic execution.

Snipaste_2020-09-11_02-05-05

Snipaste_2020-09-11_02-05-11

New actions:

Snipaste_2020-09-11_02-04-16

Note: We allow users to mark real, fake and return blocks since auto analysis might be incorrect sometimes.

Some reference:
https://github.com/obfuscator-llvm/obfuscator/wiki/Control-Flow-Flattening
https://blog.quarkslab.com/deobfuscation-recovering-an-ollvm-protected-program.html
http://ac.inf.elte.hu/Vol_030_2009/003.pdf

### DIFF
--- a/qiling/extensions/idaplugin/Qiling.py
+++ b/qiling/extensions/idaplugin/Qiling.py
@@ -602,6 +602,8 @@ class QlEmuPlugin(plugin_t, UI_Hooks):
         self.userobj = None
         self.customscriptpath = None
         self.bb_mapping = {}
+        self.registered = False
+        self.run()
 
     ### Main Framework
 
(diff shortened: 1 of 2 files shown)


xwings:
ok, this is awesome!

PR #512, opened 2020-09-15
Use logging instead of print and rewrite examples.

wtdcode:
As title.


xwings:
merge for now. thanks.

PR #517, opened 2020-09-17
add udp socket support

cq674350529:
add support for udp sockets.

ps: I'm not sure why `ql_syscall_send` is missing in `qiling/os/posix/syscall/net.py`, so just add it.


xwings:
Hi,

we do have a binary to test TCP and included in test/test_elf.py can we have a similar one for UDP ?

cq674350529:
Ok, I'll have a look and try to add a similar test for UDP.

cq674350529:
I added some tests for UDP sockets based on the TCP tests.

By the way, it seems that the current TCP tests have some drawbacks. 

First, the communications between `tcp_client` and `tcp_server` in `qiling/examples/src/linux/tcp_test.c` use `read()` and `write()` directly, instead of `recv()` and `send()`. So it didn't detect the missing `ql_syscall_send` in `qiling/os/posix/syscall/net.py`.

In addition, there are some unreachable codes in the `tcp_client`.
https://github.com/qilingframework/qiling/blob/03c8707727d05915555655c907e4f36612f2ca4e/examples/src/linux/tcp_test.c#L163-L179

xwings:
Feel free to update it. 👍 

I was thinking we should tear down those API and put it in a class.

So, we can do ql.network.tcp.send() something like this.

so all the os can use and share the same network implementation.

cq674350529:
For the TCP socket tests, I'll do it in another pull request.

As for the re-design, keep the same api for different os is appealing. But it may need more extra works.

xwings:
Of coz. let me think and redraw abit.

xwings:
anyway, thank for the PR !

PR #518, opened 2020-09-18
Add debugger level APIs

kabeor:


qiling/debugger/debugger.py
```diff
@@ -2,11 +2,27 @@
...
 
 class QlDebugger():
-    def __init__(self, ql):
+    def __init__(self, ql:Qiling):
         self.ql = ql
 
+    def dbg_start(self):
+        pass
+
+    def dbg_run(self, begin_addr=None, end_addr=None):
```
xwings, line 15: We dont need a api to call another API

qiling/debugger/debugger.py
```diff
@@ -2,11 +2,27 @@
...
+    def dbg_step(self):
+        pass
+
+    def dbg_continue(self):
+        pass
+
+    def dbg_set_breakpoint(self):
+        pass
 
     def addr_to_str(self, addr, short=False, endian="big"):
```
xwings, line 27: move this to gdb. i think this is very RSP (gdb) specific.

qiling/debugger/disassember.py
```diff
@@ -0,0 +1,44 @@
...
+from qiling import *
+from qiling.const import *
+from capstone import *
+
+
+class QlDisassember():
+    def __init__(self, ql:Qiling):
+        pass
+
+    def disasm_all_lines(self, ql:Qiling, seg_name='.text'):
```
xwings, line 15: disam_elf , disasm_pe

this part is rather file format specific and not OS specific ?

xwings:
Cool, thanks!

PR #527, opened 2020-09-23
Fix ql_syscall_writev only having the correct return value in "debug" and "dump" output modes.

nezza:
I noticed my binary was only working when running in "debug" or "dump" mode - this patch makes sure that ql_syscall_writev always returns the correct return value.


xwings:
Hi and welcome to your first PR!

I suggest the content of writeev sould only print with ql.dprnt(D_CTNT)
only higher verbose will print out.

the if statement should only once,  not twice

nezza:
Hi, thanks (though not my first PR :) )
I've attached a commit that uses ql.dprint instead of the `ifs`.

xwings:
Cool. Thanks !

PR #529, opened 2020-09-23
Fix exec_arbitrary

assafcarlsbad:
`exec_arbitrary` used to leave the stack in an inconsistent state where the return address has been placed into `arch_pc`, but the stack pointer was not modified to simulate the effects of a function return.


xwings:
Make sense. thanks !

PR #530, opened 2020-09-24
drcov_exact: coverage collection at instruction granularity

assafcarlsbad:
Sometimes, when an exception is generated in the middle of a basic block, block-level coverage might "paint" instructions which weren't really executed.

image

This PR introduces a coverage collector which works at instruction granularity, hence it is more accurate:

image


xwings:
Welcome back!

Nice !

PR #538, opened 2020-10-02
Fixed memory writes from GDB

JeffmeisterJ:
Small bugfix. Memory writes from gdb don't currently work, because the server tries to interpret the address AND the data as an integer. This is fixed by converting the data (a hexadecimal string) into bytes.


xwings:
Coool. Thanks and welcome to the community !

JeffmeisterJ:
Happy to contribute, love the project!

PR #539, opened 2020-10-04
netgear_6220_mips32el_linux plus

zi0Black:
This file's intension is to provide some examples of more advanced operation done with qiling on this binary.
Added:
File -> netgear_6220_mips32el_linux_plus.py
Custom api -> bind -> read & decode C struct from memory

Improved:
N/A

Fixed:
N/A

### DIFF
--- a/examples/netgear_6220_mips32el_linux_plus.py
+++ b/examples/netgear_6220_mips32el_linux_plus.py
@@ -1,0 +1,82 @@
+#!/usr/bin/env python3
+#
+# Cross Platform and Multi Architecture Advanced Binary Emulation Framework
+# Built on top of Unicorn emulator (www.unicorn-engine.org)
+
+# After mapping /proc there will be a /dev/mtdblock11 missing and crash
+# To fix this,
+#   - cd $yourfirmware_rootfs/dev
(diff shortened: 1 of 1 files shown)


xwings:
We just need a copy of R6220. My suggestion is either:

1. Use another firmware for example, i guess all the netgear are pretty much the same
2. Merge this into the current R6220, we dont need two 6220

Also, if you can blog this example it will be really awesome!

xwings:
Also, Welcome to the community!

zi0Black:
> We just need a copy of R6220. My suggestion is either:
> 
> 1. Use another firmware for example, i guess all the netgear are pretty much the same
> 2. Merge this into the current R6220, we dont need two 6220
> 
> Also, if you can blog this example it will be really awesome!

I think I will merge with the existing one! (I thought about creating a new file to keep the first one as simple as possible and have another one more complex e detailed were to keep posting more complex tricks). 

I don't have a blog :/

xwings:
Ok, merge the current one will be perfect.

You can always write in https://docs.qiling.io  , the github repo in https://github.com/qilingframework/rtfd.io

zi0Black:
> Ok, merge the current one will be perfect.
> 
> You can always make your own blog with github pages. https://pages.github.com/

ACK.

PR #540, opened 2020-10-04
Added example on netgear_6220

zi0Black:
Added:
Custom api -> bind -> read & decode C struct from memory

Improved:
N/A

Fixed:
N/A


xwings:
nice, thanx !

PR #541, opened 2020-10-05
add Demigod core

aquynh:
add Demigod core to Qiling, to enable kernel module emulation.

see https://groundx.io/demigod for more information.


xwings:
Fianlly! Welcome Home!

PR #544, opened 2020-10-06
Add mappings to some missing x64 partial registers.

assafcarlsbad:
This allows users of the framework to access these registers as simply as `ql.reg.r8b`, which wasn't possible before.


xwings:
Thanks !

xwings:
Hi,

Sorry my bad.

These register are belongs to x86 64 bits only.  can you please move them all to reg_map_64

PR #547, opened 2020-10-10
IDA plugin various features

wtdcode:
1. Execute Selection. Partial execution.
2. Set PC. With "Continue", it's easy to achieve "Execute from current address" and provides some extra flexibility.
3. Junk code removal by static pattern matching. Due to the limit of available samples, I only implement one pattern:

```
jb addr
jnb addr
; some instructions or junk here
addr:
; some other instructions
```

4. Junk code removal by executing and nop unexecuted instructions. This is helpful for some programs like this:

```
mov eax, 1
cmp eax, 1
jmp lb
; some junk here.
lb:
; some other instructions
```

Docs is on the way.

### DIFF
--- a/examples/extensions/idaplugin/custom_script.py
+++ b/examples/extensions/idaplugin/custom_script.py
@@ -31,3 +31,8 @@ class QILING_IDA():
         hook = []
         hook.append(ql.hook_code(step_hook))
         return hook
+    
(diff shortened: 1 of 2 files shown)


wtdcode:
Snipaste_2020-10-10_17-35-34

The screenshot for the new popup menu.

xwings:
cool, thanks

PR #553, opened 2020-10-18
IDA Deflat renew && Various fix for Qiling

wtdcode:
This PR is somewhat big and includes various features.

For Qiling:

- Now the `archtype`, `ostype` and `archendian` can be overwritten in the `__init__` of Qiling.
- Fix a bug in `function_hook`
- Add w0-w30 regs to arm64 register mapping.
- Add `create_assembler` API in utils.

For IDA plugins:

- Support deflating for arm and arm64.
- Introduce address conversion for ASLR binaries and support deflating for lib.
- Introduce IDA microcode API for experimental pattern matching and instruction identification. If that fails, it will fallback to legacy implementation.
- Optimize code patching by some IDA hacks.
- Switch to keystone instead of IDA assembler.
- Save/Restore context to make the path search much more robust.
- Add some debug logs.

Still, some work is left to be done (See TODO and NOTE in source code) and I will draft another PR when I get that done.


wtdcode:
Aslo:

- Move `add_assembler/disassembler` to core_utils.

PR #554, opened 2020-10-19
add support for x86 lkm

cq674350529:
add support for loading x86 lkm.

I compile the lkm-rootkit code (https://github.com/m0hamed/lkm-rootkit) on x86 machine to reuse the exist kernel apis. However, there are still some differences between x86 and x8664. So I only run the function `procfile_read()`.
image

```shell
$ python3 qiling_demo.py 
[+] loadbase: 0, mem_start: 1000, mem_end: 5000
init_module = 0xdc0
Writing syscall sys_close to [0x100100c]
[+] 0x11e0         55                            push ebp
[+] 0x11e1         89 e5                         mov ebp, esp
[+] 0x11e3         83 ec 04                      sub esp, 4
[+] 0x11e6         e8 15 ee ff 00                call 0x1000000
[+] 0x1000000      00 00                         add byte ptr [eax], al
mcount(Ptr = 0x0) = 0x0
[+] 0x11eb         c7 04 24 50 21 00 00          mov dword ptr [esp], 0x2150
[+] 0x11f2         e8 21 ee ff 00                call 0x1000018
[+] 0x1000018      00 00                         add byte ptr [eax], al
printk(format = "DONT YOU EVER TRY TO READ THIS FILE OR I AM GOING TO DESTROY YOUR MOST SECRET DREAMS")
[+] 0x11f7         31 c0                         xor eax, eax
[+] 0x11f9         c9                            leave
```

By the way, in function `lkm_dynlinker()`, `0x2000000` is added  when resolving `R_X86_64_64` relocation item.

https://github.com/qilingframework/qiling/blob/8bf5ac0831e753c2aecfa129dca2626c78135cb5/qiling/loader/elf.py#L520-L526

I refer to some manuals (https://www.intezer.com/blog/elf/executable-and-linkable-format-101-part-3-relocations/), and the `R_X86_64_64` item is calced as `S + A`. So why `0x2000000` is added?  Does the relocation for lkm differ from normal elf? Thanks!


qiling/os/linux/fncc.py
```diff
@@ -205,7 +205,7 @@ def x86_stdcall(ql, param_num, params, func, args, kwargs):
     result, param_num = __x86_cc(ql, param_num, params, func, args, kwargs)
 
     # update stack pointer
-    ql.sp = ql.sp + ((param_num + 1) * 4)
+    ql.reg.sp = ql.reg.sp + ((param_num + 1) * 4)
```
xwings, line 208: This should be ql.reg.arch_sp.

x86 might have issue with ESP(32bit), RSP(64bit) and SP(16bit). That is the reason we name it to ql.reg.arch_sp

xwings:
@aquynh you might be able to answer this.

in loader/elf.py line 523. why 0x200000. details in PR description.

qiling/os/linux/fncc.py
cq674350529: Thx, update it.

xwings:
One more thing is, can you make a tests cast for this?

cq674350529:
Add a simple test.

By the way, there is an undefined var `cc` in line 241. Just comment it for it's not used currently.
https://github.com/qilingframework/qiling/blob/a4e42ac28089c64c78a705bf402191b42e24bfc9/qiling/os/linux/fncc.py#L231-L250

xwings:
Hi,

I will merge this. 

> @aquynh you might be able to answer this.
> 
> in loader/elf.py line 523. why 0x200000. details in PR description.

Regarding this, can you issue a fix for this ?

cq674350529:
The existing test for x8664_m0hamed_rootkit.ko works well, so maybe fix it when come across this issue later ?

xwings:
Agree,

I will merge now.

PR #563, opened 2020-10-27
Extending snapshot mechanism to allow saving and restoring of OS and Loader information

liba2k:
Handling mainly UEFI Loader and OS, For for other OSs will have to be added as needed.

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -238,7 +238,7 @@ class Qiling(QlCoreStructs, QlCoreHooks, QlCoreUtils):
 
 
     # save all qiling instance states
-    def save(self, reg=True, mem=True, fd=False, cpu_context=False, snapshot=None):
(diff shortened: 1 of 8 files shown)


qiling/core.py
```diff
@@ -238,7 +238,7 @@ def enable_lib_patch(self):
 
 
     # save all qiling instance states
-    def save(self, reg=True, mem=True, fd=False, cpu_context=False, snapshot=None):
+    def save(self, reg=True, mem=True, fd=False, cpu_context=False, os_context=True, loader=True, snapshot=None):
```
xwings, line 241: should os_context and loader = False ?

xwings:
btw, update changelog too :)

qiling/core.py
liba2k: Whatever you think will serve the project better.

xwings:
Cool, thanks. Merged.

PR #576, opened 2020-11-12
Add LR to arm64 and improve handling of register (uppercase)

bkerler:
Improve handling of registers (now not only lowercase but also uppercase work now, like ql.reg.x0 and ql.reg.X0) and added LR to aarch64/arm64 as register.


xwings:
Thanks for the PR.

Also, welcome to the community !

xwings:
I rebased it to dev branch.

PR #578, opened 2020-11-12
Separate SMM function pointers from BootServices function pointers to…

liba2k:
… allow sanitation and access enforcement.


xwings:
awesome! thanks!

PR #581, opened 2020-11-12
Gimme more registers

bkerler:
This will add dynamic loading of registers in unicorn.

Why ?
- 1. Makes unicorn updates easier (if const are renamed)
- 2. Gives access to all registers
- 3. Less code to maintain


xwings:
Two things

1. If we need to change ARM, ARM64 and MIPS. We need to change it all
2. there is some conflict, you might wanna take a look

Last things,

Referring to "Makes unicorn updates easier (if const are renamed)", very likely this will help since unicorn will not rename the register. But, yes but. We need a way to make our engine switch easier.

bkerler:
I will have a look how we can restructure the code to implement a clean x86 solution

Issue #570, opened 2020-11-06
UnboundLocalError: local variable 'reg_list' referenced before assignment


baptistapedro:
Any thoughts on this please?

```python
>>> from qiling import *
>>> ql = Qiling(["file.sys"], 'x8664_windows')
[+] Initiate stack address at 0x7ffffffde000
[+] Loading file.sys to 0x1c0000000
[+] PE entry point at 0x1c01085a0
[+] Driver object addr is 0x6000000
[+] Registry path addr is 0x6000150
[+] EPROCESS is is 0x6000160
[+] KI_USER_SHARED_DATA is 0xfffff78000000000
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/root/.local/lib/python3.8/site-packages/qiling/core.py", line 164, in __init__
    self.loader.run()
  File "/root/.local/lib/python3.8/site-packages/qiling/loader/pe.py", line 398, in run
    self.load()
  File "/root/.local/lib/python3.8/site-packages/qiling/loader/pe.py", line 489, in load
    self.ql.os.set_function_args((self.ql.driver_object_address, self.ql.regitry_path_address))
  File "/root/.local/lib/python3.8/site-packages/qiling/os/os.py", line 143, in set_function_args
    self._x8664_set_args(args)
  File "/root/.local/lib/python3.8/site-packages/qiling/os/os.py", line 136, in _x8664_set_args
    self.ql.uc.reg_write(reg_list[i], args[i])
UnboundLocalError: local variable 'reg_list' referenced before assignment
```

xwings:
@quangnh89 any idea ?

bkerler:
should be fixed with pr #582

xwings:
fixed and close

PR #582, opened 2020-11-12
Fix ostype bug

bkerler:
This fixes issue #570


xwings:
cool. nice ! 
fix issue, https://github.com/qilingframework/qiling/issues/570

Issue #535, opened 2020-10-02
TypeError: string argument expected, got 'bytearray'


zbx911:
Testing started at 8:27 ...
D:\Anaconda3_x64_2020_7\python.exe "C:\PyCharm 2020.2.1\plugins\python\helpers\pydev\pydevd.py" --multiproc --save-signatures --qt-support=auto --client 127.0.0.1 --port 52849 --file "C:\PyCharm 2020.2.1\plugins\python\helpers\pycharm\_jb_unittest_runner.py" --path M:/qiling/qiling-learn/tests/test_android.py
pydev debugger: process 11500 is connecting

Connected to pydev debugger (build 202.6948.78)
Launching unittests with arguments python -m unittest M:/qiling/qiling-learn/tests/test_android.py in M:\qiling\qiling-learn\tests

[+] load 0x555555554000 - 0x555555555000
[+] load 0x555555555000 - 0x555555557000
[+] mem_start: 0x0 mem_end: 0x3000
[+] interp is : ../examples/rootfs/arm64_android/system/bin/linker64
[+] interp_mem_size is : 0x15d000
[+] interp_address is : 0x7ffff7dd5000
[+] mmap_address is : 0x7fffb7dd6000
[+] rel name b'__register_atfork'
[+] rel name b'puts'
[+] rel name b'__libc_init'
[+] rel name b'__cxa_atexit'
[+] log mmap - mmap(0x0, 0x5000, 0x0, 0x22, 4294967295, 0)
[+] log mmap - mmap(0x0, 0x5000, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, 4294967295, 0)
[+] log mmap - return addr : 0x7fffb7dd6000
[+] log mmap - addr range  : 0x7fffb7dd6000 - 0x7fffb7ddb000
[+] log mmap - mapping needed
[+] mmap_base 0x7fffb7dd6000  length 0x5000
mmap(0x0, 0x5000, 0x0, 0x22, 4294967295, 0) = 0x7fffb7dd6000
[+] mmap_base is 0x7fffb7dd6000
prctl() = 0
mprotect(0x7fffb7dd7000, 0x3000, 0x3) = 0
[+] mprotect(0x7fffb7dd7000, 0x3000, PROT_READ | PROT_WRITE) = 0
prctl() = 0
set_tid_address(7ffff7f2c9c0) = 11500
facccessat (4294967196, 0x7ffff7e93972, 0x4) = -1
[...]

PR #583, opened 2020-11-12
Fix write_fd issue

bkerler:
This should fix issue #535


xwings:
Maybe add a else with ql.dprint and says content is nothing ?

bkerler:
Yes, good idea. Pushed the changes and used hasattr instead of dir :)

xwings:
nice, thanks !

PR #585, opened 2020-11-13
ELF memory mapping fixes and add EDL tests

bkerler:
This addresses some issues left with memory mapping (such as double mapping) and also adds a test case for EDL files.
Enjoy :)


xwings:
Perfecto!

PR #586, opened 2020-11-15
Reimplement some System Information APIs

nullableVoidPtr:


xwings:
There is some error.

You can trace from here, https://api.travis-ci.com/v3/job/440138803/log.txt

nullableVoidPtr:
Ah, it uses [`time.time_ns`](https://docs.python.org/3/library/time.html#time.time_ns), which was introduced in 3.7.
The other option would just be to take the float and multiply and truncate.

nullableVoidPtr:
Something is wrong with the Khazar binary causing an invalid memory reference but I'm not sure what, which is strange seeing as this PR does not touch memory at all. I'll have to look at it some time.

PR #591, opened 2020-11-22
Implement mprotect syscall on Linux.

nullableVoidPtr:
`mprotect` with protection keys are too complex to be implemented for now, this PR only handles rudimentary RWX protections.
Interestingly, there's was not a single use of `ql.mem.protect` in the codebase, but I think this can be extended to any POSIX platform, and even Windows and `VirtualProtect`.
I have added one x64 test binary, so I hope that it would be all green on Travis.


xwings:
We dont need a actual mprotect.

this is a simulated OS mprotect can be fake.

nullableVoidPtr:
Hmm, but there can be instances where a binary `mprotect`s itself instead of an `mmap` or `malloc`'d page.
I can close this PR and opt to improve documentation on the `QlMemoryManager` instead, but this PR also fixes `protect` since before, it tries to align the address wrong.

xwings:
Agree,

but still you need to fix the travis CI check

nullableVoidPtr:
Is `mem.map_info` a copy of `uc.mem_regions`? `mem.save` iterates over `map_info` to dump, but when you restore, it checks each region with `is_mapped`, which uses `uc.mem_regions`.

xwings:
You need the check the code. I dont really remember. everything its in memory.py

nullableVoidPtr:
I did some extra debugging, and now they both look consistent doing the ELF tests, but I worry about `map_info` and `mem_regions` being subtly inconsistent due to some weird chain of `mprotect`s and `munmap`s, since [the code that splits regions](https://github.com/unicorn-engine/unicorn/blob/dbc6cc27c7c8d40e6061918c20ad513cae5da39f/uc.c#L826) might not be exactly equivalent to `add_mapinfo` or `del_mapinfo`.

In my opinion, this PR is good to be merged for now, but I think `map_info` and `mem_regions` can be coupled by either: reconstructing `mem.map_info` using `uc.mem_regions` after an update to memory regions; or having something like
```py
@property
def map_info():
    ...
```
which uses `uc.mem_regions`. Either would mean syscall implementations like [`mmap`](https://github.com/qilingframework/qiling/blob/5184fe8696c8a821f7c47d18900a852aa4678ca1/qiling/os/posix/syscall/mman.py#L195) wouldn't need to update map_info manually. This would have to be done in another PR.

xwings:
I will hold some PR for now.

We got one more round of internal refactoring going on. It will take a week or two.

nullableVoidPtr:
Just merged upstream dev (refactor) and `test_elf` is all green. Hope it looks good to you.

xwings:
Rechecking this PR. 

Will you be able to fix the conflict? 

Thanks!

nullableVoidPtr:
Sorry for not being active on this PR! I had exams.

I'll see if I can conform to the refactor, move tests to the rootfs, etc.

nullableVoidPtr:
Hmm, it seems like mprotect is already implemented by the time you reminded me of this PR. I can contribute my own tweaks, especially with my concerns with `map_info` (which seems to be amended with the refactor, but is also not considered in `mem.protect`), but that is all, and might be best as a another separate PR. Closing.

PR #595, opened 2020-12-05
fix a uaf hook bug

A1Lin:
when free a size-8 chunk(alloc(0)), the uaf hook end address is smaller than the begin address, any memory access instructions will trigger the uaf handler by mistake.

### DIFF
--- a/qiling/extensions/sanitizers/heap.py
+++ b/qiling/extensions/sanitizers/heap.py
@@ -113,8 +113,9 @@ class QlSanitizedMemoryHeap():
         # Install the UAF canary hook.
         self.ql.mem.write(addr, self.canary_byte * (chunk.size - 8))
         uaf_canary = (addr, addr + chunk.size - 8 - 1, CaneryType.uaf)
-        self.ql.hook_mem_valid(self.uaf_handler, begin=uaf_canary[0], end=uaf_canary[1])
-        self.canaries.append(uaf_canary)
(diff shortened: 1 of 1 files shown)


wtdcode:
Any script or test binary?

xwings:
maybe @liba2k or @assafcarlsbad can decide :)

A1Lin:
> Any script or test binary?


[test.zip](https://github.com/qilingframework/qiling/files/5649123/test.zip)
This is the test script

assafcarlsbad:
Great catch @A1Lin! That's definitely a bug in the UAF detection logic. I'm just wondering if the right way to fix it is like you proposed in your PR or if we need to somehow instrument differently 0-size allocations in the first place.
@xwings @wtdcode @liba2k @A1Lin WDYT?

xwings:
@assafcarlsbad i agree. since is a sanitizer we might need some sort of instrumentation

A1Lin:
> Great catch @A1Lin! That's definitely a bug in the UAF detection logic. I'm just wondering if the right way to fix it is like you proposed in your PR or if we need to somehow instrument differently 0-size allocations in the first place.
> @xwings @wtdcode @liba2k @A1Lin WDYT?

It's better if we have some instrumentation

xwings:
in this case, we might need to allow user to set ql.os.heap.uaf_canary ?

PR #601, opened 2020-12-15
fix sin_family for MSB binary

ucgJhe:
as title.


xwings:
There is some conflict. can you please fix it.

xwings:
Cool, thanks!

PR #604, opened 2020-12-19
Make gdbserver great again

bet4it:
- [x] Have you checked that there aren't other open [pull requests](https://github.com/qilingframework/qiling/pulls) for the same update/change?
- [x] Is this pull request for [dev](https://github.com/qilingframework/qiling/tree/dev) branch? (Do not pull request for **master** branch).

-----

`reg_map` was used when gdbserver send all the registers, registers like `w0` shouldn't be included in `reg_map`

### DIFF
--- a/qiling/arch/arm64_const.py
+++ b/qiling/arch/arm64_const.py
@@ -37,6 +37,14 @@ reg_map = {
             "x28": UC_ARM64_REG_X28, 
             "x29": UC_ARM64_REG_X29,
             "x30": UC_ARM64_REG_X30, 
+            "sp": UC_ARM64_REG_SP,
+            "pc": UC_ARM64_REG_PC,
+            "lr": UC_ARM64_REG_LR,
+            "cpacr_el1": UC_ARM64_REG_CPACR_EL1,
+            "tpidr_el0": UC_ARM64_REG_TPIDR_EL0,
+}
+
+reg_map_part = {
             "w0" : UC_ARM64_REG_W0,
             "w1" : UC_ARM64_REG_W1,
             "w2" : UC_ARM64_REG_W2,
@@ -68,9 +76,4 @@ reg_map = {
             "w28" : UC_ARM64_REG_W28,
             "w29" : UC_ARM64_REG_W29,
             "w30" : UC_ARM64_REG_W30,
-            "sp": UC_ARM64_REG_SP, 
-            "pc": UC_ARM64_REG_PC,
-            "lr": UC_ARM64_REG_LR,
-            "cpacr_el1": UC_ARM64_REG_CPACR_EL1,
-            "tpidr_el0": UC_ARM64_REG_TPIDR_EL0,
 }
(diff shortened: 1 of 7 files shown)


qiling/arch/arm64_const.py
```diff
@@ -37,6 +37,14 @@
...
             "x29": UC_ARM64_REG_X29,
             "x30": UC_ARM64_REG_X30, 
+            "sp": UC_ARM64_REG_SP,
+            "pc": UC_ARM64_REG_PC,
+            "lr": UC_ARM64_REG_LR,
+            "cpacr_el1": UC_ARM64_REG_CPACR_EL1,
+            "tpidr_el0": UC_ARM64_REG_TPIDR_EL0,
+}
+
+reg_map_part = {
```
xwings, line 47: this changes is gdb specific. gdb should handle the map and split.

bet4it: This regression was introduced in ef0ac2a39ac408d2b61ba235894591ad8fdd7e25

xwings: What i am saying is, the split should be done by gdb. not the main constant

bet4it: `reg_map` is used to generate `self.tables`:
https://github.com/qilingframework/qiling/blob/d59436e673a0f7feaf3865cf0fb9245d4ec0d3a5/qiling/debugger/gdb/gdb.py#L91-L98
The order of registers in `self.tables` should be same with their order in gdb, which is used in:
https://github.com/qilingframework/qiling/blob/d59436e673a0f7feaf3865cf0fb9245d4ec0d3a5/qiling/debugger/gdb/gdb.py#L234
https://github.com/qilingframework/qiling/blob/d59436e673a0f7feaf3865cf0fb9245d4ec0d3a5/qiling/debugger/gdb/gdb.py#L292
https://github.com/qilingframework/qiling/blob/d59436e673a0f7feaf3865cf0fb9245d4ec0d3a5/qiling/debugger/gdb/gdb.py#L379
https://github.com/qilingframework/qiling/blob/d59436e673a0f7feaf3865cf0fb9245d4ec0d3a5/qiling/debugger/gdb/gdb.py#L435
We could handle the split in these places, but the changes will be will ugly and unnecessary.

xwings: make sense,

and again, what does reg_map_part means ? maybe a better na,e ?

bet4it: `part` represents `partial registers` or just `part of a register`
I'm not sure how to call registers like `eax` or `w0` in a better way
I thought of name like `reg_map_lb`, where `lb` represents `lower bits`, but I don't know if it's proper to use this on registers like `ah`

xwings: extend sounds better
reg_map_extend

bet4it: Do you mean `w0` is an extend of `x0`? I don't think it's suitable.

xwings: its a extension of the current main set

bet4it: That's up to you, you can change it by yourself😄

xwings: your PR :) once you update i will merge.

xwings:
I will do it :)

PR #606, opened 2020-12-19
Some minor usability tweaks

learn-more:
* Log emu_error on the error logger
* Fix disasm logging to be on one line again
* Add windows registry folders to .gitignore
* PE: Only load ntoskrnl on drivers
* Hide sality test behind QL_FAST_TEST

If you disagree with any of those changes please let me know, I'll drop that commit then

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -349,17 +349,17 @@ class QlLoaderPE(QlLoader, Process):
         self.path       = self.ql.path
 
     def run(self):
-        self.init_dlls = [b"ntoskrnl.exe", b"ntdll.dll", b"kernel32.dll", b"user32.dll"]
-        self.sys_dlls = [b"ntoskrnl.exe", b"ntdll.dll", b"kernel32.dll"]
+        self.init_dlls = [b"ntdll.dll", b"kernel32.dll", b"user32.dll"]
+        self.sys_dlls = [b"ntdll.dll", b"kernel32.dll"]
         self.pe_entry_point = 0
         self.sizeOfStackReserve = 0        
 
-        if self.ql.shellcoder:
-            self.init_dlls.remove(b"ntoskrnl.exe")
-            self.sys_dlls.remove(b"ntoskrnl.exe")
-        else:
+        if not self.ql.shellcoder:
             self.pe = pefile.PE(self.path, fast_load=True)
             self.is_driver = (self.pe.OPTIONAL_HEADER.Subsystem == 1)
+            if self.is_driver:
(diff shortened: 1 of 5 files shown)


xwings:
We dont need a fast test.

well, you can always have it if u want. But fast test should not be in Qiling.

xwings:
But other changes should be cool :) thanks

learn-more:
> We dont need a fast test.
> 
> well, you can always have it if u want. But fast test should not be in Qiling.

The fast test is the option from #372 to skip malware.

xwings:
You are right, i am blid.

qiling/loader/pe.py
```diff
@@ -349,17 +349,17 @@ def __init__(self, ql):
...
-        if self.ql.shellcoder:
-            self.init_dlls.remove(b"ntoskrnl.exe")
-            self.sys_dlls.remove(b"ntoskrnl.exe")
-        else:
+        if not self.ql.shellcoder:
             self.pe = pefile.PE(self.path, fast_load=True)
             self.is_driver = (self.pe.OPTIONAL_HEADER.Subsystem == 1)
+            if self.is_driver:
+                self.init_dlls = [b"ntoskrnl.exe"]
+                self.sys_dlls = [b"ntoskrnl.exe"]
```
learn-more, line 362: Nothing should import from both ntoskrnl.exe and ntdll.dll
Considering the 'shellcoder' case already removed ntoskrnl, this should leave that case in the same state.
The is_driver case now completely drops those usermode dll's, and replaces them all with ntoskrnl.
Services (f.e. svchost.exe or umpnpmgr.dll) Do not have the Subsystem `1` set, and import from ntdll and more. (and not ntoskrnl)

qiling/os/os.py
```diff
@@ -92,32 +92,32 @@ def find_containing_image(self, pc):
...
     def emu_error(self):
-        logging.info("\n")
+        logging.error("\n")
 
         for reg in self.ql.reg.register_mapping:
             if isinstance(reg, str):
                 REG_NAME = reg
                 REG_VAL = self.ql.reg.read(reg)
-                logging.info("[-] %s\t:\t 0x%x" % (REG_NAME, REG_VAL))
+                logging.error("%s\t:\t 0x%x" % (REG_NAME, REG_VAL))
```
learn-more, line 101: I removed the prefix here, since the `logging` formatter already puts a prefix like this in front.
I assumed that this was a leftover from a global replace with the `ql.nprint` to `logging.info` ?

xwings:
Cool. Let me merge this.

PR #609, opened 2020-12-22
Fixes needed to run drivers

learn-more:
- Fix QlOsUtils X8664 MDL generation
- Fix passthru smashing the stack
- Fix ntoskrnl_IoCreateDevice when called from IoCreateDevice

### DIFF
--- a/qiling/os/os.py
+++ b/qiling/os/os.py
@@ -304,7 +304,7 @@ class QlOs(QlOsUtils):
 
     def x86_stdcall(self, param_num, params, func, args, kwargs, passthru=False):
         # if we check ret_addr before the call, we can't modify the ret_addr from inside the hook
-        result, param_num = self.__x86_cc(param_num, params, func, args, kwargs)
+        result, param_num = self.__x86_cc(param_num, params, func, args, kwargs, passthru)
 
         # get ret addr
         ret_addr = self.ql.stack_read(0)
@@ -312,10 +312,10 @@ class QlOs(QlOsUtils):
         # append syscall to list
         self._call_api(func.__name__, params, result, self.ql.reg.arch_pc, ret_addr)
 
-        # update stack pointer
-        self.ql.reg.arch_sp = self.ql.reg.arch_sp + ((param_num + 1) * 4)
-
         if not passthru and self.PE_RUN:
+            # update stack pointer
+            self.ql.reg.arch_sp = self.ql.reg.arch_sp + ((param_num + 1) * 4)
+
             self.ql.reg.arch_pc = ret_addr
 
         return result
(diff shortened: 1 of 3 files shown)


qiling/os/utils.py
```diff
@@ -122,10 +122,7 @@ def normalize(path):
 class QlOsUtils:
     def __init__(self, ql):
         self.ql = ql
-        self.archtype = None
-        self.ostype = None
         self.path = None
-        self.archendian = None
```
learn-more, line 128: These were never initialized, so I removed them.

qiling/os/windows/dlls/ntoskrnl.py
```diff
@@ -179,7 +179,7 @@ def ntoskrnl_IoCreateDevice(ql, address, params):
     device_object.CurrentIrp.value = 0
     device_object.Timer.value = 0
     device_object.Flags = 0x00000080  # DO_DEVICE_INITIALIZING
-    if params['Exclusive']:
+    if 'Exclusive' in params and params['Exclusive']:
```
learn-more, line 182: Exclusive was not passed in params when this was called from `IoCreateDevice`

qiling/os/os.py
```diff
@@ -304,18 +304,18 @@ def _call_api(self, name, params, result, address, return_address):
...
 
         # append syscall to list
         self._call_api(func.__name__, params, result, self.ql.reg.arch_pc, ret_addr)
 
-        # update stack pointer
-        self.ql.reg.arch_sp = self.ql.reg.arch_sp + ((param_num + 1) * 4)
-
         if not passthru and self.PE_RUN:
+            # update stack pointer
+            self.ql.reg.arch_sp = self.ql.reg.arch_sp + ((param_num + 1) * 4)
```
learn-more, line 317: If the original binary is handling the implementation (passthru), the original function should also clean up the stack.

qiling/os/windows/dlls/ntoskrnl.py
nullableVoidPtr: Can't this be `params.get('Exclusive') is not None`?

qiling/os/utils.py
wtdcode: No problem.

wtdcode:
Thanks for your PR. Could you add some tiny tests?

qiling/os/windows/dlls/ntoskrnl.py
learn-more: That will not handle the case of Exclusive being `False` correctly

learn-more:
> Thanks for your PR. Could you add some tiny tests?

I asked about this in the telegram group, and it was suggested there to put this in a separate PR.

That PR will be created when #397 is merged, since that functionality is needed.

PR #610, opened 2020-12-23
Massively refactored and improved UEFI

elicn:
UEFI code greatly revamped, including:
- Massive refactoring and cleanup to the existing code
- Files and folders restructured to be a bit more similar to EDK codebase
- Improvements to existing emulated UEFI functionality
- Implemention of additional UEFI protocols and functionality
- Changes to make the code more "Pythonic"


xwings:
@liba2k @assafcarlsbad  maybe you guys wanna comment :)

qiling/profiles/uefi.ql
```diff
@@ -1,52 +1,37 @@
...
+Guid = 7739f24c-93d7-11d4-9a3a-0090273fc14d
+Table = 0x00000000ffffffff
 
 [DXE_SERVICE_TABLE]
-guid = 05ad34ba-6f02-4214-952e-4da0398e2bb9
-data1 = 0x05AD34BA
-data2 = 0x6F02
-data3 = 0x4214
-data4 = [0x95, 0x2E, 0x4D, 0xA0, 0x39, 0x8E, 0x2B, 0xB9]
+Guid = 05ad34ba-6f02-4214-952e-4da0398e2bb9
```
xwings, line 25: why Guid, Table and not guid, table ?

tests/test_uefi.py
```diff
@@ -102,7 +102,7 @@ def my_onexit(ql, address, params):
             ql.run()
 
             self.assertEqual(0, self.set_api)
-            self.assertEqual(21475885153, self.set_api_onenter)
+            self.assertEqual(2014052449, self.set_api_onenter)  # magic value was 21475885153 (?)
```
xwings, line 105: we need @lib2k and @assafcarlsbad to confirm the value.

xwings (commented):
Try not to put caps on the filename ?

xwings:
@elicn thanks for the PR. overall its great. But, this is a rather long PR and i guess all of us will take sometime to review/update before can merge it.

Once again, thanks!

qiling/profiles/uefi.ql
elicn: The code convention in EDK for names is to start with a capital letter. Since these entries replace specific UEFI variables with these names, they were changed to fit the names of the original vars they replace.

tests/test_uefi.py
elicn: My guess is that the value was derived from the original stack / heap locations that were changed. But I couldn't really find out..

elicn:
Yes, of course.
I just played around with the UEFI emulation and customized it to my needs.. didn't think much about whether it is going to be pushed upstream. Soon enough I found the changes to be quite substantial and mutually dependent, so I couldn't spearate them into different commits anymore.

IMOH, it would be easier to review the new code rather the specific changes. Let me know where clarifications are needed.
 
> 
> Try not to put caps on the filename ?

As I mentioned above, the EDK naming convention is first capital letter - which applies to filenames as well. While refactoring the code I gathered code and definitions of the same logical component to their corresponding filenames as they appear in the EDK codebase. Hence the filenames with first uppercase letter. For logical components that couldn't be moved I used the Python convension of all-lowercase filenames.

qiling/loader/pe_uefi.py
```diff
@@ -153,188 +143,156 @@ def map_and_load(self, path, execute_now=False):
...
         self.ql.mem.write(self.end_of_execution_ptr, b'\xcc')
-        system_table_heap_ptr += pointer_size
         self.ql.hook_address(hook_EndOfExecution, self.end_of_execution_ptr)
 
         self.execute_next_module()
 
     def restore_runtime_services(self):
-        self.ql.mem.write(self.runtime_services_ptr, convert_struct_to_bytes(self.runtime_services))
-
+        pass # not sure why do we need to restore RT
```
liba2k, line 298: When running multiple modules, some of them overwrite some runtime_services functions. Then we lose out hooks.

tests/test_uefi.py
liba2k: No clue, I didn't write this file. Looks like it was added in this commit: https://github.com/qilingframework/qiling/commit/a4e560ef75a7fddd1a1d6ce93212a6e8451c2fef

liba2k (commented):
I like the separate contexts a lot. We were trying to only add UEFI spec common protocols to qiling and implement rare or OEM specific ones in our external projects. If you can, please go over the protocols you are adding and see if they are relevant to the rest of the community and not just for your current project/rom.

qiling/os/uefi/smst.py
```diff
@@ -0,0 +1,325 @@
...
+
+	return EFI_NOT_FOUND
+
+@dxeapi(params = {
+	"Protocol"		: GUID,		# PTR(EFI_GUID)
+	"Event"			: POINTER,	# EFI_EVENT
+	"Registration"	: POINTER	# PTR(PTR(VOID))
+})
+def hook_SmmRegisterProtocolNotify(ql, address, params):
+	event = params['Event']
```
liba2k, line 216: I've just found a bug (my bug), `SmmRegisterProtocolNotify` takes a function pointer, not an event like `RegisterProtocolNotify`. This will be a good opportunity to fix it if you can. Thank you! 
`EFI_STATUS
EFIAPI
SmmRegisterProtocolNotify (
  IN  CONST EFI_GUID              *Protocol,
  IN  EFI_SMM_NOTIFY_FN           Function,
  OUT VOID                        **Registration
  );`

elicn: Yes, looks like a copy & paste error from its BS counterpart.
It looks like it would require implementing an additional notification list, but I must admit I haven't fully understood the existing mechanism. I applied a temp fix for now, and commented out the existing code.

elicn:
Rebased on top of updated dev branch

qiling/os/uefi/smst.py
xwings: I will wait for the commit :) before i merge

xwings:
Looks like there is a test failure. Will you be able to check ?

liba2k:
> Looks like there is a test failure. Will you be able to check ?

I was looking at the error and it looks like the function pointer stored as the unload function in the LOADED_IMAGE_PROTOCOL struct of TcgPlatformSetupPolicy is corrupt, @elicn any idea why?

elicn:
@liba2k: TL;DR: it's not a bug - it's a feature (: ... however the test should be updated

The `test_x8664_uefi_santizier` method passes a value of 1 to `FaultType`, which is used to determine which fault should be generated at runtime. According to the table shown in the `usage` method, that should be mapped to `POOL_UNDERFLOW_MEMCPY`.
In the execution trace below (emitted through a `hook_code`); you may notice that this is more or less what the code does:
1. Gets a pointer to a buffer allocated on the heap by the Python code
2. Decreases the pointer by one
3. Calls `SetMem` to set 8 bytes to `0xaa` to trigger the underflow

When allocating that buffer, the heap returns the next available chunk, `0x78040190`, which happens to be located right after `EFI_LOADED_IMAGE_PROTOCOL` (the last structure to be allocated, ends at `0x7804018f`). Decreasing that pointer by one leads to overwriting the last byte (highest byte of the `Unload` pointer) with `0xaa`. So the memory corruption was intentional, but should have been done elsewhere.
```
[+] [utils.py:351]      0x78000238: GetVariable(VariableName = "FaultType", VendorGuid = "8be4df61-93ca-11d2-aa0d-00e09803aabb", Attributes = 0x77ffffa8, DataSize = 0x77ffffb8, Data = 0x77ffff90) = 0x0
 | 78000238 : 3802                 cmp        byte ptr [rdx], al             | rdx = 0x77000600, al = 0x0
 | 77000311 : 8a442430             mov        al, byte ptr [rsp + 0x30]      | rsp = 0x77ffff60
 | 77000315 : 88442434             mov        byte ptr [rsp + 0x34], al      | rsp = 0x77ffff60, al = 0x1
 | 77000319 : 807c243400           cmp        byte ptr [rsp + 0x34], 0       | rsp = 0x77ffff60
 | 7700031e : 744e                 je         0x7700036e                     | rflags = 0x0
 | 77000320 : 807c243401           cmp        byte ptr [rsp + 0x34], 1       | rsp = 0x77ffff60
 | 77000325 : 747b                 je         0x770003a2                     | rflags = 0x44
 | 770003a2 : 488b442438           mov        rax, qword ptr [rsp + 0x38]    | rsp = 0x77ffff60
 | 770003a7 : 48ffc8               dec        rax                            | rax = 0x78040190  <--
 | 770003aa : 488b8c2498000000     mov        rcx, qword ptr [rsp + 0x98]    | rsp = 0x77ffff60
 | 770003b2 : 488b4960             mov        rcx, qword ptr [rcx + 0x60]    | rcx = 0x78000000
 | 770003b6 : 48894c2468           mov        qword ptr [rsp + 0x68], rcx    | rsp = 0x77ffff60, rcx = 0x78000078
 | 770003bb : 41b0aa               mov        r8b, 0xaa                      | 
 | 770003be : 488b542440           mov        rdx, qword ptr [rsp + 0x40]    | rsp = 0x77ffff60
 | 770003c3 : 488bc8               mov        rcx, rax                       | rax = 0x7804018f   <--
 | 770003c6 : 488b442468           mov        rax, qword ptr [rsp + 0x68]    | rsp = 0x77ffff60
 | 770003cb : ff9068010000         call       qword ptr [rax + 0x168]        | rsp = 0x77ffff60, rax = 0x78000078
[+] [utils.py:351]      0x780001e0: SetMem(Buffer = 0x7804018f, Size = 0x8, Value = 0x77ffffaa)
```
As far as I understand it, the Python code is OK and the test should be modified to allocate a buffer somewhere else to avoid stepping on its own toes.

As a side note, I spent a good deal of time trying to understand why the test code says `POOL_UNDERFLOW_MEMCPY` but `SetMem` is actually called. Since `CopyMem` and `SetMem` are adjacent members of the BS structures, I tried to understand why this is happening.. RE the binary and tracing the execution show clearly that was actually meant to be `SetMem`..

elicn:
The test is now fixed: it doesn't crash anymore and the magic value was replaced with its original meaning, which is an address of the allocated buffer for `CopyMem`. Now the test wouldn't fail if the heap is located somewhere else.

tests/test_uefi.py
```diff
@@ -25,30 +23,29 @@ def enable_sanitized_heap(ql, fault_rate=0):
             ql.os.heap.bad_free_handler = lambda *args: my_abort("Double free or bad free detected")
             ql.os.heap.uaf_handler = lambda *args: my_abort("Use-after-free detected")
 
+            # make sure future allocated buffers are not too close to UEFI data
+            ql.os.heap.alloc(0x1000)
```
liba2k, line 27: I'm not sure I understand it correctly, but shouldn't we reserve some padding space as part of the loader instead of every script doing it?

elicn: Normally there shouldn't be a problem. This test deliberately underflows to the last byte of the previous chunk. Since the heap returns the first address available on allocation and only UEFI structures were allocated to that point, that previous chunk happens to belong to UEFI structures. 

To make sure we can still test for underflow without corrupting UEFI structures (like happened to the Unload pointer), we add a dummy chunk. I don't think it should be added automatically as a safety measure.

xwings:
@elicn will you be able to update the file with the standard header :)

tests/test_uefi.py
liba2k: Thank you very much for the detailed explanation!

elicn:
@xwings I am not sure that I follow.. which file are you referring to?

xwings:
> @xwings I am not sure that I follow.. which file are you referring to?

We have the standard "header" for all the files

```
#!/usr/bin/env python3
# 
```

xwings:
Not many files. I will add it. :)

Thanks for the PR and Happy New Year!

PR #611, opened 2020-12-24
Support for custom env on IDA plugin

zi0Black:
Support for custom env on IDA plugin:

- Added the new field inside the GUI
- used `eval` to obtain a valid object also if you need to place a lot of 0x41 ;)

-----


wtdcode:
Hi! Thanks for your PR.

Your commits log seems quite strange, would you like to make a cleaner one?

zi0Black:
Yep, give me 10 minutes.

PR #612, opened 2020-12-24
Support for custom env on IDA plugin

zi0Black:
Support for custom env on IDA plugin:

- Added the new field inside the GUI
-  `eval` is used to obtain a valid object also if you need to place a lot of 0x41 ;)

-----

### DIFF
--- a/qiling/extensions/idaplugin/qilingida.py
+++ b/qiling/extensions/idaplugin/qilingida.py
@@ -71,11 +71,11 @@ class Colors(Enum):
 class IDA:
     def __init__(self):
         pass
-    
+
     @staticmethod
     def get_function(addr):
         return ida_funcs.get_func(addr)
-    
+
     @staticmethod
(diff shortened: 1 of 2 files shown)


kabeor (approved):
Very useful， thx

qiling/extensions/idaplugin/qilingida.py
```diff
@@ -2051,14 +2055,22 @@ def ql_set_rootfs(self):
...
         rootfspath = setupdlg.path_name.value
         customscript = setupdlg.script_name.value
+        env = setupdlg.env_var.value
 
         if customscript != '':
             self.customscriptpath = customscript
 
+        if env != '':
+            try:
+                self.qlemu.env = eval(env)
```
wtdcode, line 2065: Eval is evil for python codes. Please use json for this form field.

kabeor:
@zi0Black save env path in a json file plz.

PR #613, opened 2020-12-25
IDA plugin now support custom envs (from file)

zi0Black:
Added support for custom envs variables, for IDA plugin inside qilingida.py.

Updated ChangeLog.

-----


kabeor (approved):
Thx, Merry Christmas.

PR #614, opened 2020-12-25
Simplify syscall errno

nullableVoidPtr:
- [x] Have you checked that there aren't other open [pull requests](https://github.com/qilingframework/qiling/pulls) for the same update/change?
- [x] Is this pull request for [dev](https://github.com/qilingframework/qiling/tree/dev) branch? (Do not pull request for **master** branch).

-----

Currently, each implementation of syscalls would need to call a function setting the syscall `errno`, but with this PR, only the dispatcher has to do it, and the handlers can easily return early.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -7,7 +7,6 @@ BREAK CHANGE
 - ql.filename is renamed to ql.argv.
 - ql.output and ql.verbose now has slightly different meanings and can be adjusted runtime. See their docstring for details.
 - ql.filter now accepts a regular expression.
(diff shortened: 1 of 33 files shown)


nullableVoidPtr:
Currently, `test_gdb.py` is broken. I'll have to look into what's causing this.

nullableVoidPtr:
Hmm, it seems to be fixed now. I think the issue was on my end with gdb and port allocation.

xwings:
ok, cool.

thanks!

PR #615, opened 2020-12-25
Add some basic tests for Windows drivers

learn-more:
- Run the same tests both on x86 and x8664
- Add function to clear the stored syscalls
- Add accessor for DriverUnload function. (Not used yet, other than checking it is initialized correctly).


qiling/os/os.py
```diff
@@ -299,7 +305,7 @@ def _call_api(self, name, params, result, address, return_address):
             "position": self.syscalls_counter
         })
 
-        self.ql.os.syscalls_counter += 1
```
wtdcode, line 302: Any reason to move this property to `ql`?

learn-more: I did not move this property, but since we are already in the `os` object, this just removes an indirection where it would use `self.ql.os.` instead of `self.`

wtdcode: Oh, got it.

wtdcode:
Thanks for your tests. I will merge this one after we update our TravisCI.

Again, please remember to update Changelog.

learn-more:
> Thanks for your tests. I will merge this one after we update our TravisCI.
> 
> Again, please remember to update Changelog.

Ok!

I have updated the ChangeLog.

tests/test_pe.py
```diff
@@ -554,5 +554,96 @@ def hook_stop_address(ql):
             ql.amsint32_driver.run(0x102D0)
 
 
+    def test_pe_win_x86_driver(self):
```
learn-more, line 557: Would it be better to add this in a new file, for example: `test_pe_driver.py` ?

xwings: here should be fine. since unless we have more.

again if we wanna move (which is ok), we need to move linux and mac too. just to be in sync.

xwings:
also, please add your name into credits.txt, contributors sections

xwings:
We are still waiting for response from travis-ci. Or else i gonna spend some money to get their plan.

learn-more:
> We are still waiting for response from travis-ci. Or else i gonna spend some money to get their plan.

Take your time, the most time a PR from me is open on this repository is because I did not update it yet.

PR #617, opened 2020-12-26
Add hook_CommandLineToArgvW  (not tested)

baptistapedro:
Did not test it but I guess the logic is close... This might also come as a reminder/warning of a missing windows api.
-----

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,4 +1,4 @@
-Features request and TODO please refer to issue [#333](https://github.com/qilingframework/qiling/issues/333)
+Features request and TODO please refer to issue 333 https://github.com/qilingframework/qiling/issues/333
 
 -------------------- CHECKLIST before TAG --------------------
 Release 


xwings:
https://docs.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-commandlinetoargvw

should be in shellapi ?

and you pick the wrong branch. Please refer to https://docs.qiling.io/en/latest/contribution/

qiling/os/windows/dlls/kernel32/processenv.py
```diff
@@ -37,6 +37,25 @@ def hook_GetCommandLineW(ql, address, params):
...
+# );
+
+# Note: Didn't test it. Need second thoughts.
+@winsdkapi(cc=STDCALL, dllname=dllname)
+def hook_CommandLineToArgvW(ql, address, params):
+        cmdline = ql.loader.cmdline + params["lpCmdLine"]
+        cmdline = ql.loader.cmdline.decode('ascii').encode('utf-16le')
+        numArgs = ql.os.heap.alloc(len(cmdline))
+        ptrs = ql.mem.write(numArgs, cmdline)
+        return ptrs
```
learn-more, line 54: This implementation is incorrect.
It should parse the passed in string in 'lpCmdLine'.
Only if that is an empty string will it return the current process executable as only argument.

And it parses the `lpCmdline` to split it into multiple arguments, which is not something you do here.

learn-more: See https://git.reactos.org/?p=reactos.git;a=blob;f=dll/win32/shell32/wine/shell32_main.c;hb=9c8ed68544f1adf49fb941d83a807159433e4323#l79 for an example c implementation

xwings:
Close due to no response from author since Dec 2020

PR #618, opened 2020-12-27
Allow enabling disassembly output at runtime

learn-more:
This allows the user to turn disassembly output on/off whenever they want, instead of only being able to configure this for the entire session.

For example, loading a dll could be done on 'info',
and then before a second call to `run` (to call an export for example) the output level can be set to `disasm`

### DIFF
--- a/CREDITS.TXT
+++ b/CREDITS.TXT
@@ -49,7 +49,7 @@ iamyeh
 alfink
 bambu
 bkerler (viperbjk)
-learn-more
+Mark Jansen (learn-more)
 cq674350529
 
 Alpha testers (in no particular order, named by github id)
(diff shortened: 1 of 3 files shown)


xwings:
what does this do again ?

learn-more:
> what does this do? :) screenshot or somethibg,

I have elaborated a bit in the PR.

xwings:
Cool, that makes sense now.

Would you also mind to add it in some quick writeup in docs.qiling.io too?

learn-more:
> Cool, that makes sense now.
> 
> Would you also mind to add it in some quick writeup in docs.qiling.io too?

Sure, I'll add that later today.

learn-more:
> Cool, that makes sense now.
> 
> Would you also mind to add it in some quick writeup in docs.qiling.io too?

I have created https://github.com/qilingframework/rtfd.io/pull/29 for that (from the web editor, please check that the output is good)

kabeor:
Cool, thx.

PR #619, opened 2020-12-27
Some performance improvements

elicn:
- QlCoreStructs: Resolve pack / unpack formats once instead of every time methods are called
- print_function: Reduce string copy / recreation by leveraging faster Python3 f-strings

### DIFF
--- a/qiling/core_struct.py
+++ b/qiling/core_struct.py
@@ -32,7 +32,7 @@ class QlCoreStructs:
 		}
 
 		if bit not in handlers:
-			raise QlErrorStructConversion("[!] Unsupported Qiling conversion!")
+			raise QlErrorStructConversion("[!] Unsupported Qiling struct conversion")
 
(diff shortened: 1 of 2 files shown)


nullableVoidPtr:
Do you think there are some functions which would greatly benefit from the `functools.lru_cache` decorator?

elicn:
Haven't seen any so far.
`print_function`, however, is not a good candidate for that since `lru_cache` caches results and `print_function` yields different results based on the emulated parameters and their values.

xwings:
Thanks!

Will merge now.

xwings:
i revert back to the old print_function. it did not pass the ci test

elicn:
Couldn't understand what exactly went wrong on the Windows run.. I can see CI complaining it failed, I just don't see any assertion failure there. Could you please point out what's wrong?

elicn:
Found and fixed the problem.
Resubmitted in a new PR.

PR #623, opened 2020-12-29
Various fixes

learn-more:
- Include exception on module load failure
- Split out instructions over multiple lines when disassembling
- Fix a warning related to the binary cache
- Explicitly set some properties in the constructor

One commit per feature, so that it's easy to only include the ones you approve of.

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -206,6 +206,14 @@ class Qiling(QlCoreHooks, QlCoreStructs):
         # Add extra guard options when configured to do so
         self._init_stop_guard()
 
+        ###############################
+        # Properties configured later #
+        ###############################
+        self.entry_point = None
+        self.exit_point = None
+        self.timeout = None
+        self.count = None
+        self._initial_sp = None
     
     #####################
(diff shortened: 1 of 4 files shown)


qiling/core.py
```diff
@@ -206,6 +206,14 @@ def __init__(
         # Add extra guard options when configured to do so
         self._init_stop_guard()
 
+        ###############################
+        # Properties configured later #
+        ###############################
```
xwings, line 211: all these could be in os ?

learn-more: This is merely documenting which properties are set by various other functions later on :)

xwings: i am trying to "slim" down core.py. and move os/arch related properties out from core.py

xwings: ah, i might misunderstand what you trying to do. 

again, what is this for.

xwings: Ok, i guess we move this up after definition, line 85

xwings:
I will move it :)

Issue #620, opened 2020-12-28
Ranged hooks do not intercept memory accesses crossing into range


elicn:
**Observed behavior**
Ranged hooks set-up with `hook_mem_read` and `hook_mem_write` will be fired up only when the memory access _starts_ within the hooked range. Memory accesses that start below the hooked range and crosses into it, will go unnoticed. E.g. for a hooked memory range `[0x2000, 0x3000]`, a QWORD read starting at `0x1ffc` will be missed.

This is probably because the `bound_check` method of the `Hook` class (defined at `core_hooks.py`) does not take the access size into account.

**Expected behavior**
Ranged hooks are expected to intercept memory accesses that cross the range, not only ones that starts within it.

**Sample Code**
Attached a sample test code; change `QILING_HOME` as appropriate.
The code demonstrates the behavior for read accesses, but can be easily modified to show write behavior as well.
[memhook.zip](https://github.com/qilingframework/qiling/files/5747855/memhook.zip)

xwings:
I need to take a look. might be something on unicorn that we need to fix ? or a Qiling bug ?

elicn:
To my understanding this is Qiling. See my comment regarding the `bound_check` method.

xwings:
my bad. hook_mem_read. not mem.read i must be blind.

let me take a look and get back to you.

PR #625, opened 2020-12-30
A few more improvements

elicn:
- QlCoreStructs: Added missing 8 bits pack / unpack
- print_function: Reduce string copy / recreation by leveraging faster Python3 f-strings (resubmitted, fixed bad `bytearray` handling)
- Fix issue #620


xwings:
Nice! Thanks!

PR #627, opened 2020-12-31
fix stack setup in elf.py

ucgJhe:
## Problem description
according to this
https://github.com/qilingframework/qiling/blob/58a73b173f5e9f48241d5cb43cad7229683438c0/qiling/loader/elf.py#L143-L153

argv will be encoded before copy to stack, its working fine while all inputs are all ASCII character.

for those non-ASCII character will be converted to UTF-8 and messed up with original input

this is an easy bof example with input `'A'*0xcc + 'B'*4`
image

before this commit with input `'A'*0xcc + '\xc2'*4`, you can see the input `\xc2` were converted to `\xc3\x82`
image

after this commit with input `b'A'*0xcc + b'\xc2'*4`
image

## conclusion

- this commit made qiling accepts bytes input

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -140,16 +140,8 @@ class QlLoaderELF(QlLoader, ELFParse):
         l_addr = []
         s_addr = addr
         for s in strs:
-            bs = s.encode("utf-8") + b"\x00"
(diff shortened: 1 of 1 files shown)


xwings:
Thanks !

PR #628, opened 2020-12-31
check shellcoder first

cq674350529:
Check `ql.shellcoder` first to avoid unnecessary profile variable parse.

When try to run a FreeBSD x86 shellcoder, since there is no `OS32` entry in `freebsd.ql` currently, error will occur.
It's possible to add an `OS32` entry to the `freebsd.ql` to solve this issue. However, this way is more logical.

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -93,21 +93,19 @@ class QlLoaderELF(QlLoader, ELFParse):
         self.ql = ql
 
     def run(self):
+        if self.ql.shellcoder:
+            self.ql.mem.map(self.ql.os.entry_point, self.ql.os.shellcoder_ram_size, info="[shellcode_stack]")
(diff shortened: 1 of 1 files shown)


xwings:
i think we need to add OS32 to freebsd.

Since profile stack address is usable for shellcode ?

cq674350529:
According to the code, I believe the profile stack address is not usable for shellcode:
1. the `arch_sp` is initialized by another value `self.ql.os.entry_point + 0x200000 - 0x1000` 
2. the profile stack address is assigned to local variable `stack_address`, and have no more references when handling  `self.ql.shellcoder` logic. 

https://github.com/qilingframework/qiling/blob/97a9b59cd03d829d9e618edf34cbec5ee9d47943/qiling/loader/elf.py#L96-L110

xwings:
fair enuf :)

cool and i am old.

Happy New Year.

PR #639, opened 2021-01-10
OS improvements

elicn:
Some improvements to the OS module:
- New and improved function arguments handling
  - Selects between `amd64`, `cdecl` and `ms` calling conventions instead of just 32 vs 64 bits
  - Uses a unified generic handling instead of re-selecting appropriate handling every single time
  - Supports up to 8 function call arguments in each cc
- Code cleanup and tidy-up
- Documentation

Note that not all the OS methods are triggered in the current test suite. This may need a more aggressive testing

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/os.py
+++ b/qiling/os/os.py
@@ -3,15 +3,16 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
-import logging, os, sys, types
+import sys, logging
+from typing import Callable, Sequence, Mapping, MutableMapping, Any
 
-from .utils import QlOsUtils
 from .const import *
 from .filestruct import ql_file
 from .mapper import QlFsMapper
+from .utils import QlOsUtils
 
-from qiling.const import *
-
+from qiling.const import QL_ARCH, QL_OS, QL_INTERCEPT, QL_OS_POSIX
+from qiling.exception import QlErrorArch
 from unicorn.x86_const import *
 
 class QlOs(QlOsUtils):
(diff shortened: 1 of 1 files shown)


xwings:
Ah, I did not see this PR. and I updated the header.

Can you please update your code?

elicn:
Rebased

elicn:
It worth mentioning that `QlOs` is extended by several classes, but doesn't serves well as a base class: its implementation is very Intel-oriented, which is irrelevant to some other OS classes, and some of the OS "core" features are unnecessarily re-implemented by the extending classes. Consider extracting the Intel stuff out of the `QlOs` class, providing only generic implementation to OS derivatives - and on the other hand, minimize unnecessary implementation duplication across OS derivatives.

Another thing I realized when I worked on this class, is that some OS derivatives share the same `print_function` method, but use different / ambiguous type tags for the parameters. That may cause some of the parameters to be interpreted in an unexpected way (e.g. `BOOL` is 8-bits on some systems, but 32-bits on others; `VA_ARGS` may be implemented in different ways).

xwings:
totally agree.

The story behind is: different people wrote the code at the same time. I guess you can see the coding style is different.

Refactoring is always on going and i guess we got long way to go.

Feel free to join in the fun :)

xwings:
these code will be affecting ida pro, @kabeor  needs to check before we can merge it

kabeor:
IDA Pro 7.5 test pass

xwings:
Cool :)

xwings:
I revert this merge. Looks like it kill sality test. let me try to bring sality back to the CI

elicn:
Sality tries to access a function call parameter beyond the 12th.
Fixed by increasing supported number of arguments to 16 to make it happy.
Resubmitting the fixed version.

xwings:
I am still having issue with others, sality too. I still cannot replicate in my box locally. let me check

elicn:
Sality is totally broken.
I'll fix it too.

Issue #642, opened 2021-01-15
AttributeError: ql_file object has no attribute fcntl


cq674350529:
**Describe the bug**
When trying to run `xz -d <file>`, an error occurred.
```shell
Traceback (most recent call last):
  File "qiling_demo.py", line 780, in <module>
    qiling_test()
  File "qiling_demo.py", line 771, in qiling_test
    ql.run()
  File "/home/cq/workspace/git/qiling/qiling/core.py", line 765, in run
    self.os.run()
  File "/home/cq/workspace/git/qiling/qiling/os/linux/linux.py", line 149, in run
    self.ql.emu_start(self.ql.loader.elf_entry, self.exit_point, self.ql.timeout, self.ql.count)
  File "/home/cq/workspace/git/qiling/qiling/core.py", line 957, in emu_start
    raise self._internal_exception
  File "/home/cq/workspace/git/qiling/qiling/utils.py", line 157, in wrapper
    return func(*args, **kw)
  File "/home/cq/workspace/git/qiling/qiling/core_hooks.py", line 119, in _hook_intr_cb
    ret = h.call(ql, intno)
  File "/home/cq/workspace/git/qiling/qiling/core_hooks.py", line 35, in call
    return self.callback(ql, *args)
  File "/home/cq/workspace/git/qiling/qiling/os/linux/linux.py", line 71, in hook_syscall
    return self.load_syscall(intno)
  File "/home/cq/workspace/git/qiling/qiling/os/posix/posix.py", line 183, in load_syscall
    raise e
  File "/home/cq/workspace/git/qiling/qiling/os/posix/posix.py", line 171, in load_syscall
    ret = self.syscall_map(self.ql, self.get_func_arg()[0], self.get_func_arg()[1], self.get_func_arg()[2], self.get_func_arg()[3], self.get_func_arg()[4], self.get_func_arg()[5])
  File "/home/cq/workspace/git/qiling/qiling/os/posix/syscall/fcntl.py", line 121, in ql_syscall_fcntl64
    ql.os.fd[fcntl_fd].fcntl(fcntl_cmd, fcntl_arg)
AttributeError: 'ql_file' object has no attribute 'fcntl'
```

**Sample Code**
```python
rootfs_path = "/home/cq/Desktop/qiling_tmp"
bin_path = "xz"
tar_path = "hello.txt.xz"

[...]

Issue #640, opened 2021-01-12
pe.py set_cmdline is incorrect


learn-more:
- The allocation on the heap is done for a char if the target is 32 bit, and as wchar if the target is 64 bit.
- For every exported symbol for every dll loaded there is an allocation done on the os.heap.

xwings:
Once i have time i will take a look, or we need someone to take over the windows part :)

wolfcod:
The cmdline is an attribute available in PEB.ProcessParameters structure, that is missing.
When the program is loaded, this structure must be initialized and GetCommandLineA/GetCommandLineW return a pointer to cmdline itself. DLLs's don't need a pointer to cmdline, because is not required into DllMain, they can retrieve cmdline only through api. It's an argument required by WinMain.

learn-more:
> The cmdline is an attribute available in PEB.ProcessParameters structure, that is missing.
> When the program is loaded, this structure must be initialized and GetCommandLineA/GetCommandLineW return a pointer to cmdline itself. DLLs's don't need a pointer to cmdline, because is not required into DllMain, they can retrieve cmdline only through api. It's an argument required by WinMain.

That is not wrong, but also completely unrelated to this.

learn-more:
Fixed, thanks.

PR #645, opened 2021-01-17
add fcntl for ql_file

cq674350529:
try to fix https://github.com/qilingframework/qiling/issues/642


cq674350529:
The failed test seems to be related to https://github.com/qilingframework/qiling/issues/640

qiling/os/filestruct.py
```diff
@@ -50,6 +48,12 @@ def close(self):
     def fstat(self):
         return Fstat(self.__fd)
     
+    def fcntl(self, fcntl_cmd, fcntl_arg):
+        try:
+            return fcntl.fcntl(self.__fd, fcntl_cmd, fcntl_arg)
+        except Exception:
+            pass
+
```
xwings, line 56: why try and except. anything we need to worry about ?

cq674350529: There is a chance that the module `fcntl` may be imported unsuccessfully, as shown in the begin of the file. So maybe this `try-except` is used for this?

xwings: i guess is smart for us to remove all of this
i mean try import above and try import here.

cq674350529: I try to remove both `try-except` catch, and test failed on windows for no module `fnctl`. There is another `fcntl.ioctl()` in `qiling/os/filestruct.py`, I'm not sure if it's used ot not.  So I try another way: only call `fcntl()` on ql_socket, which is submitted in another pull request.

xwings:
let me take a look and see how can we fix this mess :)

PR #646, opened 2021-01-17
OS Improvements

elicn:
[Resubmitting #639 with additional fixes]

Some improvements to the OS module:
- New and improved function arguments handling
  - Selects between `amd64`, `cdecl` and `ms` calling conventions instead of just 32 vs 64 bits
  - Uses a unified generic handling instead of re-selecting appropriate handling every single time
  - Supports up to 16 function call arguments in each cc
- Code cleanup and tidy-up
- Documentation

Fixes to Sality (both test and example):
- BUGFIX: `os.windows.fncc` could not find DLL functions hooked by sality and assign them with parameters, as it expects a name prefix of 5 chars (maybe `"hook_"` ? https://github.com/qilingframework/qiling/blob/5ddb8d68cc9ea518902d4c9fccf9d49a57c186b3/qiling/os/windows/fncc.py#L42)
- BUGFIX: `StartServiceA` is defined in `advapi32.dll` and not in `kernel32.dll`
- BUGFIX: Added support for `HC_HANDLE` parameters type needed for `StartServiceA`
- Reduced namespace pollution
- Code cleanup

### DIFF
--- a/examples/sality.py
+++ b/examples/sality.py
@@ -29,7 +29,7 @@ def init_unseen_symbols(ql, address, name, ordinal, dll_name):
 #   LPDWORD                 lpThreadId
 # );
 @winsdkapi(cc=STDCALL, dllname='kernel32_dll')
-def sality_CreateThread(ql, address, params):
+def hook_CreateThread(ql, address, params):
     # set thread handle
     return 1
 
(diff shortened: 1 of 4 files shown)


xwings:
Thanks!

PR #647, opened 2021-01-17
Acmdln

### LINKED ISSUE #640: pe.py set_cmdline is incorrect

learn-more:
Fix _acmdln and _wcmdln handling.
Generix x64 printf fixes.
Includes a test (with source)

Should fix #640


qiling/os/utils.py
```diff
@@ -371,7 +371,10 @@ def printf(self, address, fmt, params_addr, name, wstring=False):
                 params.append(
```
learn-more, line 371: This is not valid on x64 for normal arguments, only when an argument list is passed in!

qiling/os/windows/dlls/msvcrt.py
```diff
@@ -188,7 +195,7 @@ def hook_sprintf(ql, address, _):
     p_args = sp + ql.pointersize * 3
 
     format_string = ql.os.read_cstring(format_ptr)
-    str_size, str_data = ql.os.printf(address, format_string, p_args, "sprintf")
+    str_size, str_data = ql.os.vprintf(address, format_string, p_args, "sprintf")
```
learn-more, line 198: This is probably still not correct on x64, but it behaves as before!

xwings:
awesome!

windows is back on track!

PR #649, opened 2021-01-17
Print warnings after loading a dll

learn-more:
`pefile` has a builtin method to retrieve warnings,
so when there are warnings display those.


xwings:
thanks !

PR #650, opened 2021-01-18
call fcntl() only on ql_socket

cq674350529:
try to solve https://github.com/qilingframework/qiling/issues/642 again


xwings:
Do we still need commit #645 ?

cq674350529:
Not needed if this fix is in right direction

xwings:
cool, can you please close the PR ?

cq674350529:
Ok, I closed that PR.

PR #652, opened 2021-01-18
Fix Sality

elicn:
- Fixed a bug that affected Sality
- Other minor fixes


xwings:
Cool, let me try.

thanks!

PR #654, opened 2021-01-19
Rewrite PE cache to only store relevant symbols

### LINKED ISSUE #641: pe.py libcache contains too much symbols
When storing a file for the libcache, not only the symbols for that file are stored, but also all symbols loaded until then.

```python
    with open(fcache, "rb") as fcache_file:
        data, cmdlines, import_symbols, import_address_table = pickle.load(fcache_file)
    print(fcache.name, 'import_address_table:', import_address_table.keys())
```
Prints something like:

```
wsock32.dll.10b4b000.cache import_address_table: dict_keys(['ntdll.dll', 'kernel32.dll', 'msvcrt.dll', 'user32.dll', 'advapi32.dll', 'gdi32.dll', 'ole32.dll', 'oleaut32.dll', 'wininet.dll', 'wsock32.dll'])
```
Because all those dll's were loaded in the target before wsock32.dll was loaded.

Same with the `import_symbols`, they also contain _all_ import symbols up to the point the cache is written.

learn-more:
- Refactored it into a class that can be customized
- Added a test validating cache behavior

This fixes #641

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -15,6 +15,35 @@ from qiling.arch.x86_const import *
 from .loader import QlLoader
 from qiling.os.memory import QlMemoryHeap
 
+
+class QlPeCacheEntry:
+    def __init__(self, data, cmdlines, import_symbols, import_table):
+        self.data = data
+        self.cmdlines = cmdlines
+        self.import_symbols = import_symbols
+        self.import_table = import_table
+
+# A default simple cache implementation
+class QlPeCache:
+    def create_filename(self, path, address):
+        return path + ".%x.cache2" % address
+
+    def restore(self, path, address):
+        fcache = self.create_filename(path, address)
+        # pickle file cannot be outdated
+        if os.path.exists(fcache) and os.stat(fcache).st_mtime > os.stat(path).st_mtime:
+            with open(fcache, "rb") as fcache_file:
+                return QlPeCacheEntry(*pickle.load(fcache_file))
+        return None
+
+    def save(self, path, address, entry):
+        fcache = self.create_filename(path, address)
+        data = (entry.data, entry.cmdlines, entry.import_symbols, entry.import_table)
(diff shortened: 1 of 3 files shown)


qiling/loader/pe.py
```diff
@@ -15,6 +15,35 @@
...
+    def __init__(self, data, cmdlines, import_symbols, import_table):
+        self.data = data
+        self.cmdlines = cmdlines
+        self.import_symbols = import_symbols
+        self.import_table = import_table
+
+# A default simple cache implementation
+class QlPeCache:
+    def create_filename(self, path, address):
+        return path + ".%x.cache2" % address
```
learn-more, line 29: Since the file format is changed slightly, I made sure that the names are different so it will not try to load old cache files.

qiling/loader/pe.py
```diff
@@ -346,7 +380,10 @@ class QlLoaderPE(QlLoader, Process):
     def __init__(self, ql):
         super(QlLoaderPE, self).__init__(ql)
         self.ql         = ql
-        self.libcache   = self.ql.libcache
+        if type(self.ql.libcache) == bool:
```
learn-more, line 383: Specifying `True` for libcache still works, and it will use the default cache.
Specifying a class for `libcache` (or something callable),
it will directly use that as cache so the user can override it.

qiling/loader/pe.py
```diff
@@ -15,6 +15,35 @@
...
+        self.cmdlines = cmdlines
+        self.import_symbols = import_symbols
+        self.import_table = import_table
+
+# A default simple cache implementation
+class QlPeCache:
+    def create_filename(self, path, address):
+        return path + ".%x.cache2" % address
+
+    def load(self, path, address):
```
xwings, line 31: we are using ql.save and ql.restore.
lets follow here too ?

xwings:
cool. 

Thankssss!

PR #657, opened 2021-01-22
Fix logging initialization

learn-more:
When initializing logger before logging, there is no need to clear the default loggers.
This allows other applications that integrating Qiling to register logger handlers before starting qiling

### DIFF
--- a/qiling/utils.py
+++ b/qiling/utils.py
@@ -485,7 +485,7 @@ def os_setup(archtype, ostype, ql):
 
 
 def profile_setup(ostype, profile, ql):
-    logging.debug("Customized profile: %s" % profile)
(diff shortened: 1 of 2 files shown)


qiling/utils.py
```diff
@@ -485,7 +485,7 @@ def os_setup(archtype, ostype, ql):
 
 
 def profile_setup(ostype, profile, ql):
-    logging.debug("Customized profile: %s" % profile)
+    debugmsg = "Customized profile: %s" % profile
```
learn-more, line 488: Because this was logged before logging initialization, there was a default console logger appended.
When the logging is initialized first, this does not happen and only the registered loggers will be present.

xwings:
Cool. Thanks!

PR #663, opened 2021-01-27
Add callback that will be called on logger creation

learn-more:
This allows to modify the logger after it is created, and before the first event is logged.

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -48,6 +48,7 @@ class Qiling(QlCoreHooks, QlCoreStructs):
             stdin=0,
             stdout=0,
             stderr=0,
+            on_logger_created=None
     ):
         """ Create a Qiling instance.
 
@@ -167,6 +168,9 @@ class Qiling(QlCoreHooks, QlCoreStructs):
                                                               self._multithread)
 
         self.log.setLevel(ql_resolve_logger_level(self._output, self._verbose))
+
+        if on_logger_created:
+            on_logger_created(self, self.log)
 
         # Now that the logger is configured, we can log profile debug msg:
         self.log.debug(debugmsg)


wtdcode:
Could you add your exact use cases?

learn-more:
> Could you add your exact use cases?

```
class LogWindow(logging.Handler):
    def __init__(self):
        super().__init__()
        self._msgs = []
        self._scroll_to_bottom = 0

    def on_logger_created(self, ql, logger):
        formatter = QilingColoredFormatter(ql, FMT_STR)
        self.setFormatter(formatter)
        logger.level = logging.DEBUG
        logger.handlers = []    # Remove the default console logger
        logger.addHandler(self)

    def emit(self, record):
        msg = self.format(record)
        msg = msg.replace('\033[9', '\033[3')   # pyimgui does not support the 'BRIGHT' colors
        self._msgs.append(msg)
        self._scroll_to_bottom = 2

    def clear(self):
        self._msgs = []

    def frame(self):
        imgui.set_next_window_position(10, 440, imgui.FIRST_USE_EVER)
        imgui.begin('Log')
        imgui.set_window_size(1000, 250, imgui.FIRST_USE_EVER)
        if imgui.button('Clear'):
            self._msgs = []
        imgui.separator()
        imgui.begin_child('scrolling', flags=imgui.WINDOW_ALWAYS_AUTO_RESIZE)
        imgui.push_style_var(imgui.STYLE_ITEM_SPACING, imgui.Vec2(0, 1))
        for line in self._msgs:
            imgui.text_ansi(line)
        if self._scroll_to_bottom > 0:
            self._scroll_to_bottom -= 1
            imgui.set_scroll_here(1.0)
        imgui.pop_style_var()
        imgui.end_child()
        imgui.end()
```

And registration:
```
        self.ql = Qiling([ROOTFS_PATH_32 / 'bin' /  'cmdln32.exe'],
                    ROOTFS_PATH_32,
                    libcache=True,
                    on_logger_created=self.log_window.on_logger_created)
```

wtdcode:
hmmm, how about allowing users to pass in a `logger` object to override the default logger?

learn-more:
> hmmm, how about allowing users to pass in a `logger` object to override the default logger?

fine with me, if that is what you prefer

wtdcode:
> > hmmm, how about allowing users to pass in a `logger` object to override the default logger?
> 
> fine with me, if that is what you prefer

Okay, I will push that these days.

PR #665, opened 2021-01-28
Add generic ql.mem.read_ptr helper function

learn-more:


qiling/os/memory.py
```diff
@@ -182,6 +187,14 @@ def restore(self, mem_dict):
     def read(self, addr: int, size: int) -> bytearray:
         return self.ql.uc.mem_read(addr, size)
 
+    def read_ptr16(self, addr):
```
xwings, line 190: i guess we only need read_ptr, since we can identify the bit by the address / or using ql.archbit ?

learn-more: I think it's bad performance wise to do this check on every call to a function like this,
especially considering the recent refactors added a lot of indirections with the new properties.

learn-more: And as an added bonus, if for some reason someone wants to use the specific version, they can.
For example, following indirect 16-bit pointers in structs or something like that.

cq674350529: Maybe `readptr()` with another arg `ptr_size`? If `ptr_size` is not set, then use the `ql.archbit // 8` ?

xwings: > And as an added bonus, if for some reason someone wants to use the specific version, they can.
> For example, following indirect 16-bit pointers in structs or something like that.

That make sense. but having 3 is abit overkill. maybe cq idea make some sense. what do you think?

learn-more: I have updated the PR with your suggestions.
Also added a test for this.

xwings:
pointer_size can be size ?
since read_ptr implies size is ptr size

xwings:
Cool. thanks!

PR #667, opened 2021-01-31
Add support for SMM_RUNTIME_SERVICES_TABLE.

assafcarlsbad:
SMM drivers often need to access some of the functionality offered by the `EFI_RUNTIME_SERVICES` table. However, calling any of these runtime services directly from SMM is a bad idea since that will lead to a so-called [SMM callout vulnerability](http://blog.cr4.sh/2016/02/exploiting-smm-callout-vulnerabilities.html).

To allow SMM drivers use these interfaces without risking security, UEFI implementations usually contain a second copy of the `EFI_RUNTIME_SERVICES` table. This copy is guaranteed to reside in SMRAM and therefore SMI handlers can use it safely. The table itself is pointed to by dedicated `EFI_CONFIGURATION_TABLE` entry which can be identified by the GUID `EFI_SMM_RUNTIME_SERVICES_TABLE_GUID`. This PR simply adds support this table for the purpose of properly emulating the aforementioned SMM drivers.

For an actual example which searches and uses this table, please refer [here](https://github.com/raywugithub/zprj/blob/b7c51c9cf4864df6aabb99a1ae843becd577237c/Board/EM/Isct/Smm/IsctSmm.c).

### Coding convention?

### Extra tests?

### Changelog?

- [x] Changelog has been updated in my PR.

### Target branch?

### One last thing

-----


elicn (commented):
This is mostly copying & pasting existing code around..
Instead of duplicating, is it possible to generalize the existing code to fit both SMM and non-SMM cases?

ChangeLog
```diff
@@ -8,6 +8,7 @@ BREAK CHANGE
 - ql.output and ql.verbose now has slightly different meanings and can be adjusted runtime. See their docstring for details.
 - ql.filter now accepts a regular expression.
 - Remove ql.log_dir, ql.log_split, ql.append but add ql.log_file instead.
+- Add support for SMM_RUNTIME_SERVICES_TABLE.
```
xwings, line 11: wrong place, should be under Jan [something]

xwings:
Ok, I moved it :)

PR #671, opened 2021-02-01
Verbose posix syscall error logging

PeterStolz:
The debug logging of syscall return values just returns the actual number instead of the error name like -0xa instead of SIGCHILD.   
In this version the number is returned and if the error code is known it is printed aswell.
`ioctl() = -0x1` becomes    
`ioctl() = -0x1|EPERM` which saves a lot of time understanding what went wrong.
The actual error codes come from the [official linux kernel repository](https://github.com/torvalds/linux.git) from the files ["tools/include/uapi/asm-generic/errno-base.h"](https://github.com/torvalds/linux/blob/master/tools/include/uapi/asm-generic/errno-base.h) and ["tools/include/uapi/asm-generic/errno.h"](https://github.com/torvalds/linux/blob/master/tools/include/uapi/asm-generic/errno.h) and were added to posix/const.py as a dict.

I also changed SIGCHILD fom -10 to 10, which fixes a bug I intruduced in [my last PR](https://github.com/qilingframework/qiling/pull/600) where the syscall wait returns 10 instead of -10 upon the SIGCHILD error.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----


aquynh:
"|" can be misunderstood as logical OR. Can you come up with something else?

PeterStolz:
Would ":" do the job?

xwings:
I suggest we go for ()  as in ioctl() = -0x1(EPERM)

PeterStolz:
Ok. It looks better now.

xwings:
Thanks!

Merged.

PR #676, opened 2021-02-02
Update counter of UEFI configuration tables.

assafcarlsbad:
During runtime, UEFI code might install additional configuration tables via [`InstallConfigurationTable`](https://edk2-docs.gitbook.io/edk-ii-uefi-driver-writer-s-guide/5_uefi_services/readme.2/5210_installconfigurationtable) or via [`SmmInstallConfigurationTable`](https://github.com/tianocore/edk2/blob/master/MdePkg/Include/Pi/PiSmmCis.h). Until now, the counter associated with each configuration table was static and thus new entries would remain hidden from client code. This PR remedies this deficiency by simply incrementing the counter upon successful installation of a new configuration table entry.

@elicn 

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----


elicn:
Thanks, I was too lazy to implement that (:

Note that the conf install methods are used also for updating existing entries. Increasing the counter by one in these cases would result in a wrong number. Perhaps setting it to `len(ql.loader.efi_conf_table_array)` would be better.

I would prefer the update to be done on the install methods, where they belong.
Could you please elaborate on when conf tables might be installed before `gST` / `gSmst` are initialized..? (any specific test case?)
I'll try to look into it.

assafcarlsbad:
In `st.py` we call `install_configuration_table` twice before initializing the memory pointed to be `gST`:

```
install_configuration_table(ql, "HOB_LIST", None)
install_configuration_table(ql, "DXE_SERVICE_TABLE", gDS)

instance = EFI_SYSTEM_TABLE()
instance.RuntimeServices = gRT
instance.BootServices = gBS
instance.NumberOfTableEntries = len(confs)	# HOB_LIST and DXE_SERVICES
instance.ConfigurationTable = cfg

instance.saveTo(ql, gST)
```

elicn:
Yes, that's a good idea..

xwings:
> maybe we can reverse the order of operations (first initialize the `EFI_SYSTEM_TABLE`, then install the configuration tables). WDYT?

Do you want to do it in this PR or the next one ?

elicn (approved):

xwings:
Cool.

merged.

PR #677, opened 2021-02-04
move mips ccall

cq674350529:
move mips o32 call convention into os/fncc.py

> Currently only add o32 call convention, since it's most ubuquitious. There is another conventions, such as o64, n32/n64, maybe added later.
> MIPS Call Convention Reference: https://en.wikipedia.org/wiki/Calling_convention#MIPS


xwings:
thanks for the help!

PR #681, opened 2021-02-06
Added correct return values for stat, stat64, lstat and lstat64

PeterStolz:
Before stat, stat64, lstat and lstat64 could only return 0 or -1. Now they return the actual os error code. Since the functions only differ in the struct they create (stat or stat64) and the method they use for the info (Stat or Lstat) I moved that function that is now just invoked by the syscalls.

The fstat* syscalls still only have one error code. I don't understand how the code `ql.os.fd[fstat64_fd].fstat() == -1` in ql_syscall_fstat64. How can this .fstat() call ever return -1, since it is basically just a wrapper for the os.fstat function. Is there exception handling? If I know that I will create another PR for adding correct error codes to them as well.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/posix/syscall/stat.py
+++ b/qiling/os/posix/syscall/stat.py
@@ -230,6 +230,22 @@ def create_stat64_struct(ql, info):
     return fstat64_buf
 
 
+def statFamily(ql, path, ptr, name, stat_func, struct_func):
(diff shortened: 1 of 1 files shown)


xwings:
this is nice.

thanks!

PR #693, opened 2021-02-14
Add accessor to retrieve hooks for a specific address

learn-more:
I needed this to translate an address to a human readable name,
since f.e. uefi samples 'hook' addresses to use api's


xwings:
Merged. Thanks

PR #709, opened 2021-02-26
Re-enable multithread tests

bet4it:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/linux/thread.py
+++ b/qiling/os/linux/thread.py
@@ -23,17 +23,11 @@ from qiling.exception import *
 LINUX_THREAD_ID = 2000
 
 
-def new_thread_id():
-    global LINUX_THREAD_ID
-    old = LINUX_THREAD_ID
-    LINUX_THREAD_ID += 1
-    return old
-
 class QlLinuxThread(QlThread):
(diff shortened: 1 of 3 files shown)


qiling/os/linux/thread.py
```diff
@@ -360,9 +354,14 @@ def is_running(self):
...
-    def update_global_thread_id(self):
-        QlLinuxThread.LINUX_THREAD_ID = os.getpid()
+    def new_thread_id(self):
+        global LINUX_THREAD_ID
+        self._thread_id = LINUX_THREAD_ID
+        LINUX_THREAD_ID += 1
 
+    def update_global_thread_id(self):
+        global LINUX_THREAD_ID
+        LINUX_THREAD_ID = os.getpid()
```
xwings, line 364: hi, i think its best we keep the pid value to qiling space. getting the value from os might be a bad idea. (i know its our own code). maybe we should start from fresh.

bet4it: Yeah that's right, I also notice it, but it's not easy to fix it so let's leave it to another PR 😀

wtdcode:
Reviewed.

Nice work, thanks!

PR #725, opened 2021-03-17
Add support for MAP_FIXED + mmap system call refactoring

0xPhoeniX:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----


bet4it:
You should remove trailing whitespace

0xPhoeniX:
@bet4it should be fixed now.

wtdcode:
Thanks for your PR.

One question: Why would you like to refactor the mmap syscall?

0xPhoeniX:
@wtdcode because when I fixed the bug in one system call, I needed to write the same code twice in other mmap version.
It looks to me that having the same code in one function will make it

- much less effort to maintain code
- prevent bugs
- reduce code duplication (the very first thing that I was  taught during sw engineering courses)

Do you have other thoughts ?

xwings:
Cool,

i think we got similar call for macos. Can we merge it in to 1 ?

0xPhoeniX:
@xwings - is this the right place for mmap implementation -> https://github.com/qilingframework/qiling/blob/master/qiling/os/macos/syscall.py ?

As you know qiling arch better, any suggestions where to move the common implementation of `mmap` to use it across platforms ?

bet4it:
You may refer to this commit: ac73e1eb61b40d3b56283b9bf68e77e6429b9267

0xPhoeniX:
> Cool,
> 
> i think we got similar call for macos. Can we merge it in to 1 ?

Done.

0xPhoeniX:
@xwings Hey, can you please fix CI/CD pipeline to point to the right file : https://github.com/qilingframework/rootfs/raw/master/x8664_macos/System/Library/Kernels/kernel.development ?

That should fix the test.

wtdcode:
> @xwings Hey, can you please fix CI/CD pipeline to point to the right file : https://github.com/qilingframework/rootfs/raw/master/x8664_macos/System/Library/Kernels/kernel.development ?
> 
> That should fix the test.

That line is commented already. You could merge the latest dev branch to yours and the CI shall pass.

0xPhoeniX:
@xwings
```
open(/private/var/db/dyld/dyld_shared_cache_x86_64, O_RDONLY, 0o0) = -2
[+] QlThread None	File Not Found /Users/distiller/project/tests/../examples/rootfs/x8664_macos/private/var/db/dyld/dyld_shared_cache_x86_64
[+] QlThread None	open() = -0x2 (ENOENT)
```
Most probably fails the test. When I was testing locally with dyld (> 1Gb) present, the test succeeded.

xwings:
Hi,

Just remove the code if not needed. Then I will merge it.

Thanks!

0xPhoeniX:
@xwings  - what code?

xwings:
there are some unused code which is being comment out ?

0xPhoeniX:
@xwings  - done. 

Do note: I suggest to to leave the comment for mmap2 syscall so that others will be aware how the mmap2 for mac is implemented in qiling

xwings:
Actually, if the syscall is not using in macos. we should just remove it as well. 

the reason why there is a macos os mmap is due to the incompatibility and quick fix. 

since we got it all working now. we should just remove it.

0xPhoeniX:
@xwings  should be all fixed now.

xwings:
Cool, thanks

PR #727, opened 2021-03-17
IDA plugin now support arguments

zi0Black:
Added support for custom binary arguments, for IDA plugin inside qilingida.py.

Updated ChangeLog.


qiling/extensions/idaplugin/qilingida.py
```diff
@@ -2059,11 +2061,16 @@ def ql_set_rootfs(self):
 
         rootfspath = setupdlg.path_name.value
         customscript = setupdlg.script_name.value
+        parameter = setupdlg.parameters.value
         env = setupdlg.env_var.value
 
         if customscript != '':
             self.customscriptpath = customscript
 
+        para_array = parameter.split(' ')
```
wtdcode, line 2070: Need a more robust parser. e.g. parsing strings including spaces.

wtdcode (changes requested):

qiling/extensions/idaplugin/qilingida.py
zi0Black: Probably a regex or a proper grammar for ast could do the job, WDYT?

zi0Black: Or we can just demand everything from the user, and ask for a JSON as we did for the custom ENV.

wtdcode: Not sure. One case I'm considering is like:

`printf "Hello %s" "world!"`

wtdcode: A relevant post: https://stackoverflow.com/questions/1706551/parse-string-into-argv-argc

zi0Black: Just have a user-supplied JSON (from a file)?

e.x:
{"-c","aaaaa","bbbbb","\"c deefedede\""}

zi0Black: > A relevant post: https://stackoverflow.com/questions/1706551/parse-string-into-argv-argc

BTW, I will look at it, in the next few days.

xwings:
please update the conflict too

qiling/extensions/idaplugin/qilingida.py
zi0Black: I'm now using `shlex` in order to parse arguments @wtdcode.
Input: `-c aaaa -j "aa bb"`

parsed: `['-c', 'aaaa', '-j', 'aa bb']`

wtdcode:
@zi0Black Nice work, thanks!

PR #732, opened 2021-03-19
make filter functioning again

ucgJhe:
its now working as expected


wtdcode:
Why not something like `^(open|brk)`?

ucgJhe:
> Why not something like `^(open|brk)`?

just want do demo the comma thing like we did mention in the description , or we just remove it ?

wtdcode:
> > Why not something like `^(open|brk)`?
> 
> just want to demo the comma thing like we did mention in the description , or we should remove that feature ?

I'm not sure why we have to support comma considering the regex has included that functionality. I suggest removing that.

tests/test_qltool.py
```diff
@@ -43,6 +43,17 @@ def test_qltool_json(self):
         except subprocess.CalledProcessError as e:
             raise RuntimeError("command '{}' return with error (code {}): {}".format(e.cmd, e.returncode, e.output))
 
+    def test_qltool_filter(self):
+        create = [sys.executable, '../qltool', 'run', '-f', '../examples/rootfs/arm_linux/bin/arm_hello', '--rootfs', '../examples/rootfs/arm_linux', '-e', '^(open|brk)', '--log-file', 'filtered.qlog']
```
wtdcode, line 47: Why not use pipe here?

ucgJhe: silly me ~

wtdcode:
Merged, thanks.

PR #737, opened 2021-03-20
Refactored DOS

elicn:
Changelog highlights:
- Turned the monolithic implementation into a modular one by extracting the interrupt handlers out to dedicated modules
- Broke interrupt handlers down to leaf functions for better readability
- Implemented DOS header parsing for EXE files (as opposed to naked COM files)
- Cleaned and tidied code up
- Refreshed DOS tests

### DIFF
--- a/tests/test_dos.py
+++ b/tests/test_dos.py
@@ -7,12 +7,32 @@ import sys, unittest
 sys.path.append('..')
 
 from qiling import Qiling
+from qiling.const import QL_INTERCEPT
+
+class Checklist:
+    def __init__(self) -> None:
+        self.visited_onenter = False
+        self.visited_onexit = False
 
 class DOSTest(unittest.TestCase):
 
     def test_dos_8086_hello(self):
-        ql = Qiling(["../examples/rootfs/8086/dos/HI.DOS_COM"], "../examples/rootfs/8086/dos")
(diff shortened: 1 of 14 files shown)


tests/test_dos.py
```diff
@@ -7,12 +7,32 @@
...
+class Checklist:
+    def __init__(self) -> None:
+        self.visited_onenter = False
+        self.visited_onexit = False
 
 class DOSTest(unittest.TestCase):
 
     def test_dos_8086_hello(self):
-        ql = Qiling(["../examples/rootfs/8086/dos/HI.DOS_COM"], "../examples/rootfs/8086/dos")
+        ql = Qiling(["../examples/rootfs/8086/dos/HI.DOS_COM"], "../examples/rootfs/8086/dos", output='debug')
```
xwings, line 20: i am thiniking we should remove output="" for good and keep only verbose ?

elicn: I always found `output` and `verbose` a bit confusing, but in the current impl both are needed.
Personally, I use `output='debug'` only when I need to enable assembly-level tracing*. We should think of a way to consolidate them both into one variable that would reflect all needed verbosity levels and allow tracing.

( * see what I did for `hello_x8664_linux_disasm.py` )

xwings: i agree. I will merge this and we need to do it next.

wtdcode:
For EXE files, PSP also needs to be implemented.

PR #755, opened 2021-04-01
Qdb and syscall_writev improvements

ucgJhe:
1. use f-string in qdb
2. fix qdb crashed when user try to continue a not-yet started qiling instance
3. make syscall writev actually write to fd


xwings:
Thanks !

PR #756, opened 2021-04-06
Additional features support  to UEFI

elicn:
Changelog highlights:
- More typing annotations and comments
- Code cleanup
- Bugfix in notify handler
- Implemented SW SMI handlers Register and UnRegister, based on @liba2k implementation

@liba2k let me know you are OK with the changes to event handling and SW SMI registration.

### DIFF
--- a/qiling/core_hooks.py
+++ b/qiling/core_hooks.py
@@ -127,12 +127,20 @@ class QlCoreHooks:
         ql, hook_type = args[-1]
 
         if hook_type in self._insn_hook.keys():
+            retval = None
+
             for h in self._insn_hook[hook_type]:
                 if h.bound_check(ql.reg.arch_pc):
                     ret = h.call(ql, *args[ : -1])
-                    if isinstance(ret, int) == True and ret & QL_HOOK_BLOCK  != 0:
+
+                    if type(ret) is tuple:
+                        ret, retval = ret
+
+                    if type(ret) is int and ret & QL_HOOK_BLOCK:
                         break
 
+            # use the last return value received
+            return retval
 
     def _callback_type4(self, uc, addr, size, pack_data):
         ql, user_data, callback = pack_data
--- a/qiling/loader/pe_uefi.py
+++ b/qiling/loader/pe_uefi.py
@@ -50,7 +50,7 @@ class QlLoaderPE_UEFI(QlLoader):
         for member in QlLoaderPE_UEFI.__save_members:
             saved_state[member] = getattr(self, member)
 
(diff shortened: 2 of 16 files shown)


liba2k:
Hello @elicn I'm on vacation and can't run the code to check that it doesn't break anything. I did go over the changes and the y look good to me.
I did add support for another type of SMI (SMMC) in our efi_fuzz project, maybe we should implement the (or some of them here). Please see https://github.com/Sentinel-One/efi_fuzz/compare/smmc.
Thank you!

xwings:
@liba2k  any decision :)

liba2k:
@elicn What do you thing about supporting smmc?

elicn:
@liba2k 
Sure. As far as I remember this is just a way of conveying data into SMM, instead of pulling a pointer out of a CPU register stored in SSA. It is still SW SMI, so it is registered / unregistered / invoked exactly in the same way -- except that it needs different preparation.

To my understanding, the code already handles SMMC registration and unregistration (because it's yet another SW SMI) - it just doesn't do the preparation needed for the invocation. IMHO, the invocation itself belongs to tools that use Qiling rather than Qiling itself, and so is the preparation beforehand.

Suppose you implemented a fuzzer, you'd expect Qiling to handle the registration part - but the fuzzer itself would be responsible to prepare the data before it fires up the SW SMI.

If you think the current code needs to be modified to better support SMMC, let me know what you have in mind.

liba2k:
@elicn The difference as far as qiling is concerned is that SMMCs are registers using gSmst->SmiHandlerRegister and the other SWSMIs using the EFI_SMM_SW_DISPATCH_PROTOCOL and EFI_SMM_SW_DISPATCH2_PROTOCOL.

elicn:
@liba2k That can be implemented, but it looks orthogonal to the current one.
Anything on this PR you think needs to change?

liba2k:
@elicn The only reason I was referring you to the efi_fuzz change is that I've consolidated all three register/unregister to a single function, but we can leave that for a separate PR or just leave is external to qiling.

elicn:
@liba2k I think it would be better to have the registration methods separated. We could either keep them in separated lists (that is, a list for regular SW SMI, and another for SMMC), or on the same one with some kind of a "type" differentiator. Either way, that list(s) would be picked up by the utility / fuzer once the module(s) execution is done.

What do you think?

liba2k:
@elicn Sounds good to me.
@xwings I have no more comment.
Thank you both!

xwings:
cool cool

PR #757, opened 2021-04-06
Fix gdb arm PC and CPSR

JeffmeisterJ:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

I was running into some problems using the Qiling GDB server for ARM shellcode, mostly related to the ARM CPSR register not being read properly. This in turn causes GDB single stepping to fail in Thumb mode, skipping instructions and leading to a runaway condition of the emulator. This is now fixed and tested with the Qiling GDB server against Ghidra and ret-sync, which seems to be very stable now.


qiling/debugger/gdb/gdb.py
```diff
@@ -364,7 +363,7 @@ def handle_p(subcmd):
                             reg_value = self.addr_to_str(reg_value, short = True)
                     
                     elif self.ql.archtype== QL_ARCH.ARM:
-                        if reg_index < 17:
+                        if reg_index < 26:
```
xwings, line 366: 26 ? it never work for arm ?

JeffmeisterJ: Thanks for your fast response!

Only in the **handle_g** call does it support a max of 16 core registers. In the **handle_p** and **handle_P** methods it will request registers by offset, where 25 is the CPSR. The registers in between are the old FP registers, which have become obsolete. 

As you can see with the debug logging, this is visible if we write the CPSR register from GDB:
image

Which shows up in the logging as:

image

In the GDB serial protocol, the 19 is the hexadecimal representation, which is decimal 25.

xwings:
Hi,

can you update with the latest code?

JeffmeisterJ:
Merged latest dev branch, no conflicts.

xwings:
Cool. Merge now.

PR #765, opened 2021-04-09
Enhanced tracing POC

elicn:
A proof of concept for an enhanced tracing feature.

The enhanced tracing comes to replace the basic disassembly available today with an analyzed diassembly, featuring live CPU state effective address calculation and symbols resolving. The analyzed disassembly greatly simplifies the ability to track execution and analyze crashes.

Main features:
- Better formatted output, shown in a darker color to be easily distinguished from other log entries
- Values of registers referenced (read) by the instruction are shown alongside it
- Indirect memory references are replaced with the calculated effective address (e.g. `lea rax, [rbx + rsi*8]` where `rbx=0x2000` and `rsi=0x10`, will result in `lea rax, [0x2080]`)
- If a symbols mapping is set in `ql.loader.symsmap` before enabling the trace, memory addresses will be searched to see if they have a matching symbol

The following screenshot shows enhanced tracing during a UEFI module execution, with symbols mapping:

index

The enhcanced tracing may operate in two modes: `full` and `history`:
- `full` mode: a trace line will be emitted for each assembly instrcution before it gets executed
- `history` mode: trace lines will not be emitted during execution till the program crashes. In that case, the last `n` trace lines will be emitted, where `n` is configurable. This mode comes to reduce the tracing overhead and allow enough information to analyze a crash 

Notes:
- To show tracing info, the `Qiling` instance should be set to `verbose=QL_VERBOSE.DEBUG` or higher
- Only Intel architecture is currently supported by the POC

Simple example:
```python
from qiling import Qiling
from qiling.extensions import trace

ql = Qiling(myargv, myrootfs, myenv, verbose=QL_VERBOSE.DEBUG)

# load symbols mapping here, before enabling trace
ql.loader.symsmap = {
    0x00000000 : 'NULL'
}

# choose trace mode
trace_mode = 'full'

# enable full tracing throughout execution
if trace_mode == 'full':
    trace.enable_full_trace(ql)

# enable tracing in history mode, showing last 32 lines on crash
elif trace_mode == 'history':
    trace.enable_history_trace(ql, nrecords=32)
```

### DIFF
--- a/qiling/extensions/trace.py
+++ b/qiling/extensions/trace.py
@@ -1,0 +1,198 @@
+#!/usr/bin/python3
+
+from collections import UserList
+from typing import Iterable, Iterator, Mapping, Tuple
+
+from capstone import Cs, CsInsn, CS_OP_IMM, CS_OP_MEM, CS_OP_REG
+from capstone.x86 import X86Op
+from capstone.x86_const import X86_INS_LEA
+
+from unicorn.x86_const import UC_X86_REG_INVALID, UC_X86_REG_RIP
+
+from qiling import Qiling
+
+TraceRecord = Tuple[CsInsn, Iterable[Tuple[int, int]]]
+
+def __get_trace_records(ql: Qiling, address: int, size: int, md: Cs) -> Iterator[TraceRecord]:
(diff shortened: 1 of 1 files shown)


xwings:
wow, thanks !

Issue #767, opened 2021-04-10
Cannot start gdbserver because self._debugger.run() is after self.os.run() in core.py


itewqq:
***Describe the bug**
I am trying to use qiling to simulate a piece of code in the upnpd of netgear xr300, but I found that enabling ql.debugger did not start any gdbserver. After debugging the source code, I found that your logic in the run function in core.py is

```
def run(self, begin=None, end=None, timeout=0, count=0):
        # replace the original entry point, exit point, timeout and count
        self.entry_point = begin
        self.exit_point = end
        self.timeout = timeout
        self.count = count
        self.write_exit_trap()

        # init debugger
        if self._debugger != False and self._debugger != None:
            self._debugger = debugger_setup(self._debugger, self)

        # patch binary
        self.__enable_bin_patch()

        # emulate the binary
        self.os.run()

        # run debugger
        if self._debugger != False and self._debugger != None:
            self._debugger.run()
```

I found that when I run my emulator, I get stuck at the self.os.run() step, so I cannot execute the following _debugger.run(). When I moved the debugger.run() line before os.run, it can run normally, and my ida can also connect to this remote debugger:
![image](https://user-images.githubusercontent.com/30570177/114258852-1b137380-99fc-11eb-9217-89a3cf833b22.png)

So I am a little curious why this is happening, what is the relationship between os.run and debugger? Is this set deliberately or is it a bug?

**Sample Code**
My emu.py
```python
#!/usr/bin/env python3
# coding=utf-8
import sys
import os
from qiling import *
from qiling.const import QL_VERBOSE
from pwn import ELF, u32, p32, log
import random

# pre-set nvram key-value
nvram_profile = {"upnp_turn_on": ["1"], "lan_ipaddr": ["127.0.0.1"], "upnp_duration": ["60"], "upnp_advert_period": ["5"]}

# important address
MAIN_ADDR = 0x22d10
VULN_FUNC_START_ADDR = 0x214B8
STRNCPY_ADDR = 0x21514
VULN_FUNC_RET_ADDR = 0x21554
SIM_START = 0x199B4
SIM_END = 0x1A12C


def print_stack(_ql, cnt):
    cont = _ql.mem.read(_ql.reg.sp, cnt*4)
    for i in range(cnt):
        print("%d(+0x%x,0x%x)" % (i, i*4, _ql.reg.sp+i*4),
              ": 0x%x" % u32(cont[i*4:i*4+4]))


'''
def log(msg):
[...]

Issue #593, opened 2020-11-27
partial emulation enhance


cq674350529:
Currently, it's often to run partial emulation with saved context. However in many cases, there is no way to get and save the execution context.

If running partial emulation independently, it only supports pure codes without imported function calls, for these `lib.so` haven't been mapped.

As can be seen, if `self.ql.entry_point` is set (`begin` param in `ql.run()`), the emulation starts from `self.ql.entry_point` directly. Since the loader (`ld.so`) is not executed, the related `lib.so` won't be loaded and mapped.
https://github.com/qilingframework/qiling/blob/3ed13762d52027d9b469d3fa9a4a57e68c3bd7f4/qiling/os/linux/linux.py#L123-L131

A simple example is shown as follows.

```c
int main(int argc, char* argv[])
{
    char* buf[10] = {0};
    memcpy(buf, "hello", strlen("hello"));
    printf("%s\n", buf);
    return 0;
}
```

```assembly
.text:004007D0  # int __cdecl main(int argc, const char **argv, const char **envp)
.text:004007D0 addiu   $sp, -0x48
.text:004007D4 sw      $ra, 0x48+var_4($sp)
.text:004007D8 sw      $fp, 0x48+var_8($sp)
.text:004007DC move    $fp, $sp
.text:004007E0 li      $gp, 0x4189A0
.text:004007E8 sw      $gp, 0x48+var_38($sp)
.text:004007EC sw      $a0, 0x48+arg_0($fp)
.text:004007F0 sw      $a1, 0x48+arg_4($fp)
.text:004007F4 sw      $zero, 0x48+var_30($fp)
.text:004007F8 sw      $zero, 0x48+var_2C($fp)
.text:004007FC sw      $zero, 0x48+var_28($fp)
.text:00400800 sw      $zero, 0x48+var_24($fp)
.text:00400804 sw      $zero, 0x48+var_20($fp)
.text:00400808 sw      $zero, 0x48+var_1C($fp)
.text:0040080C sw      $zero, 0x48+var_18($fp)
.text:00400810 sw      $zero, 0x48+var_14($fp)
.text:00400814 sw      $zero, 0x48+var_10($fp)
.text:00400818 sw      $zero, 0x48+var_C($fp)
.text:0040081C li      $a2, 5           # n
.text:00400820 lui     $v0, 0x40
.text:00400824 addiu   $a1, $v0, (aHello - 0x400000)  # "hello"
[...]

PR #768, opened 2021-04-10
Make gdb work with snapshot in Partial Execution.

itewqq:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/debugger/gdb/gdb.py
+++ b/qiling/debugger/gdb/gdb.py
@@ -65,6 +65,11 @@ class QlGdb(QlDebugger, object):
         else:
             self.entry_point = self.ql.os.entry_point
 
+        # Only part of the binary file will be debugged.
+        if self.ql.entry_point is not None and self.ql.exit_point is not None:
+            self.entry_point = self.ql.entry_point
+            exit_point = self.ql.exit_point
+
         self.gdb.initialize(self.ql, self.entry_point, exit_point=exit_point, mappings=[(hex(load_address))])
 
         #Setup register tables, order of tables is important
--- a/examples/hello_x8664_linux_part_debug.py
+++ b/examples/hello_x8664_linux_part_debug.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# 
+# Cross Platform and Multi Architecture Advanced Binary Emulation Framework
+#
+
+import sys
+sys.path.append("..")
+
+from qiling import Qiling
+from qiling.const import QL_VERBOSE
+
+def dump(ql, *args, **kw):
+    ql.save(reg=False, cpu_context=True, snapshot="/tmp/snapshot.bin")
+    ql.emu_stop()
+
+if __name__ == "__main__":
+    ql = Qiling(["rootfs/x8664_linux/bin/sleep_hello"], "rootfs/x8664_linux", verbose=QL_VERBOSE.DEFAULT)
+    # load base address from profile file
+    X64BASE = int(ql.profile.get("OS64", "load_address"), 16)
+    # take a snapshot
+    ql.hook_address(dump, X64BASE + 0x1094)
+    ql.run()
+
+    ql = Qiling(["rootfs/x8664_linux/bin/sleep_hello"], "rootfs/x8664_linux", verbose=QL_VERBOSE.DEBUG)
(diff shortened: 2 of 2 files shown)


qiling/debugger/gdb/gdb.py
```diff
@@ -65,6 +65,11 @@ def __init__(self, ql, ip, port):
         else:
             self.entry_point = self.ql.os.entry_point
 
+        # Only part of the binary file is debugged.
+        if self.ql.entry_point is not None and self.ql.exit_point is not None:
+            self.entry_point=self.ql.entry_point
```
aquynh, line 70: please put a space before and after `=`  (so " = ", not "=").

itewqq: Oh, my bad. Fixed now.

xwings:
Hi,

We still need a elf header?

itewqq:
@xwings Ok, sorry I didn't explain my context clearly. 

In fact, if you take a look at my python script, you will find that I am actually performing the so-called "Partial Execution" simulation, as mentioned in the [document snapshot](https://docs.qiling.io/en/latest/snapshot/). 

I first let the program run, hook at the main address and take a snapshot. Then I resume the snapshot, but directly assign the pc pointer to the beginning of the function I want to simulate.

In this way, I can simply construct the call_state (registers, memory mapping, dynamic library loading, etc) of the function, and at the same time avoid the tedious operation (including various hardware conditions such as NVRAM, and other processes Interaction) from the entry point of the elf to the part I am really interested in. I think this is a very important reason why I use qiling instead of directly using qemu and other large-scale simulations. After finished the static analysis, I only need to simulate the functions of interest, and then write exploit scripts...save me a coffee time.

But when I did this, I found that qiling could not start gdbserver (actually I could only use hook and print to debug), so I dig into the source code of qiling and found that it only takes a little work to make this become possible.

aquynh:
Interesting! Will be nice if you can contribute a simple sample to examples/ so others can learn from this case.

itewqq:
A simple example ```examples/hello_x8664_linux_part_debug.py``` is added based on the existing sleep_hello program.

cq674350529:
@itewqq Just to complement: it's possible to emulate part of the code in binary and enable gdb without snapshot, please see https://github.com/qilingframework/qiling/issues/593.

Indeed, if running with snapshot, the gdb won't work as you describe.

itewqq:
> @itewqq Just to complement: it's possible to emulate part of the code in binary and enable gdb without snapshot, please see #593.
> 
> Indeed, if running with snapshot, the gdb won't work as you describe.

@cq674350529 Well, my snapshot was taken at the beginning of ```main```... So I did not encounter the problem in this issue？

Can you explain how to emulate part of the code in binary and enable gdb without snapshot? And why if running with snapshot, the gdb won't work as me describe, have you tried https://github.com/qilingframework/qiling/pull/768/commits/d3d696fcaa13c6ad10d67eb7b54aa468d92bf863 ?

cq674350529:
@itewqq Sorry for getting you misunderstand. There is nothing wrong with your pull request, and it can solve the issue https://github.com/qilingframework/qiling/issues/767 you encoutered.

What I mean is that without your fix, there may be some issues when running gdb with snapshot.

> Can you explain how to emulate part of the code in binary and enable gdb without snapshot?

As mentioned in https://github.com/qilingframework/qiling/issues/593, what we need is to hook at the beginning of `main`, and hijack `PC` to address you want ro run start.

```python
def sample():
    def partial_run(ql, *args, **kwargs):
        ql.reg.arch_pc = <address_to_run>    # hijack pc

    ql = Qiling([bin_path], rootfs_path, console=True, verbose=QL_VERBOSE.DEBUG)
    ql.hook_address(partial_run, <begin_of_main>)   
    ql.debugger = "127.0.0.1:1234"

    ql.run()
```
 I usually do partial code emulation in this way.

itewqq:
Wow, amazing! I didn’t expect to do this before, because my main reference is Partial Execution from https://docs.qiling.io/en/latest/snapshot/ 😆 Thank you very much!!!

cq674350529:
@itewqq Yeah, using snapshot mechanism can achieve the same goal. But I prefer this way, for it's more simple and clear.

xwings:
Cool. Nice work buddy.

Thanks.

PR #779, opened 2021-04-23
qltool refresh

elicn:
Refactored and cleaned up `qltool` to make it simpler to use (and maintain).

Highlights:
- The mutually exclusive flags `asm`, `hex` and `bin` are now consolidated into one option: `--format {asm, hex, bin}`
- The mutually exclusive flags `dump`, `debug`, `disasm` and `trace` are now consolidated into `--verbose` that now accepts a logging level name (instead of a meaningless integer value before): `--verbose {off, default, debug, diasm, dump}`
- Changed `--console False` to `--no-console` to make it more intuitive
- Changed `--bigendian` to `--endian {little, big}` to make it more intuitive (default: `little`)
- Removed manual arguments validation (e.g. allowing choosing only from specific choices in `arch` and `os`), and set `argparse` to handle it
- Removed manual hack for handling `--version`, and set it as a formal flag
- Usage message including all verbosed examples was changed to `examples` and set as a subparser. It will no longer be shown on every `qltool` command line syntax error, but with `qltool examples`
- Enhanced help / usage
- Refreshed the examples message (e.g. removed obsolete / nonexisting flags)
- Code cleaned and tidied up
- Few minor bugfixes
- Test file updated to meet the new options / flags

### DIFF
--- a/qiling/arch/utils.py
+++ b/qiling/arch/utils.py
@@ -7,100 +7,87 @@
 This module is intended for general purpose functions that are only used in qiling.arch
 """
 
-from unicorn import UcError, UC_ERR_READ_UNMAPPED, UC_ERR_FETCH_UNMAPPED
-from keystone import *
-from capstone import *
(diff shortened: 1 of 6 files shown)


xwings:
would you mind to update docs too :)

xwings:
as usual. nice work!

elicn:
Created a PR for the doc

Issue #780, opened 2021-04-24
can't run example scripts crackme_x86_linux.py 


23R3F:
~/qiling/examples$ python3 crackme_x86_linux.py 
Traceback (most recent call last):
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/os/posix/posix.py", line 282, in load_syscall
    ret = syscall_hook(self.ql, *arg_values)
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/os/posix/syscall/stat.py", line 884, in ql_syscall_fstat64
    fstat64_buf = pack_stat64_struct(ql, fstat64_info)
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/os/posix/syscall/stat.py", line 811, in pack_stat64_struct
    stat64.__setattr__(field, int(info[field]))
TypeError: tuple indices must be integers or slices, not str
Traceback (most recent call last):
  File "crackme_x86_linux.py", line 96, in <module>
    solve()
  File "crackme_x86_linux.py", line 76, in solve
    prev_ic = run_one_round(bytes(flag))
  File "crackme_x86_linux.py", line 65, in run_one_round
    ql.run()
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/core.py", line 724, in run
    self.os.run()
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/os/linux/linux.py", line 132, in run
    self.ql.emu_start(self.ql.loader.elf_entry, self.exit_point, self.ql.timeout, self.ql.count)
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/core.py", line 868, in emu_start
    raise self._internal_exception
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/utils.py", line 157, in wrapper
    return func(*args, **kw)
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/core_hooks.py", line 118, in _hook_intr_cb
    ret = h.call(ql, intno)
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/core_hooks.py", line 34, in call
    return self.callback(ql, *args)
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/os/linux/linux.py", line 87, in hook_syscall
    return self.load_syscall()
  File "/home/aaa/.local/lib/python3.6/site-packages/qiling/os/posix/posix.py", line 298, in load_syscall
    raise e
[...]

PR #781, opened 2021-04-24
fix the bug of crackme_x86_linux.py in qiling/example/

23R3F:
details in follow:
https://github.com/qilingframework/qiling/issues/780

### DIFF
--- a/examples/crackme_x86_linux.py
+++ b/examples/crackme_x86_linux.py
@@ -2,7 +2,7 @@
 # 
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
-
+from qiling.os.posix.stat import Fstat
 import sys
 sys.path.append("..")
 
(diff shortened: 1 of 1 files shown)


elicn:
Can you please explain what was the bug and how you fixed it?
It appears that you delted a whole bunch of good code, comments and annotations in the process.

xwings:
i agree with @elicn 

the code in example should be more longer and detail compare to the test.

the concept is the same, but code example should be more detailed.

23R3F:
@xwings @elicn i know that ,sorry to reply too late, i will write more detail when i have time （sometimes 996 working orz

elicn:
I can't edit the content, so here's what I think should be kept from the changes:
- Replacing `import os` with `from qiling.os.posix.stat import Fstat`, but place the line under the Qiling path addition
- Replacing `return os.fstat(sys.stdin.fileno())` with `return Fstat(sys.stdin.fileno())`
That's it.

23R3F:
@elicn yeah，you are right，i have change the commit

elicn:
> @elicn yeah，you are right，i have change the commit

It looks like the new commit still replaces or deletes good commented code for no visible reason..

PR #785, opened 2021-04-26
Enhancing QlOsUtils print_function

elicn:
As of today function invocations details are handled by `QlOsUtil.print_function`, which determines how to display them.
This enhancement extracts the arguments values processing to the OS, letting each OS derivative override the way function call arguments are displayed by overriding the OS `process_fcall_params` method. In other words, each OS may decide how to display certain arguments types: e.g. use a certain text formatting or colors.

This PR includes a basic POC for the UEFI OS, which lets it display arguments values of certain types in a more convinient way:
- Wide strings are prefixed with `L`
- Known GUIDs are presented by their names (unknown ones retain the original GUID string)
- All other arguments types remain the same

For example, an excerpt from `examples/rootfs/x8664_efi/bin/TcgPlatformSetupPolicy` execution log:

image

All OS derivates benefit from this enhancement, however only UEFI currently makes use of it.
There is no visible change to other OS's output.


xwings:
Nice work buddy!

PR #790, opened 2021-04-28
Misc. UEFI improvements and bugfixes

elicn:
Highlights:
- Fix `GetCapabilities` method to return the right size
- Fix notifcations handling
- More HOB handling methods
- Styling and typing annotations


xwings:
Cool,

thanks!

PR #801, opened 2021-05-24
Fix bug: DLL address still use ImageBase address in spite of changing lib_base.

nacayoshi00:
…  lib_base

- [ x ] This PR only contains minor fixes.

### Coding convention?

- [ x ] The new code conforms to Qiling Framework naming convention.

### Extra tests?

- [ x ] No extra tests are needed for this PR.

### Changelog?

- [ x ] Changelog will be updated after some proper review.

### Target branch?

- [ x ] The target branch is dev branch.

### One last thing

- [ x ] I have read the [contribution guide](https://docs.qiling.io/en/latest/contribution/)

-----


xwings:
There are some test with fixed based DLL address. Will you be able to update the test ?

nacayoshi00:
I understood. I will add the test to test_pe.py. Please be patient.

nacayoshi00:
I updated test_pe.py to check the fix of lib address relocation. 
And I also update 2 fixes as below:
 - Implement PE loader to load all of the DLLs recursively when PE loaded.
 - Implement PE loader to be able to load API DLL (like api-ms-xxxx.dll).

I added a sample for the test. So Should I send a pull request to https://github.com/qilingframework/rootfs/tree/master?

xwings:
> I updated test_pe.py to check the fix of lib address relocation.
> And I also update 2 fixes as below:
> 
> * Implement PE loader to load all of the DLLs recursively when PE loaded.
> * Implement PE loader to be able to load API DLL (like api-ms-xxxx.dll).
> 
> I added a sample PE file for the test. So Should I send a pull request to https://github.com/qilingframework/rootfs/tree/master?

Yes. If there is a new exe, you need to update rootfs

nacayoshi00:
OK, I sent the pull request of rootfs. And fix some code to pass the test

examples/scripts/dllscollector.bat
```diff
@@ -55,15 +55,7 @@ if exist %WINDIR%\SysWOW64\winhttp.dll xcopy /f /y %WINDIR%\SysWOW64\winhttp.dll
...
-if exist %WINDIR%\SysWOW64\downlevel\api-ms-win-crt-heap-l1-1-0.dll xcopy /f /y %WINDIR%\SysWOW64\downlevel\api-ms-win-crt-heap-l1-1-0.dll "examples\rootfs\x86_windows\Windows\System32\"
-if exist %WINDIR%\SysWOW64\downlevel\api-ms-win-core-synch-l1-2-0.dll xcopy /f /y %WINDIR%\SysWOW64\downlevel\api-ms-win-core-synch-l1-2-0.dll "examples\rootfs\x86_windows\Windows\System32\"
-if exist %WINDIR%\SysWOW64\downlevel\api-ms-win-core-fibers-l1-1-1.dll xcopy /f /y %WINDIR%\SysWOW64\downlevel\api-ms-win-core-fibers-l1-1-1.dll "examples\rootfs\x86_windows\Windows\System32\"
-if exist %WINDIR%\SysWOW64\downlevel\api-ms-win-core-localization-l1-2-1.dll xcopy /f /y %WINDIR%\SysWOW64\downlevel\api-ms-win-core-localization-l1-2-1.dll "examples\rootfs\x86_windows\Windows\System32\"
-if exist %WINDIR%\SysWOW64\downlevel\api-ms-win-core-sysinfo-l1-2-1.dll xcopy /f /y %WINDIR%\SysWOW64\downlevel\api-ms-win-core-sysinfo-l1-2-1.dll "examples\rootfs\x86_windows\Windows\System32\"
+xcopy /f /y %WINDIR%\SysWOW64\downlevel\api-*.dll "examples\rootfs\x86_windows\Windows\System32\"
```
xwings, line 58: I think we need to limit and copy only needed DLL. It seems the test took 4 hours due to additional dlls

nacayoshi00:
OK, I updated examples/scripts/dllscollector.bat to copy only needed DLL.

xwings:
Redoing the test and see how long will it take to do the test.

qiling/loader/pe.py
```diff
@@ -575,28 +580,98 @@ def load(self):
...
-                            try:
-                                addr = self.import_address_table[dll_name][imp.name]
-                            except KeyError:
-                                self.ql.log.debug("Error in loading function %s" % imp.name.decode())
-                        else:
-                            addr = self.import_address_table[dll_name][imp.ordinal]
-
-                        if self.ql.archtype == QL_ARCH.X86:
-                            address = self.ql.pack32(addr)
+            # load all DLL files linked from the target PE file recursively and make IAT of all dlls.
```
xwings, line 585: Here onwards

The test before this take around 15-20 min. With the new implementation. It takes around 4 hours to complete the test.

We need to find a better way, This is way too slow.

One method is to cache like how we cache the dll address. maybe we can do the same

qiling/loader/pe.py
```diff
@@ -96,6 +96,11 @@ def load_dll(self, dll_name, driver=False):
                 self.ql.log.warning(f'Warnings while loading {path}:')
                 for warning in warnings:
                     self.ql.log.warning(f' - {warning}')
+
+            # [Room for Improvement] too much time when kernelbase.dll is loaded. 
+            self.ql.log.debug('relocate {}, {:x}'.format(dll_name, self.dll_last_address))
```
xwings, line 101: We need to standardize all the quote. Double quote. " and not '

nacayoshi00:
I see, but the result was unexpected for me because the test took only 1638s on my laptop PC. 
I think the reason why the test is so long is below:
 - All DLLs are loaded at first even if some DLLs may not be used. 
 - DLL reload_image() may take quite long
 
Since I think these fixes are needed because emulation fails when DLL's address is not resolved, I have another solution to solve the problem.
1. Make resolve_dll_address function that resolves address when emulator encounters unmapped address.
2. Register the function to hook_address.

If u r OK, I'll try to update.

xwings:
1638s  is one test or all test ?

The test is being done from github CI, so i guess we need to make that as benchmark.

I guess your idea is fine. We need to fix CI result. Else gonna be bad if we need to spend 4 hours for each test. Since currently only 10min for all test.

nacayoshi00:
It took 1638s  to finish all the tests.
OK, I'll try to fix it. Please be patient.

xwings:
Sure. Will be waiting :)

nacayoshi00:
I'm sorry for late the response. I update these files to load dll dynamically. 
But the problem still remained because the reason why take quite so long is to relocate some dlls such as shell32.dll (see below log.). So I add libcache=True in all_tests of test_pe.py.

`>>> timeit.timeit("import pefile;pefile.PE('qiling/examples/rootfs/x86_windows/Windows/System32/shell32.dll').relocate_image(0x7fff00000)", number=1)
205.60229980000076`

I checked the time dramatically reduced in my PC when libcache is enabled. 
 - previous ver:1638s
 - new version(libcache off): 1000s
 - new version(libcache on): 80s)

nacayoshi00:
I resolved the conflict and re-uploaded

nacayoshi00:
Could you mind if I ask you to run the workflow? Or should I do something to proceed with the workflow?

xwings:
There is still a conlifct with wscok32

nacayoshi00:
Sorry too late. I removed the conflict. So could you please run the workflow?

elicn:
Hi @nacayoshi00,
Based on the recent changes made to `dev` branch, could you please test to see whether this PR is still relevant?

xwings:
Since author did not repnse since Aug 2021. PR will be close for now.

PR #803, opened 2021-05-25
Fix Linux fuzzing example

elicn:
AFL could not find any crash in the sample executable due to an incorrect pipe `fstat` implementation.
The fuzzing example was fixed, cleaned and reduced to the minimum necessary to keep it simple.


xwings:
Cool. Thanks!

PR #811, opened 2021-06-02
syscall/unistd.py: append setresuid && setresgid

rootkiter:
Warning Message:

[!] [Thread 1855]       0x774cb514: syscall ql_syscall_setresuid number = 0x1059(4185) not implemented

Reference:

https://man7.org/linux/man-pages/man2/setresuid.2.html


xwings:
y0, thanks!

Issue #816, opened 2021-06-08
Invalid instruction (UC_ERR_INSN_INVALID)


leepeter2019:
***Describe the bug**
As soon as the emulation begins, it stop with error message Invalid instruction (UC_ERR_INSN_INVALID)
when i checked the log, it says '[!]	api __stdio_common_vsprintf is not implemented',  '[!]	api __stdio_common_vsprintf_s is not implemented'. 
I think UC_ERR_INSN_INVALID error message is generated because of __stdio_common_vsprintf , __stdio_common_vsprintf_s.
But I added VS dynamic library 'msvcrt.dll' in the dll directory.
I don't know why it cannot find correspond api in the dll.

**Sample Code**
```python
ql = Qiling(["test.exe", "model", "test", log.txt"], "x8664_windows")
ql.run()
```

**Expected behavior**
It generates code coverage on target sw

**Screenshots**
```python
C:\Users\Steve\AppData\Local\Programs\Python\Python37\python.exe A:/Project/qiling-dev/test.py
[=]	Initiate stack address at 0x7ffffffde000
[=]	Loading helloworld.exe to 0x140000000
[=]	PE entry point at 0x140004788
[=]	TEB addr is 0x6000030
[=]	PEB addr is 0x60000b8
[=]	Loading examples/rootfs/x8664_windows\Windows\System32\ntdll.dll to 0x7ffff0000000
[!]	Warnings while loading examples/rootfs/x8664_windows\Windows\System32\ntdll.dll:
[!]	 - SizeOfHeaders is smaller than AddressOfEntryPoint: this file cannot run under Windows 8.
[!]	 - AddressOfEntryPoint lies outside the sections' boundaries. AddressOfEntryPoint: 0x0
[=]	Done with loading examples/rootfs/x8664_windows\Windows\System32\ntdll.dll
[=]	Loading examples/rootfs/x8664_windows\Windows\System32\kernel32.dll to 0x7ffff01f5000
[=]	Done with loading examples/rootfs/x8664_windows\Windows\System32\kernel32.dll
[=]	Loading examples/rootfs/x8664_windows\Windows\System32\user32.dll to 0x7ffff02b2000
[=]	Done with loading examples/rootfs/x8664_windows\Windows\System32\user32.dll
[=]	Loading examples/rootfs/x8664_windows\Windows\System32\shlwapi.dll to 0x7ffff0452000
[...]

PR #818, opened 2021-06-09
Add missing __stdio_common functions

elicn:
Changelog:
- Added a few more __stdio_common functions, per #816 
- Modified parameters names to follow MS stdio
- Update test with new parameters names


xwings:
Awesome. Thanks!

PR #820, opened 2021-06-11
Examples cpp

elicn:
Change Log:
- Added a "Hello World" C++ example
- Integrated the new example into Makefile, adding  missing C++ toolchain
- Facy up Makefile
  - Now it is possible to build targets by arch
  - Some make-fu
- Removed unnecessary includes from `patch_bin`


xwings:
thank you sir!

PR #823, opened 2021-06-17
core_hooks.py: Iterate on the copy of the hook list.

rootkiter:
When you change the hook list in the callback function, it will affect the iteration effect, so the iteration should working on the copy.

POC [Pseudo code]:
```
def callback(ql, *args, userdata = hookhandle):
      ql.hook_del(hookhandle)
```

### DIFF
--- a/qiling/core_hooks.py
+++ b/qiling/core_hooks.py
@@ -59,7 +59,7 @@ class QlCoreHooks:
         ql, hook_type = pack_data
         catched = False
         if hook_type in self._hook.keys():
-            for h in self._hook[hook_type]:
+            for h in self._hook[hook_type][:]:
                 if h.check(ql, intno):
                     catched = True
                     ret = h.call(ql, intno)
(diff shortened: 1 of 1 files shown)


wtdcode:
One thing you have to know: Deleting the hook itself during the callback context is not safe. See https://github.com/unicorn-engine/unicorn/issues/492 .

rootkiter:
Not only delete, but also add and replace. This patch will make it more secure to adjust hooklist.

wtdcode:
> Not only delete, but also add and replace. This patch will make it more secure to adjust hooklist.

Even adding new hooks isn't safe for Unicorn. Unicorn implicitly only guarantees it safe to add hooks before `uc_emu_start`.

rootkiter:
> > Not only delete, but also add and replace. This patch will make it more secure to adjust hooklist.
> 
> Even adding new hooks isn't safe for Unicorn. Unicorn implicitly only guarantees it safe to add hooks before `uc_emu_start`.

In my test, didn't found that. When something goes wrong on it, I'll patch it.

wtdcode:
See https://github.com/unicorn-engine/unicorn/issues/537<https://github.com/unicorn-engine/unicorn/issues/537#issuecomment-860202200>

PR #831, opened 2021-06-25
append ql_syscall_rmdir

rootkiter:
Reference:

https://pubs.opengroup.org/onlinepubs/007904875/functions/rmdir.html


xwings:
Cool

Thanks!

Issue #852, opened 2021-07-07
Modify dll base address: msvbvm60.dll use hard-coded address in its instructions.


frozenkp:
**Describe the bug**
We found that msvbvm60.dll uses several hard-coded addresses in its instructions. These addresses are based on the base address recorded in the dll header, and the base address should be `0x66000000`. I read the code in `loader/pe.py`. The `load_dll` function maps each dll one by one from `0x10000000`.

Here are the logs. Please look at instructions and machine codes in `0x102b2c96`, `0x102b2c9b`, and `0x102b2cbc`.
```
[+] Sample is 32 bit
[+]     Profile: Default
[+]     Map GDT at 0x30000 with GDT_LIMIT=4096
[+]     Write to 0x30018 for new entry b'\x00\xf0\x00\x00\x00\xfeO\x00'
[+]     Write to 0x30028 for new entry b'\x00\xf0\x00\x00\x00\x96O\x00'
[+]     Write to 0x30070 for new entry b'\x00`\x00`\x00\xf6@\x00'
[+]     Write to 0x30078 for new entry b'\x00\x00\x00\x00\x00\xf6@\x06'
[+]     Windows Registry PATH: examples/rootfs/x86_windows\Windows\registry
[=]     Initiate stack address at 0xfffdd000
[=]     Loading c:\Users\VMadmin\Desktop\pestring\justificante_8726.exe\justificante_8726.exe to 0x400000
[=]     PE entry point at 0x401314
[=]     TEB addr is 0x6000
[=]     PEB addr is 0x6044
[=]     Loading examples/rootfs/x86_windows\Windows\System32\ntdll.dll to 0x10000000
[!]     Warnings while loading examples/rootfs/x86_windows\Windows\System32\ntdll.dll:
[!]      - SizeOfHeaders is smaller than AddressOfEntryPoint: this file cannot run under Windows 8.
[!]      - AddressOfEntryPoint lies outside the sections' boundaries. AddressOfEntryPoint: 0x0
[=]     Done with loading examples/rootfs/x86_windows\Windows\System32\ntdll.dll
[=]     Loading examples/rootfs/x86_windows\Windows\System32\kernel32.dll to 0x101a3000
[=]     Done with loading examples/rootfs/x86_windows\Windows\System32\kernel32.dll
[=]     Loading examples/rootfs/x86_windows\Windows\System32\msvbvm60.dll to 0x10288000
[...]

PR #855, opened 2021-07-08
Let DLL files load to their preferred base address

elicn:
Addressing #852 

Changelog:
- DLL files will load to their preferred base address, if available
- Cached DLL files no longer have the base address in their names
- Cached DLL files will try to load to the same address; if it is not available, the cached content is useless and the DLL will reload to have its symbols relocated according to the alternate address
- PE cache test adjusted accordingly

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -25,20 +25,22 @@ class QlPeCacheEntry:
 
 # A default simple cache implementation
 class QlPeCache:
-    def create_filename(self, path, address):
-        return path + ".%x.cache2" % address
+    def create_filename(self, path: str) -> str:
+        return f'{path}.cache2'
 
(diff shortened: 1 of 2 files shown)


elicn:
The only test that fails is Sality, which loads `ntoskrnl.exe` that has a preferred base address beyond 32-bits maxval.
It turns out that `examples/rootfs/x86_windows/Windows/System32/ntoskrnl.exe` is a 64-bit file, loaded by 32-bit system. It doesn't make much sense to me.. does anybody know whether this should be the case?

xwings:
Ah. You got it fix. Nice. I will take a look tomorrow and merge it. Time to sleep, been a long day.

xwings:
Nice, thanks!

PR #859, opened 2021-07-11
correct the parameters of mips-syscall, the same as other CPUs.

rootkiter:
Didn't repair it completely last time( https://github.com/qilingframework/qiling/pull/824).

It should be OK this time.

### DIFF
--- a/qiling/os/posix/posix.py
+++ b/qiling/os/posix/posix.py
@@ -123,8 +123,8 @@ class QlOsPosix(QlOs):
                 self.ql.reg.a1,
                 self.ql.reg.a2,
                 self.ql.reg.a3,
-                self.ql.reg.sp + 0x10,
-                self.ql.reg.sp + 0x14
+                self.ql.stack_read(0x10),
+                self.ql.stack_read(0x14),
             )
 
         def __syscall_args_x86():
--- a/qiling/os/posix/syscall/mman.py
+++ b/qiling/os/posix/syscall/mman.py
@@ -90,8 +90,6 @@ def syscall_mmap_impl(ql, addr, mlen, prot, flags, fd, pgoffset, ver):
     if (ql.archtype == QL_ARCH.ARM64) or (ql.archtype == QL_ARCH.X8664):
         fd = ql.unpack64(ql.pack64(fd))
     elif (ql.archtype == QL_ARCH.MIPS):
-        fd = ql.unpack32s(ql.mem.read(fd, 4))
-        pgoffset = ql.unpack32(ql.mem.read(pgoffset, 4))
         MAP_ANONYMOUS = 2048
         if ver == 2:
             pgoffset = pgoffset * 4096
--- a/qiling/os/posix/syscall/sched.py
+++ b/qiling/os/posix/syscall/sched.py
@@ -48,9 +48,6 @@ def ql_syscall_clone(ql, clone_flags, clone_child_stack, clone_parent_tidptr, cl
         clone_child_tidptr = clone_newtls
         clone_newtls = ori_clone_newtls
 
-    if ql.archtype== QL_ARCH.MIPS:
-        clone_child_tidptr = ql.unpack32(ql.mem.read(clone_child_tidptr, 4))
-
     f_th = ql.os.thread_management.cur_thread	
     newtls = None
     set_child_tid_addr = None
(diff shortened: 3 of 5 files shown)


elicn:
It is not clear why you had to make these changes, and unless I am missing something here - they look incorrect according to MIPS spec. You changed the 5<sup>th</sup> and 6<sup>th</sup> system calls arguments to a derefernece of themselves (i.e. from pointers to pointed value), and then changed all their usages from dereferencing the pointer to using the value.

From the changes you made it seems like the 5<sup>th</sup> and 6<sup>th</sup> system calls arguments are always pointers, which is a wrong assumptions to make. That would also prevent a possible enhancement to replace the arguments (like we do on regular function calls) with other ones, because the original pointers were gone: they were replaced with the values they point to.

Can you please clarify why did you make these specific changes?

elicn:
The change you made to the 5<sup>th</sup> and 6<sup>th</sup> MIPS system calls arguments seems to break the principal of returning the _location_ of the arguments rather than their _values_. Using args locations allow hooks to change them and provide alterate set of arguments.

rootkiter:
> The change you made to the 5th and 6th MIPS system calls arguments seems to break the principle of returning the _location_ of the arguments rather than their _values_. Using args locations allow hooks to change them and provide alterate set of arguments.

The parameter value obtained by hook function is always copied and cannot be used for modification. If we want to modify the execution environment, we always need ql.reg.XX = XXX or ql.mem.write(). 

In MIPS, the return value of the function is passed through the V0 register(qiling/os/posix/posix.py:line-57), so the stack reading operation does not affect the return value of the function.

elicn:
I wasn't talking about the return value of the function / syscall, but returning the arguments locations (basically, what the `get_syscall_args` method does). The intention is to do something similar to what we have in `fcall` (see the `call` method in `qiling/os/fcall.py`).

Anyway, I don't think having args 5-6 handled differently than 1-4 is the right way to go here..

rootkiter:
There will always be data on the stack, so stack reading will not affect the stability of the code.

rootkiter:
In MIPs, 1-4 is different from 5-6, which is determined by instruction set, not us.

elicn:
You misunderstood my comments..

MIPS ABI defines that first 4 args are passed through registers, where the rest are passed on the stack; we are completely aligned here. I just wondered why you changed the way args 5 and 6 are returned by `get_syscall_args` and handled afterwards. While the contents of args 1-4 would need to be read, args 5 and 6 would already hold their values. The intent was that the args list would be a list of _locations_ rather than _values_.

I think that you are confused by the way the regs are read: because they are Python properties, they are read by simply referencing them in the code, and you wanted it to be similar for args 5 and 6. That, however, breaks the principle of returning the locations and not values. The reason for returning locations is to maintain the ability for hooks to patch them; if you replace them with the values - that cannot be done anymore.

rootkiter:
image

The stack pointer is not a parameter, but the data it points to.

elicn:
Yes, that's right - and you deleted this piece of code to accomodate it to your change.

I want to maintain the ability to do things like:
`ql.mem.write(fd, some_other_value)`
Which would not be possible anymore if you change the 5 and 6 args to hold their values.

The right way to go here will be to refactor the code to use the `qiling/cc` methodology.

rootkiter:
http://math-atlas.sourceforge.net/devel/assembly/mipsabi32.pdf

"Argument Passing"  3-17

https://stackoverflow.com/questions/2298838/mips-function-call-with-more-than-four-arguments

rootkiter:
In a word, I think this kind of realization is right. Then I mentioned pr.

elicn:
Why are you keep referring me to the MIPS specification..?
Isn't it clear already that I know how MIPS ABI is defined?

Your code change **breaks consistency of how the arguments are handled**. Unless there is a really good reason to do that (that is, break the consistency), I don't think it should be applied.

wtdcode:
@rootkiter 我觉得他的意思是他知道mips的调用约定，但是对于qiling中获取参数的方式应该参考cc.py那个文件

@elicn Do you refer to his previous PR mentioned in the description?

elicn:
@wtdcode both the mentioend PR and this one.

As far as I see, it deosn't really fix anything - it just changes that way the arguments are handled (i.e. holds the dereferenced value on the stack rather than the address on the stack). Like I said multiple times here, holding the values rather than the locations is a bit probelmatic.

Anyway. if we were to change that to a more elegant implementation, that would make use of `qiling/cc` classes.

rootkiter:
The function name which I patch is  `__syscall_args_mips` , not `__syscall_args_address_mips`. What `ql_syscall_xxx` need is parmeters, not parmeter's address. That is my point. 

Maybe it's right for you to use `qiling/cc` class. But I don't understand the difference.

elicn:
Hey guys, please see #860 and let me know what you think.
POSIX syscall changes are courtesy of @rootkiter

PR #860, opened 2021-07-11
Refactor POSIX syscalls handling to use CC methodology

elicn:
Changelog Highlights:
- POSIX syscalls arguments and retval are now accessed using the `cc` abstraction
- Fixed a bug in MIPS `cc` where args on the stack were accessed as if there was a return address (and there is not)
- POSIX syscalls adjusted to meet the change in arguments access (as made by @rootkiter !)

This PR comes to enhance and replace PR #859


xwings:
Nice, thanks :)

sorry for the delay

PR #862, opened 2021-07-12
Adjust mmap syscall to handle subsequent calls to the same region

danielmoos:
The loading of shared libraries in QNX performs multiple mmap calls on the same memory region. The first mmap call has the form
mmap(addr = 0, len = sizeof library, prot = PROT_NONE, flags = MAP_PRIVATE | MAP_LAZY | MAP_ANONYMOUS)
then for the different ELF sections of the library mmap calls are performed to parts of the memory region returned from this first mmap call. The current implementation of mmap has serveral problems:

* calls to mapped regions use the aligned effective size to call mem.protect(), but the mem.protect() function also performs page alignment and this double alignment of the size can result in sizes larger than the memory page
* every call to mmap causes a zero-initialization which (in combination with page alignment) can cause access to unmapped memory
* usage of the effective size for add_mapinfo() can also exceede the page size and results in the failure of subsequent mmap calls

The changes in this pull request result in a functioning mmap under QNX for my application, but this mixing of memory page alignment in the mmap implemenation seems a bit fragile. I guess this could be improved in the future.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/posix/syscall/mman.py
+++ b/qiling/os/posix/syscall/mman.py
@@ -106,6 +106,9 @@ def syscall_mmap_impl(ql, addr, mlen, prot, flags, fd, pgoffset, ver):
     mmap_base = addr
     need_mmap = True
     eff_mmap_size = ((mlen + 0x1000 - 1) // 0x1000) * 0x1000
+    # align eff_mmap_size to page boundary
+    aligned_address = (addr >> 12) << 12
+    eff_mmap_size -= mmap_base - aligned_address
 
     # initial ql.loader.mmap_address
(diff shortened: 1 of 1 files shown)


danielmoos:
Fascinating, some tests are failing. Does someone immediately spots the part of my changes that is responsible?

danielmoos:
It seems like we can not move the zeroing of the page into the "if need_map" block. I'll will look into alternatives tomorrow.

xwings:
Cool. Nice work.

PR #866, opened 2021-07-22
Windows SDK Refactoring

elicn:
Changelog highlights:
- All Windows SDK hooks were refactored to have their parameters defined within the `winsdkapi` decorator, instead of being loaded from SDK JSON files at runtime and have a parameters override list
- The `winsdkapi` decorator was significantly reduced, removing the JSON loading and parameters override functionalities (that is, getting rid of `replace_params` and `replace_params_type`). That reduces the overhead of `winsdkapi` decorator and makes it consistent with other OS API decorators
- Moved some dislocated API hooks to their appropriate locations
- Some duplicated hooks were found and removed
- Some hooks were re-implemented to better match their corresponding API spec
- Some bugs were fixes
- Typing annotations
- Cleaned imports clutter
- Numerous styling imporvements


qiling/os/windows/dlls/ntoskrnl.py
```diff
@@ -899,14 +902,17 @@ def hook_IoCreateDriver(ql: Qiling, address: int, params):
     init_sp = ql.os.init_sp
 
     ql.os.fcall = ql.os.fcall_select(STDCALL)
-    ql.os.fcall.writeParams(((POINTER, ql.driver_object_address), (POINTER, ql.regitry_path_address)))
+    ql.os.fcall.writeParams((
+        (POINTER, ql.driver_object_address),
```
xwings, line 906: maybe we should move this to ql.os and not expose it to the core ?

this must be some old code i over looked

elicn: Move which part exactly..?
This was a manual function call code I replaced with the new-style fcall. Didn't really got into why it is there at the first place.

We can fix that in another PR.

xwings:
Cool. Thanks

PR #870, opened 2021-07-28
Flexible Linux set_api hooks

elicn:
Linux `set_api` hooks will work only if set up before `ql.run()`, which means all hooks should be set up ahead of time.
This little modification gives more flexibility by allowing `set_api` hooks to be added as necessary during emulation time.

For example: hooking an API only after entering a specific function, when the hook becomes relevant [and not before, potentially trapping irrelevant calls to the same API]

That code change also fixes a bug where the `intercept` value was passed as the hook `user_data`.

### DIFF
--- a/qiling/os/linux/linux.py
+++ b/qiling/os/linux/linux.py
@@ -3,13 +3,14 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
+from typing import Callable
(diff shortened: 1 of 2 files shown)


xwings:
Hi,

Can we also have an example and also a test?

elicn:
Sure. Let me work on it and add it to the PR.

PR #891, opened 2021-08-19
Re-implemented Linux crackme example

elicn:
Changelog:
- Greatly improved the Linux cracme example by re-implementing it and commenting each step for extra clarity. The new implementation saves and restores Qiling state each round, instead of instantiating a new Qiling object. That makes it faster and smarter example to learn from. Replaces #781
- Fixed Linux system call parameters enumeration to use `inspect` correctly, which now lets the system calls methods use typing annotations
- A few negligible changes

### DIFF
--- a/examples/crackme_x86_linux.py
+++ b/examples/crackme_x86_linux.py
@@ -6,91 +6,129 @@
 import sys
 sys.path.append("..")
 
-import os
+import string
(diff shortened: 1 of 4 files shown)


xwings:
Cool. we need that. badly :)

Issue #333, opened 2020-06-12
TODO List - Short Term Goal


xwings:
**Windows:**
- try to run malware/Windows PE on qiling, find and implement the missing APIs

**Linux:**
- Try all the Linux binaries under /bin/, find & implement missing syscalls
- Try IoT firmware binaries & implement missing syscalls

**Documents:**
- Write more tutorials for Qiling Framework at https://docs.qiling.io (https://github.com/qilingframework/rtfd.io)
- Improved the documentation on Qiling Framework at https://docs.qiling.io (https://github.com/qilingframework/rtfd.io)

**Tools**
- Building a new tool on top of Qiling Framework. Example, qltools that comes with Qiling Framework.

**Misc:**
- New feature can be implemented into QIiling Framework
- New tools can be build ontop of Qiling Framework

lucroida:
Suggestion: contribution guidelines (see https://github.com/qilingframework/qiling/issues/479)

nullableVoidPtr:
> implement the missing APIs
> implement missing syscalls

I think a milestone or a separate issue with a checklist would be of great help to new contributors, like the following:
> - [ ] `mmap`
> - [ ]  `mprotect`

and so on.

xwings:
Cool, can you help to list it out.

In that case we need to provide  a binary too and more detailed example :)

wtdcode:
Maybe a developer guide would help?

________________________________

nullableVoidPtr:
> Javascript
How would this work? If we can somehow link in V8, we can also put in WASM support if need be, but would Qiling hook into TurboFan or something like that?

wtdcode:
> #479

Finally, here it comes :).

See https://docs.qiling.io/en/latest/contribution/ .

xwings:
> > Javascript
> 
> How would this work? If we can somehow link in V8, we can also put in WASM support if need be, but would Qiling hook into Ignition or something like that?

No idea with Javascript yet. there are wish list. some of them got no clear path and idea yet.

PR #892, opened 2021-08-20
Fix restore causing unicorn cpu exception

crass:
When restoring a snapshot created with reg=True a unicorn cpu exception can
be thrown when restoring the GS register. It looks like for at least 32bit
mode on X86 processors unicorn will validate memory structures pointed to by
loaded GS reg. If the memory is not loaded the validation will surely fail,
and thus the exception. The fix is then to load the memory maps first,
before loading the registers. Note, this validation does not happen when
restoring via cpu_context.

Included is a test case which will have errors if this fix is not applied.

### Coding convention?

### Extra tests?

### Changelog?

- [x] Changelog will be updated after some proper review, if needed

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -804,14 +804,14 @@ class Qiling(QlCoreHooks, QlCoreStructs):
             with open(snapshot, "rb") as load_state:
                 saved_states = pickle.load(load_state)
 
+        if "mem" in saved_states:
+            self.mem.restore(saved_states["mem"])
+
         if "cpu_context" in saved_states:
             self.arch.context_restore(saved_states["cpu_context"])
 
         if "reg" in saved_states:
             self.reg.restore(saved_states["reg"])
-
-        if "mem" in saved_states:
-            self.mem.restore(saved_states["mem"])
 
         if "fd" in saved_states:
             self.os.fd.restore(saved_states["fd"])
--- a/tests/test_elf.py
+++ b/tests/test_elf.py
@@ -60,6 +60,48 @@ class ELFTest(unittest.TestCase):
 
         del ql
 
+    def _test_elf_linux_x86_snapshot_restore_common(self, reg=False, ctx=False):
+        rootfs = "../examples/rootfs/x86_linux"
+        cmdline = ["../examples/rootfs/x86_linux/bin/x86_hello"]
+        snapshot = os.path.join(rootfs, 'snapshot_restore_reg_ctx.snapshot')
+
+        ql = Qiling(cmdline, rootfs, verbose=QL_VERBOSE.DEBUG)
+
+        X86BASE = int(ql.profile.get("OS32", "load_address"), 16)
+        hook_address = X86BASE + 0x542 # call printf
+
+        def dump(ql):
+            nonlocal snapshot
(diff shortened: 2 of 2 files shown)


crass:
Also, I've added the new test_snapshot.py file to the linux script for running all its tests. I think the test should be capable of running on windows, but can't test that right now. The only reason why it might not run on windows is the hardcoded path separators which are different between linux and windows (but then again, python might be smart enough to convert them).

xwings:
The test should be in test_elf. but again, we already have the same test? any reason why we need a new code and there are no changes in the code ?

crass:
There is a change to the code, see changes to file `qiling/core.py `. Or am I misunderstanding you? What is the "same test" you are referring to? Can you point me to the specific test?

xwings:
besides moving the location for mem restore. i did not see any changes in core.py

xwings:
There are few similar snapshot test eg,

https://github.com/qilingframework/qiling/blob/35db51a78a6764cb996bf7852fbdb3d5b912eaac/tests/test_elf.py#L33

will you be able to move it to your new snapshot test ?

crass:
> besides moving the location for mem restore. i did not see any changes in core.py

Moving the location of mem restore _is_ the fix. That way memory is mapped before the GS register is loaded.

> will you be able to move it to your new snapshot test ?

Yes, I can move that to the `test_snapshot.py` file. I will also add `test_snapshot.py` to windows tests and make sure the path issue I mentioned above is not a problem.

xwings:
> Yes, I can move that to the `test_snapshot.py` file. I will also add `test_snapshot.py` to windows tests and make sure the path issue I mentioned above is not a problem.

linux and windows test always seperate. due to dll restriction. So, you can either

1. move linux snapshot test to test_elf, and Windows snapshottest to test_pe
2. make two different snapshot test, test_snapshot_elf and test snapshot_pe

method 1 is cleaner

crass:
> > Yes, I can move that to the `test_snapshot.py` file. I will also add `test_snapshot.py` to windows tests and make sure the path issue I mentioned above is not a problem.
> 
> linux and windows test always seperate. due to dll restriction. So, you can either
> 
>     1. move linux snapshot test to test_elf, and Windows snapshottest to test_pe
> 
>     2. make two different snapshot test, test_snapshot_elf and test snapshot_pe
> 
> 
> method 1 is cleaner

Now I'm thinking about this again. and I think option 1 is better, to put my test in `test_elf.py`. And in a separate change we should get **all** linux tests running in windows. I don't think I should put the test in test_pe.py` because its not technically a PE test.

I agree that the linux and windows tests should be separate, but I think that `test_pe.bat` should be renamed to `test_onwindows.bat` and `test_onwindows.bat` should include all windows tests. As you say, we can not put all the windows tests in `test_onlinux.sh` because of the dll restriction.

xwings:
There are some issue with elf test runs on windows. that is the reason why we did not run linux test on windows.

crass:
What are some of the issues?

xwings:
checkout issue #333 or just run the test in windows.

xwings:
merge for now

Issue #877, opened 2021-08-07
Failed to hijack stdin using the example from the doc


aesophor:
**Describe the bug**
I followed [this page in the documentation](https://docs.qiling.io/en/latest/hijack/), but I couldn't successfully hijack stdin.

**Sample Code**
Unable to hijack stdin (extracted and modified from the doc):
```python
ql = Qiling(path, rootfs)
ql.stdin = StringBuffer()
ql.stdin.write(b"AAAA\n")
```

Able to hijack stdin:
```python
ql = Qiling(path, rootfs, stdin=StringBuffer())
ql.stdin.write(b"AAAA\n")
```

**Expected behavior**
The doc showed this example, and users should be able to hijack stdin with:
```python
ql = Qiling(path, rootfs)
ql.stdin = StringBuffer()
```

**Screenshots**
(None)

**Additional context**
(None)

elicn:
Thanks for reporting this.

It looks like the reason for that is the OS initialization that takes the current values of `stdin`, `stdout` and `stderr` at the time of Qiling instance initialization and do not update it upon setting new values. To solve this, the OS object would need to both run:
```python
if self.ql.stdin != 0:
     self.stdin = self.ql.stdin

if self.ql.stdout != 0:
    self.stdout = self.ql.stdout

if self.ql.stderr != 0:
    self.stderr = self.ql.stderr
```
and
```python
self._fd[0] = self.stdin
self._fd[1] = self.stdout
self._fd[2] = self.stderr
```
at the time of OS `run` (confirmed to work for Linux OS).
That said, an elegant fix will most likely have to standardize the access to `stdin`, `stdout` and `stderr` across all OS types (perhaps by turning them into properties)

elicn:
@xwings 
I am thinking of detaching `stdin`, `stdout` and `stderr` from the `Qiling` class and move them to OS as properties.
That would help offloading some of the `Qiling` accessors to their appropriate locations.
What do you think?

xwings:
so u gonna set using ql.os.stdin ?

i dont remember why we put it in core. but u can try

elicn:
Done (see #898)
@aesophor, let me know whether that solved the problem (once the PR merges).

aesophor:
@elicn
Many thanks for your time and effort. I'll let you know once it has been merged.

xwings:
looks ok, let me do some test. should be quick

elicn:
@aesophor merged.

[...]

PR #898, opened 2021-08-22
Implement standard streams as QlOs properties instead of Qiling's

elicn:
This PR addresses issue #877 .

Changelog:
- Made standard streams (i.e. `stdin`, `stdout` and `stderr`) `QlOs` properties instead of `Qiling`'s.
   Now they may be accessed by:
  - `ql.os.stdin`
  - `ql.os.stdout`
  - `ql.os.stderr`
  That should improve readability by offloading properties from the core to where they intuitively belong.
  Additionally, that would enable users to switch standard streams at any point during emulation instead of up until calling `ql.run`.
- Typing annotations FTW

### DIFF
--- a/examples/crackme_x86_linux.py
+++ b/examples/crackme_x86_linux.py
@@ -74,7 +74,7 @@ class Solver:
         hobj = self.ql.hook_code(__count_instructions)
 
         # feed stdin with input
-        self.ql.stdin.write(input + b'\n')
+        self.ql.os.stdin.write(input + b'\n')
 
         # resume emulation till function returns
         self.ql.run(begin=self.replay_starts, end=self.replay_ends)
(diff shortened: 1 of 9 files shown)


xwings:
no changes in Linux stdin ?

xwings:
Sorry!

i mean you updated examples/crackme_x86_linux.py  and did not update test_elf ?

i did not check the code yet. just asking

elicn:
No, I didn't have to.
The crackme-based test is implemented differently and creates a new `Qiling` instance each round instead of re-using the same one. When it does that, it uses the `Qiling` constructor arg to redirect `stdin`; it doesn't do that post-initialization, as the doc suggested (which was broken on POSIX till now).
The new and improved example re-uses the same instance [by saving the state and restoring it each round] and needs to feed the `stdin` stream with new input. Since the `stdin` pipe was declared elsewhere, the bruteforcing method needs to refer it by `ql.os.stdin`.

xwings:
cool. merge!

PR #900, opened 2021-08-25
Mock pipes extention

elicn:
The `MyPipe` and `StringBuffer` classes are used by various tests and examples to redirect the process standard input and output streams. However, they are re-implemented each and every time. This new extentions module implements a few handy mock pipes that can be used to replace them (and reduce code duplication):
- `SimpleStringBuffer` - basic bufferred FIFO pipe
- `SimpleInStream` - basic input stream, may be used to mock `stdin`
- `SimpleOutStream` - basic output stream, may be used to mock `stdout` or `stderr`
- `NullOutStream` - nullified output stream, may be used to disregard program output (as if it was redirected to `/dev/null`)

Note: Windows program may also use `SimpleStringBuffer` to mock their standard IO stream, however using the `SimpleInStream` and `SimpleOutStream` classes would provide a better clarity.

Example:
```python3
from qiling import Qiling
from qiling.extensions import pipe

mock_stdin = pipe.SimpleInStream(sys.stdin.fileno())
mock_stdout = pipe.NullOutStream(sys.stdout.fileno())

# create a silent qiling instance
ql = Qiling(argv, rootfs,
    verbose=QL_VERBOSE.OFF, # thwart qiling logger output
    stdin=mock_stdin,       # take over the input to the program using a fake stdin
    stdout=mock_stdout)     # disregard program output

# ...

ql.os.stdin.write(b'hello world!\n')
```

### DIFF
--- a/examples/crackme_x86_linux.py
+++ b/examples/crackme_x86_linux.py
@@ -7,39 +7,23 @@ import sys
 sys.path.append("..")
 
 import string
-from typing import TextIO
 
 from qiling import Qiling
-from qiling.os.posix import stat
+from qiling.const import QL_VERBOSE
+from qiling.extensions import pipe
 
 ROOTFS = r"rootfs/x86_linux"
 
-class MyPipe(TextIO):
-    def __init__(self):
-        self.buf = bytearray()
-
-    def write(self, s: bytes):
-        self.buf.extend(s)
-
-    def read(self, size: int) -> bytes:
-        ret = self.buf[:size]
-        self.buf = self.buf[size:]
-
-        return bytes(ret)
-
-    def fileno(self) -> int:
-        return sys.stdin.fileno()
-
-    def fstat(self):
-        return stat.Fstat(self.fileno())
-
 class Solver:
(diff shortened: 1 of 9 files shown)


xwings:
@kabeor just reminded me, why we have that messy stdio implementation.

The main reason is we need to do for IDA plugin. We need to check again.

xwings:
lets hold this PR for the moment until we test IDA pro

elicn:
I don't think they are actually related.

These new classes only come to reduce code duplication, and replace the `MyPipe` and `StringBuffer` classes that were used across various tests and examples. They do not replace any fundamental mechanism, rather they serve as conveniency classes [and this is why they are placed in extentions].

Let's wait for the IDA test, but as far as can tell IDA uses the `QLStdIO` class which is based on `ql_file`. It looks like this is entirely unrelated.

xwings:
i mean the previous PR (which i already merge). not this

elicn:
OK, that makes more sense now.
However, I didn't change anything around IDA iostreams so they _should_ behave the same..
Let me know what you find.

PR #903, opened 2021-08-29
Arch Refactor

elicn:
Changelog:
- Vastly refactored `arch` classes
- Intel arch classes are now consolidated into a single file
- Eliminated many `if ql.archtype` conditions in favor of OOP inheritence
- The `create_assembler` and `create_disassembler` mtehods now cache created instances instead of re-creating them every single time (except ARM, which has to do that to support thumb switch)
- Type annotations and comments
- Cleaner and nicer code

@xwings, I am not sure whether EVM is included in the current test suit; if not, I would appreciate if you could test this for EVM.


xwings:
Yup. EVM included in the current test.

This does looks better!

Issue #904, opened 2021-08-29
Memory search


nescafename:
https://github.com/qilingframework/qiling/blob/4cae8b6cca92ec9423ba0d8e9a8bb814be2ace10/qiling/os/memory.py#L315

Is that supposed logic? I don't understand why mem.search doesn't cover the case(3) when full region(begin-end) in range of mapped region(lbound-ubound)?

```js
   begin                                        end
1) [                      [                      ] 
                        lbound

   begin                                        end
2) [                      ]                      ] 
                        ubound

   lbound                                              ubound
3) [                      [           ]                   ] 
                      begin           end
```

ref: https://stackoverflow.com/questions/325933/determine-whether-two-date-ranges-overlap/325964#325964

example:

![Snipaste_2021-08-29_23-34-25](https://user-images.githubusercontent.com/77633834/131264608-d2298cac-7b5f-4420-8839-f9cc5e9f9c71.png)
```
[PE]: 0x18026830b <= 0x180000000 < 0x18026833d or 0x18026830b < 0x1802a7000 <= 0x18026833d
```

It won't return any occurences if full range in mapped region.
Meanwhile:
```python
def overlap(start1, end1, start2, end2):
    """Does the range (start1, end1) overlap with (start2, end2)?"""
    return not (end1 < start2 or end2 < start1)

def mem_search(ql, needle: bytes, begin: int = None, end: int = None):

    assert begin < end, 'search arguments do not make sense'

    ranges = [(max(begin, lbound), min(ubound, end)) for lbound, ubound, _, _ in ql.mem.map_info if overlap(begin, end, lbound, ubound)]

    results = []

    for lbound, ubound in ranges:
        haystack = ql.mem.read(lbound, ubound - lbound)
        local_results = (match.start(0) + lbound for match in re.finditer(re.escape(needle), haystack))

        results.extend(local_results)

    return results
```
works fine

elicn:
This is indeed a good catch; the existing code seems to overlook that case.
Would you create a PR for it?

xwings:
Good catch indeed. Please create a PR and thanks @andrwnoskov in advance

PR #905, opened 2021-08-31
Fix memory search to handle range overlaps correctly

andrwnoskov:
Fix issue #904 

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ VERSION = gb['__version__']
 requirements = [
     "capstone>=4.0.1",
     "unicorn>=1.0.2",
-    "pefile>=2019.4.18",
+    "pefile==2021.5.24",
     "python-registry>=1.3.1",
     "keystone-engine>=0.9.2",
     "pyelftools>=0.26",


andrwnoskov:
Please note, the initial code was
```(begin <= lbound < end) or (begin < ubound <= end)```, which probably had to be
```(lbound <= begin < ubound) or (lbound < end <= ubound)```, but meant `(lbound <= begin < ubound) or (lbound < end < ubound)`

Isn't it?

andrwnoskov:
Current code treats ranges as `[a, b]`, which means if we have intersection on the ends, there will be an overlap.
For example: `[1, 5] and [5, 8] - overlap`

Is that correct? Or it should be `[1, 5) and [5, 8] - no overlap`

xwings:
Thanks and super thanks! few suggestion
1. update change log, this is a important fix
2. add in a test in test_elf.py. make sure no one mess with this next time.

elicn:
There is no problem of overlap really.. the range is used to calculate base + size, so effectively there is no overlap.
Taking you example, where there are two ranges `[1, 5]` and `[5, 8]`: the first range will be taken as `base = 1` and `size = 5 - 1`, so no overlap here.

xwings:
Not too sure this got any issue with PE loader. Need to investigate abit more.

xwings:
okay, issue was the pefile module. i downgraded to  one older version.

PR #914, opened 2021-09-14
POSIX Fixes

elicn:
Changelog:
- Reverted some changes made recently to log output which could get it to break
- Removed unnecessary safeguard from `mem.write`, aligning it to `mem.read` behavior

### DIFF
--- a/qiling/os/memory.py
+++ b/qiling/os/memory.py
@@ -281,12 +281,7 @@ class QlMemoryManager:
             data: bytes to write
         """
 
-        try:
-            self.ql.uc.mem_write(addr, data)
-        except:
-            self.ql.log.debug(f'addresss write length: {len(data):d}')
-            self.ql.log.error(f'addresss write error: {addr:#x}')
-            raise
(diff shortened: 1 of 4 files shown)


qiling/os/memory.py
```diff
@@ -281,12 +281,7 @@ def write(self, addr: int, data: bytes) -> None:
             data: bytes to write
         """
 
-        try:
-            self.ql.uc.mem_write(addr, data)
-        except:
-            self.ql.log.debug(f'addresss write length: {len(data):d}')
-            self.ql.log.error(f'addresss write error: {addr:#x}')
```
xwings, line 288: Better for us to keep the debug msg ?

elicn: It's a matter of approach. I would suggest to leave it unguarded and let the user decide whether to catch exceptions or let their program crash.
Anyway, `read` and `write` should share the same approach for consistency.

elicn:
@xwings
Any particular reason this PR is hanging on..? Let me know if there're any concerns.

PR #917, opened 2021-09-16
Fix PE loader to recursivly import symbols and call DllMain

t-moe:
This PR fixes some of the issues with the PE loader:

- DLL Imports were not recursively handled (as also fixed by #886)
- DLLs were not properly relocated, when their preferred base address was already taken (as also fixed by #801)
- Entry point of DLLs (DllMain) were not called
- API sets were not handled properly

Calling DllMain is imperative, in order to initialize constructors (e.g. c++ static initializers and other library init stuff).

The idea proposed by #801 with delaying loading the DLLs until the first symbol is called, raises the question of when to call DllMain(). In this PR I load the DLLs recursively at startup, and also call DLLmain, but I skip the virtual DLLs in the hope that it is faster than #801. Also, by skip loading the virtual DLLs I hope that we can avoid copying the api-*.dlls entirely (which caused build problems in #886).

Unfortunately I was not able to run the testcases on my machine. Can you run the workflow?  

Credits go to @crass (#886). I've copied his idea with the `init_import` method.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -77,6 +77,11 @@ class Process():
 
             path = os.path.join(self.ql.rootfs, self.ql.dlls, dll_name)
 
+        if dll_name.startswith("api-ms-win-"):
+            self.ql.log.warning("Refusing to load virtual DLL %s" % dll_name)
+            # Usually we should not reach this point and instead imports from such DLLs should be redirected earlier
+            return 0
+
         if not os.path.exists(path):
             raise QlErrorFileNotFound("Cannot find dll in %s" % path)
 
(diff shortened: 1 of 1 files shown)


Enteee:
No worries. Just wanted to point it out. There might be more PRs that were closed this way.

elicn:
@wtdcode, about 10 PRs were closed this way.
I thought it was intentional..

elicn:
Changes were included as part of #1118 .
Thanks @t-moe !

PR #920, opened 2021-09-16
Refactored QlRegisterManager

elicn:
Changelog:
- Reduced `reg` access a little bit by removing unnecessary checks
- Fixed a bug where an MSR write with a value of `0` was considered as a read
- Eliminated mutual dependency of `reg.bit` and `arch.get_reg_bit`
- Removed the `reverse_mapping` member along with the `create_reverse_mapping` method as they are required only for retreiving a single reg; replaced their functionality with a more memory-friendly one
- Nicer Python code
- Typing annotations and comments


xwings:
Nice work

PR #924, opened 2021-09-22
UnboundLocalError, use before assignment, fixing of bytes.hex() calls

dannyquist:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/arch/utils.py
+++ b/qiling/arch/utils.py
@@ -36,7 +36,7 @@ class QlArchUtils:
         qd = ql.arch.create_disassembler()
 
         offset, name = self.get_offset_and_name(address)
-        log_data = f'{address:0{ql.archbit // 4}x} [{name:20s} + {offset:#08x}]  {tmp.hex(" "):30s}'
+        log_data = f'{address:0{ql.archbit // 4}x} [{name:20s} + {offset:#08x}]  {tmp.hex():30s}'
         log_insn = '\n> '.join(f'{insn.mnemonic:20s} {insn.op_str}' for insn in qd.disasm(tmp, address))
 
         ql.log.info(log_data + log_insn)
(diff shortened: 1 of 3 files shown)


elicn:
Why removing the space parameter from the `hex` represenation?
It's not a bug, it is there to put a space between hex bytes, which makes long hex strings more readable (e.g. `'68 65 6c 6c 6f 20 77 6f 72 6c 64'`)

dannyquist:
Any chance you could DM me on twitter? I'm @openmalware there

cq674350529:
@elicn In `python 3.6`, call `hex()` with additional paramter will cause error.

```python
Python 3.6.9 (default, Jan 26 2021, 15:33:00) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: tmp = b"00112233"                                                          

In [2]: tmp.hex()                                                                  
Out[2]: '3030313132323333'

In [3]: tmp.hex("")                                                                
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-3-b882162dbcb0> in <module>
----> 1 tmp.hex("")

TypeError: hex() takes no arguments (1 given)
```

elicn:
> @elicn In `python 3.6`, call `hex()` with additional paramter will cause error.

Yes, the reason for that is indeed an older version of Python. We continued our conversation over DM so it's not recorded here, but this is what I told @dannyquist. From looking at the project's test suits it looks like the miminal Python version is 3.8, so passing an argument to `hex` should be supported.

@xwings, perhaps it would be helpful to issue a runtime warning based on the Python version (that is, if the running version is below the one supported, issue a warning to the user)

cq674350529:
Yes, currently there is no explicit python3 version requirement. What I can find in the doc is:
> For this installation guide, Ubuntu desktop 18.04.3 LTS 64bit is the base example (Qiling Framework works in other Linux distributions that run Python 3.5 and above

I'm running with Ubuntu 18.04.6, and the default python3 version is `python 3.6`. At least, most of feaure works with `python 3.6`.

elicn:
Sadly, the documentation is often outdated.. Practically Ubuntu 20.4 / Python 3.8 are used.

Issue #929, opened 2021-09-30
Unable to use GDB to debug


Kirito0:
Environment: gdb 11.0 (or 9.2), qiling 1.3.0
Unable to attach gdb debugging, error "Truncated register 26 in remote'g' packet" appears

```python
#!/usr/bin/env python3
# -*- coding: utf-8 -*-

from qiling import Qiling
from qiling.const import QL_VERBOSE

rootfs = "./rootfs/x8664_linux/"
filepath = "./rootfs/x8664_linux/bin/x8664_hello"

ql = Qiling([filepath],rootfs,verbose=QL_VERBOSE.OFF)

ql.debugger = "gdb:127.0.0.1:9999"
ql.run()
```
![image](https://user-images.githubusercontent.com/16188541/135378823-52e4973a-6574-478a-b73b-bcb61f64795a.png)

But qiling 1.2.4 is ok

xwings:
fixed and closed.

PR #930, opened 2021-09-30
gdb bugfix

elicn:
Fixing #929 

Bug root cause: considering the flags register size as native (i.e. 32 bits on x86 and 64 on x64), however it should always be considered as 32 bits regardless.


xwings:
good catch. thanks

PR #940, opened 2021-10-12
Add support for STM32 series microcontroller simulation

cla7aye15I4nd:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

### DIFF
--- a/examples/mcu/stm32f4_hack_lock.py
+++ b/examples/mcu/stm32f4_hack_lock.py
@@ -27,7 +27,11 @@ def crack(passwd):
     ql = Qiling(["../../examples/rootfs/mcu/stm32f407/backdoorlock.hex"],                    
                         archtype="cortex_m", profile="stm32f407", verbose=QL_VERBOSE.OFF)
     
+    ql.hw.create('spi2')
+    ql.hw.create('gpioe')
+    ql.hw.create('gpiof')
     ql.hw.create('usart1')
+    ql.hw.create('rcc')
 
     ql.hw.show_info()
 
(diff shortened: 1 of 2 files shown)


elicn:
I've just finished reviewing the code and I must say kudos for the huge amount of good work invested in here, this is indeed an impressive work.

Overall the code looks clear, but as I already mentioend - I missed the comments and typing annotations. The code has too many ad-hoc conditions (like those based on `QL_ARCH_HARDWARE` and `if QL_ARCH_NONEOS`) which interfere with the other use cases and mess up the code. Usually such things are addressed elegantly by using OOP, where each object implements its own behavior.

That also relates to the significant changes to the `memory` module, which serve only the MCU arch. I've already discussed that with @wtdcode, and I truely think the MMIO handling should be implemented differently in a way that would not interfere with the other use cases too much.

wtdcode:
> That also relates to the significant changes to the `memory` module, which serve only the MCU arch. I've already discussed that with @wtdcode, and I truely think the MMIO handling should be implemented differently in a way that would not interfere with the other use cases too much.

Just go ahead with the changes you would like. I feel it okay to separate the MMIO implementation elsewhere.

cla7aye15I4nd:
May add a mmio module under hw

cla7aye15I4nd:
> The code has too many ad-hoc conditions (like those based on `QL_ARCH_HARDWARE` and `if QL_ARCH_NONEOS`) which interfere with the other use cases and mess up the code. Usually such things are addressed elegantly by using OOP, where each object implements its own behavior.
I agree, looks like shellcode and evm has the same problem, these code may need to be become more clear.

PR #945, opened 2021-10-15
Fix bytes api misuse bug

SwordFaith:
This pr is about fix SetMem api in UEFI Boot Services. The signature of `SetMem` is `SetMem(void *buffer,  size_t size, uint8 value)`. And the behavior is fill value to memory between buffer and buffer+size. Current implementation use `bytes(value) * size` to generate memory content. But the `bytes(value)` will generate `b'\x00' * value` rather than byte version of `value`. So I and @elicn  discussed, and determined to change it to `bytes(chr(value))` which follow the right behavior.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/uefi/bs.py
+++ b/qiling/os/uefi/bs.py
@@ -499,7 +499,7 @@ def hook_SetMem(ql: Qiling, address: int, params):
 	size = params["Size"]
 
 	byteorder = 'little' if ql.archendian == QL_ENDIAN.EL else 'big'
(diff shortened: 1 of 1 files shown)


elicn:
What about `bytes([value] * size)` ?
I guess that `ql.pack8(value) * size` should work either.

SwordFaith:
> @SwordFaith, can you create a new PR with the suggestions above instead?

Ok, I'll test and update them letter.

elicn:
Watch the parentheses location carefully: your fix is not going to work as intended.

Issue #931, opened 2021-10-04
Qiling can't open port although "ql.root = True" - "bind: Operation not permitted" 


RadCet:
**Describe the bug**
- When trying to emulate httpd service, process try to bind to port 80 or 8080. 
- Already kill any process running at 80 or 8080 before
- But try to run with **sudo** and config qiling with "ql.root = True" still get "bind: Operation not permitted" 
- It happen with both 80 and 8080
- Qiling version 1.3.0
- Notice that when I try to run in a pi 3, everything is ok
 
**Sample Code**
```python
import sys
sys.path.append("..")
from qiling import *
from qiling.const import QL_VERBOSE

def run_sandbox(path, rootfs):
	ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEBUG)
	#ql.debugger = "127.0.0.1:9999"
	ql.root = True
	ql.run()


if __name__ == "__main__":
	run_sandbox(["rootfs/usr/sbin/httpd","-c","/etc/httpd.conf"], "rootfs")
```

**Screenshots**
- port 80
![image](https://user-images.githubusercontent.com/24670183/135818106-fff60b65-0967-4c07-92df-361caf6035ec.png)
- port 8080
![image](https://user-images.githubusercontent.com/24670183/135818348-2b915311-f1bc-4da5-acfc-33a510739a5f.png)

elicn:
Thanks for reporting this, this is indeed a Qiling bug.

*TL;DR* - Debugging a few examples show that setting the `root` property after instantiating Qiling is rather too late.

*Details:*
Among some other things, Qiling init method sets up the OS instance, in which `os.uid` is set acording to the value appears in `ql.root` value (`False` by default). At the end of Qiling init, the loader [an ELF loader, in this case] runs and uses `os.uid` to set the correct value in the ELF auxiliary vector. It is just after that that the user may set `ql.root = True` in their code - which is already too late.

Since there is no effective way to propagate the `root` value after loader is done, I think the way to solve this would be adding an optional `root` argument to the Qiling init method, and make the existing `root` property read-only. Since there is already an argument named `rootfs`, perhaps it would be better to name the new argument `asroot` or `isroot` to prevent mixups, but perhaps `root` would be the clearest option after all..

@xwings, what do  you think? I can create a PR for this.

xwings:
Maybe we can do it in profile? What do you think

elicn:
> Maybe we can do it in profile? What do you think

[...]

PR #946, opened 2021-10-15
Have the root property backed by OS logic

elicn:
As of today the `root` core property that is used determine whether the process is running as root, is a simple flag that is not backed by any OS logic. Since OS logic varies between different operating systems, the root property was removed from core and placed in OS. Each OS derivative may choose how to implement an "_is root_" indication and a "_set root_" behavior, if at all.

To implement the root logic on POSIX, two additional properties were added: `euid` and `egid` (effective user id and effective group id, respectively). When running a program with "sudo" the values of `euid` and `egid` are expected to change to `0` to indicate root privileges. The corresponding system calls `geteuid` and `getegid` were implemented accordingly.

That somewhat covers #931

### DIFF
--- a/qiling/os/posix/syscall/unistd.py
+++ b/qiling/os/posix/syscall/unistd.py
@@ -63,7 +63,7 @@ def ql_syscall_issetugid(ql: Qiling):
 
 
 def ql_syscall_getuid(ql: Qiling):
-    return 0
+    return ql.os.uid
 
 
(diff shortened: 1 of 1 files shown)


elicn:
> @elicn what do you think about ql.os sudo

I would prefer to go with a more general term that fits all operating systems, but POSIX use "root" and Windows uses "administrator" so I have no idea..

PR #947, opened 2021-10-15
Misc. improvements to gdb module

elicn:
Mostly refining existing code and replacing hardcoded values with actual functionality.
These changes come as a result of the recent changes in ELF loader.

### DIFF
--- a/tests/test_windows_debugger.py
+++ b/tests/test_windows_debugger.py
@@ -3,82 +3,74 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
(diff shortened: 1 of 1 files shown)


xwings:
Looks like there is a error in Windows debugger ?

elicn:
Windows debugger test should have specified `"gdb:127.0.0.1:9996"` instead of just `"127.0.0.1:9996"`.
I took the opportunity to adjust the entire test file to match the one we have for Linux, which is more extensible.

xwings:
this one seems cool!

thanks!

Issue #635, opened 2021-01-05
[loader/elf.py] Qiling'env parameter can not contain byte strings 


comewel:
**Describe the bug**
In `qiling/loader/elf.py` if Qiling's `env` parameter contains byte strings the following error is raised:
```
Traceback (most recent call last):
  File ".../Downloads/test.py", line 4, in <module>
    ql = Qiling(
  File ".../lib/python3.9/site-packages/qiling/core.py", line 211, in __init__
    self.loader.run()
  File ".../lib/python3.9/site-packages/qiling/loader/elf.py", line 124, in run
    self.load_with_ld(stack_address + stack_size, argv=self.argv, env=self.env)
  File ".../ib/python3.9/site-packages/qiling/loader/elf.py", line 330, in load_with_ld
    env_addr, new_stack = self.copy_str(new_stack, [key + '=' + value for key, value in env.items()])
  File ".../lib/python3.9/site-packages/qiling/loader/elf.py", line 330, in <listcomp>
    env_addr, new_stack = self.copy_str(new_stack, [key + '=' + value for key, value in env.items()])
TypeError: can't concat str to bytes
```
I link [here](https://github.com/qilingframework/qiling/blob/master/qiling/loader/elf.py#L330 ) the line that causes the error for ease.

**Sample Code**

```python
from qiling import *

env_dict = {b"test": b"test"}
ql = Qiling(
    ["qiling/examples/rootfs/x86_linux/bin/x86_hello"],
    "qiling/examples/rootfs/x86_linux",
    env=env_dict,
)

ql.run()
```

**Expected behavior**
The loader should manage the case in which the Qiling's `env` variable contains byte strings.
I think it could be easily achieved by making the concatenation with the `=` dependent on the type of the `env` entry.

**Additional context**
A quick (and dirty) fix would be substituting the if block [here](https://github.com/qilingframework/qiling/blob/master/qiling/loader/elf.py#L328) :
```python
...
# Set env
if len(env) != 0:
    env_addr, new_stack = self.copy_str(new_stack, [key + '=' + value for key, value in env.items()])
    elf_table += b''.join([self.ql.pack(_) for _ in env_addr])
...
```
with something like this:
```python
...
# Set env
# The check with all is performed to guarantee that each key and value have the same type
if len(env) != 0 and all([type(key) == type(value) for key, value in env.items()]):
    list_of_env = []
    for key, value in env.items():
        if isinstance(key, bytes):
[...]

PR #951, opened 2021-10-20
argv and env in elf loader should be able to accept bytes as argument

ucgJhe:
as title and should be a fix for #635


xwings:
ah, good one thanks

PR #979, opened 2021-11-03
Fix mmap logic on MAP_FIXED

bet4it:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----
> If `addr` is NULL, then the kernel chooses the (page-aligned) address at which to create the mapping; this is the most portable method of creating a new mapping. **If `addr` is not NULL, then the kernel takes it as a hint about where to place the mapping;** on Linux, the kernel will pick a nearby page boundary (but always above or equal to the value specified by `/proc/sys/vm/mmap_min_addr`) and attempt to create the mapping there. **If another mapping already exists there, the kernel picks a new address that may or may not depend on the hint.** The address of the new mapping is returned as the result of the call.

If `addr` is not NULL, we need to map on this address even if `MAP_FIXED` is not set, except this address is already mapped.

### DIFF
--- a/qiling/os/posix/syscall/mman.py
+++ b/qiling/os/posix/syscall/mman.py
@@ -82,11 +82,7 @@ def syscall_mmap_impl(ql: Qiling, addr: int, mlen: int, prot: int, flags: int, f
 
     mmap_base = addr
     need_mmap = True
-    eff_mmap_size = ((mlen + 0x1000 - 1) // 0x1000) * 0x1000
-    
-    # align eff_mmap_size to page boundary
-    aligned_address = (addr >> 12) << 12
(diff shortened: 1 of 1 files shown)


cla7aye15I4nd:
may be you can provide some test?

bet4it:
I can't find any existing tests about `mmap`😅
And `mmap` is used in almost all programs, so we can test it base on the real usage of it.
I will send a PR of a realworld program which will trigger this bug lately. ( hope in this year😅 )

xwings:
cool, waiting :)

bet4it:
That will be another PR, hope this can be merged.

xwings:
will you be able to solve the conflict?

bet4it:
@xwings Done

PR #982, opened 2021-11-06
add support to load bare-metal file

cq674350529:
For some bare-metal files, such as u-boot bin or some binaries from eCos, it's possible to run partial emulation with qiling in `shellcode` mode.

The bare-metal file is just a single and static binary with no dependency. And to analyze it, it must be located in a right base address. Currently, the shellcode mapped address has an offset to the address specified in profile. To make it easier, this change maps it into the specified address exactly.

In addition, it's possible to run it from arbitrary point, and stop at arbitrary point.

> PS: I just adapted the code for linux and freebsd.


xwings:
Hi, 

this is cool!
will you be able to add a test and example?

cq674350529:
Ok, I'll add an example later.

xwings:
the bin should be in rootfs, maybe a folder call bootrom/arm

cq674350529:
As discussed, I'll close it at the moment, and will open a new pull request later.

PR #992, opened 2021-11-10
Implement stateful SimpleStringBuffer

wtdcode:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/extensions/pipe.py
+++ b/qiling/extensions/pipe.py
@@ -13,16 +13,33 @@ class SimpleStringBuffer(TextIO):
 
     def __init__(self):
         self.buff = bytearray()
+        self.cur = 0
+
+    def lseek(self, offset: int, origin: int) -> int:
+        if origin == 0: # SEEK_SET
+            base = 0
+        elif origin == 1: # SEEK_CUR
+            base = self.cur
+        else: # SEEK_END
+            base = len(self.buff) - 1
+        
+        if base + offset >= len(self.buff):
+            self.cur = base + offset - 1
+        else:
+            self.cur = base + offset
+        
+        return self.cur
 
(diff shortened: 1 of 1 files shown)


elicn:
@wtdcode, can you give a little bit of context around that fix..?

As stated on the `SimpleStringBuffer` docstring, it is a FIFO stream; being able to seek is a _random access_ feature which breaks the FIFO property. That also requires storing the entire content of the stream throughout its lifetime (i.e. reading part of it doesn't let you get rid of that part), which becomes problematic when using it for `stdout` that may potentially yield tons of output.

Would you prefer to create a `BufferedStream` class that supports `seek`, just like `io.BufferedRandom` ?

wtdcode:
> @wtdcode, can you give a little bit of context around that fix..?
> 
> As stated on the `SimpleStringBuffer` docstring, it is a FIFO stream; being able to seek is a _random access_ feature which breaks the FIFO property. That also requires storing the entire content of the stream throughout its lifetime (i.e. reading part of it doesn't let you get rid of that part), which becomes problematic when using it for `stdout` that may potentially yield tons of output.
> 
> Would you prefer to create a `BufferedStream` class that supports `seek`, just like `io.BufferedRandom` ?

Yes, you are right. I would create another subclass.

elicn:
> Yes, you are right. I would create another subclass.

Consider adding:
```python
def seekable(self) -> bool:
    return True
```

To indicate it is a seek-able stream.

kabeor (approved):

elicn:
Not sure, but I think @wtdcode wanted to create another PR instead of this one.

wtdcode:
Yes, indeed

wtdcode:
Closed due to #1071

PR #1004, opened 2021-11-13
use self.ql.stop() instead

cq674350529:
Currently, for mcu, its start/stop routine is implemented in `arch`, not in `os`.


xwings:
Thanks and i will merge without addition approval.

Due to

i. short and simple changes.
ii. gdb related

PR #1010, opened 2021-11-16
qltool info: show extended version information

elicn:
Added a new `qltool` command that emits extended version information about Qiling and its hosting environment. Whenever a user encounters an error, pasting the output of this command as part of the bug report would help troubleshooting it.

Command: `./qltool info`
Expected output:
```
Qiling 1.4.1-dev on Linux 5.4.72-microsoft-standard-WSL2 (x86_64), with Python 3.8.10
commit: 978299a4e14977c946f3c85273bafea2dac27468 (show-version)
libs: unicorn=2.0.0, capstone=4.0.2, keystone=0.9.2
```

IMO showing extended info on-demand is a cleaner practice, comparing to a banner displayed by default when a program executes.

### DIFF
--- a/qiling/__init__.py
+++ b/qiling/__init__.py
@@ -1,2 +1,78 @@
 from .core import Qiling
 from .__version__ import __version__
+
+def __git_info():
+    """Retrieve git information of the current Qiling code base.
+    """
+
+    import os
+
+    gitdir = os.path.join(os.path.dirname(os.path.abspath(__file__)), r'..', r'.git')
+    head_path = 'HEAD'
+
+    try:
+        with open(os.path.join(gitdir, head_path), 'r') as head_file:
+            head_info = head_file.readline().rstrip()
+
+        branch_path = os.path.join(gitdir, head_info.split('ref:', 1)[-1].lstrip())
+
+        with open(os.path.join(gitdir, branch_path), 'r') as branch_file:
+            commit = branch_file.readline().rstrip()
+
+        branch = os.path.basename(branch_path)
+    except OSError:
+        branch = None
+        commit = None
+
+    return {
+        'branch' : branch or 'unknown',
+        'commit' : commit or 'unknown'
+    }
+
+def __platform_info():
+    """Retrieve hosting platform information.
+    """
+
+    import platform
+
+    return {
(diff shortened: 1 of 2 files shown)


xwings (approved):

xwings:
Cool,

I think we need to print this during after crash too?  

at verbose = debug maybe.

elicn:
> I think we need to print this during after crash too?

A crash is not necessarily a Qiling issue; on the contrary - it could be a lucky find

wtdcode:
> > I think we need to print this during after crash too?
> 
> A crash is not necessarily a Qiling issue; on the contrary - it could be a lucky find

I agree with @xwings. Show relevant info in crash long can help us reproduce it more easily and doesn't require an extra step.

qiling/__init__.py
```diff
@@ -1,2 +1,78 @@
 from .core import Qiling
-from .__version__ import __version__
\ No newline at end of file
+from .__version__ import __version__
+
+def __git_info():
```
wtdcode, line 4: I have a slight concern that variables and functions starting with double underlines are reserved for python internal use.

qiling/__init__.py
```diff
@@ -1,2 +1,78 @@
...
+    """
+
+    def __module_ver(mname: str):
+        """Query a module's version by name.
+        """
+
+        import importlib
+
+        try:
+            m = importlib.import_module(mname)
```
wtdcode, line 56: A better approach is `importlib.metadata.version` I think.

elicn: Nice, I didn't know this one..
It seems to pull the version number from `pip` rather than relying on the `__version__` convension. In case of unicorn, it shows `'2.0.0rc4'` instead of just `2.0.0`

wtdcode: Yes, that's exactly what I expect.

qiling/__init__.py
wtdcode: Okay, you are right.

elicn:
If the extended version info is emitted on error, there is no reason to integrate it into `qltool`. In that case the info acquiring logic can be placed at core utils. I'm closing this PR and create another one sometime.

PR #1012, opened 2021-11-17
Minor improvements to memory module

elicn:
Highlights:
- Updated stale annotations, comments and debug messages per mmio mapping functionality
- Enhanced and simplified `map_mmio`
- Excluded mmio ranges from search to avoid potential read side effects
- `pagesize` is now controlable (instead of fixed `0x1000`) to make the memory module easier to adapt to obscure archs and OS in the future

### DIFF
--- a/qiling/os/memory.py
+++ b/qiling/os/memory.py
@@ -204,18 +204,18 @@ class QlMemoryManager:
         """
 
         for lbound, ubound, perms, label, data in mem_dict['ram']:
-            self.ql.log.debug(f'To restore mapping: {lbound:#08x} {ubound:#08x} {label}')
+            self.ql.log.debug(f'restoring memory range: {lbound:#08x} {ubound:#08x} {label}')
 
             size = ubound - lbound
             if not self.is_mapped(lbound, size):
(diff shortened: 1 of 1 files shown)


qiling/os/memory.py
```diff
@@ -37,10 +40,15 @@ def __init__(self, ql: Qiling):
...
 
-        #self.read_ptr = read_ptr
         self.max_addr = max_addr
         self.max_mem_addr = max_addr
 
+        # memory page size
+        self.pagesize = 0x1000
+
+        # make sure pagesize is a power of 2
+        assert self.pagesize & (self.pagesize - 1) == 0, 'pagesize has to be a power of 2'
```
cla7aye15I4nd, line 50: Maybe assert in @pagesize.setter is better choice?

elicn: As a first step `pagesize` can be only configured statically. Other than that, the plan is to have `pagesize` set only at instance initialization (since allowing `pagesize` to change during runtime makes little sense to me), so I don't think there will be a setter for it.

cla7aye15I4nd (approved):

xwings (approved):

PR #1013, opened 2021-11-18
Abstract calls to native functions

elicn:
Users may now stage and call native functions without having to do manual frame setup.
This code abstracts away the need to create a new frame, write the arguments, set the return address and pc.

Example:
```python
from qiling.os.const import *
# ...
func_entry = 0xdeadc0de

ql.os.fcall.call_native(func_entry, (
    (POINTER, x),
    (POINTER, y),
    (POINTER, z)
), return_trap)
```

Notes:
- Only supports plain arguments types (i.e. integer values tagged with a `POINTER` type); will support complex types in the future
- Implemented needed methods for Intel and ARM architectures, not yet MIPS and RISC
- The user will have to hook the return address to do a proper cleanup, if needed

### DIFF
--- a/qiling/cc/__init__.py
+++ b/qiling/cc/__init__.py
@@ -70,6 +70,17 @@ class QlCC:
 
 		raise NotImplementedError
 
+	def reserve(self, nslots: int) -> None:
+		"""Reserve slots for function arguments.
+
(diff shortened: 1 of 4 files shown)


xwings (approved):
ah, this is nice

PR #1020, opened 2021-11-23
Work around UC2 / CS4 x86 registers definitions incompatibility

elicn:
As of Unicron 2.0.0rc4, Unicron and Capstone Intel registers definitions are no longer aligned and cannot be used interchangebly. This fix works around the definitions incompatibility by mapping Capstone x86 registers definitions to Unicorn x86 registers definitions.

See: https://github.com/unicorn-engine/unicorn/issues/1492

### DIFF
--- a/qiling/extensions/trace.py
+++ b/qiling/extensions/trace.py
@@ -14,6 +14,29 @@ from qiling import Qiling
 
 TraceRecord = Tuple[CsInsn, Iterable[Tuple[int, int]]]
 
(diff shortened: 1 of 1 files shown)


xwings:
This is interesting, @kabeor  and @aquynh is working on this and it should be fix in the latest Capstone. 

The question is do we want to merge now to wait for their fix.

aquynh:
It is always safe to have mapping, so this should be merged

PR #1021, opened 2021-11-23
Minor tweaking to core hooks

elicn:
Changes:
- Typing annotations
- Python styling corrections
- Code cleanup and simplification

### DIFF
--- a/qiling/core_hooks.py
+++ b/qiling/core_hooks.py
@@ -107,7 +107,7 @@ class QlCoreHooks:
                     if type(ret) is int and ret & QL_HOOK_BLOCK:
                         break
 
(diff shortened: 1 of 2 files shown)


qiling/core_hooks.py
```diff
@@ -46,15 +46,17 @@ def _hook_intr_cb(self, uc: Uc, intno: int, pack_data) -> None:
...
+                if hook.check(intno):
                     handled = True
                     ret = hook.call(ql, intno)
 
                     if type(ret) is int and ret & QL_HOOK_BLOCK:
                         break
 
         if not handled:
-            ql.log.warning("[+] Unhandled Interupt: %i" % (intno))
+            raise QlErrorCoreHook("_hook_intr_cb : not handled")
```
wtdcode, line 59: Why raise an exception here? It's a break change.

elicn: Hmm.. I don't remember myself changing functionality here, and indeed it turns out that this line changed about two weeks ago [changed from an exception to a log message]. See here: https://github.com/qilingframework/qiling/blob/94bf7a3bc4e3ea0cffaaa52dbc477c11030f631b/qiling/core_hooks.py#L70

Regardless of the original functionality [that raised an exception until two weeks ago]; what do you think would make more sense? I don't mind either way..

kabeor (approved):

xwings (approved):

Issue #1045, opened 2021-12-16
run ./addressNat_overflow.sh to crash to generate snapshot.bin


KA2010:
**When I am trying to run addressNat_overflow.sh in order to crash to generate snapshot.bin, I got  HTTP/1.1 404 Not Found** 
How i can solve this issue?
>`/qiling/examples/fuzzing/tenda_ac15$` ./addressNat_overflow.sh`
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 80 (#0)
> POST /goform/addressNat HTTP/1.1
> Host: localhost
> User-Agent: curl/7.68.0
> Accept: */*
> Referer: http://localhost:80/samba.html
> Cookie: password=1234
> X-Requested-With: XMLHttpRequest
> Content-Type:application/x-www-form-urlencoded
> Content-Length: 25
>
* upload completely sent off: 25 out of 25 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Date: Thu, 16 Dec 2021 13:34:10 GMT
< Server: Apache/2.4.41 (Ubuntu)
< Content-Length: 271
< Content-Type: text/html; charset=iso-8859-1
<
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>404 Not Found</title>
</head><body>
<h1>Not Found</h1>
<p>The requested URL was not found on this server.</p>
<hr>
<address>Apache/2.4.41 (Ubuntu) Server at localhost Port 80</address>
</body></html>
* Connection #0 to host localhost left intact

KA2010:
Note that if I used port 8080 it gave me connection refused.
>/qiling/examples/fuzzing/tenda_ac15$ ./addressNat_overflow.sh
*   Trying 127.0.0.1:8080...
* TCP_NODELAY set
* connect to 127.0.0.1 port 8080 failed: Connection refused
* Failed to connect to localhost port 8080: Connection refused
* Closing connection 0
curl: (7) Failed to connect to localhost port 8080: Connection refused

cq674350529:
It seemed that the httpd service wasn't up, and https://github.com/qilingframework/qiling/pull/1046 solve this issue.

xwings:
which version of Qiling are you using ?

KA2010:
> which version of Qiling are you using ?
1.4.0

KA2010:
?

cq674350529:
[...]

PR #1047, opened 2021-12-25
fix map prot issue

cq674350529:
try to fix https://github.com/qilingframework/qiling/issues/1045

Without this fix, running the tenda example will throw `UC_ERR_FETCH_UNMAPPED` under the current latest commit.
```
python3 saver_tendaac15_httpd.py
```


cq674350529:
Update: when update the `unicorn` to `2.0.0rc5`, the error is gone.

So close it.

PR #1049, opened 2021-12-27
Fail gracefully when the IP is inside unmapped memory (possible buffer overflow)

### LINKED ISSUE #988: How can I have buffer overflows (Invalid memory fetch) fail gracefully? I already tried hooking.
I wanted to test using qiling for exploitation development. So I'm passing a cyclic buffer to the program to generate a buffer overflow. A core file gets generated upon the crash, but I can't seem to figure out how to make it fail gracefully so that I can analyze the register states (core file) after the program crashes:

```
from cyclic import cyclic, cyclic_find

# Qiling
# Full system emulation
from qiling import *


buffer = cyclic(50, n=4)
print(f'generated cyclic buffer: {buffer}')

# pass it to program for an overflow
BIN_PATH = ["exploit_this", buffer]
ENV = "../examples/rootfs/x86_linux"

ql = Qiling(BIN_PATH, ENV)

def dissassem(ql, address, size):
    buf = ql.mem.read(address, size)
    for i in md.disasm(buf, address):
        print("0x%x:\t%s\t%s" %(i.address, i.mnemonic, i.op_str))

def dump_regs(ql, int_code):
    print(f'registers:')
    for reg_key in ql.reg.register_mapping:
        print(f'{reg_key.upper()}: {ql.reg.read(reg_key.upper())}')

# ql.hook_code(dissassem)
ql.hook_intr(dump_regs)
ql.hook_mem_unmapped(dump_regs)
ql.hook_mem_read_invalid(dump_regs)
ql.hook_mem_write_invalid(dump_regs)
ql.hook_mem_fetch_invalid(dump_regs)
ql.hook_mem_invalid(dump_regs)
ql.run()
this_reg_was_tainted = ql.reg.read("EAX")
print(f'EAX: {this_reg_was_tainted}')
offset = cyclic_find(this_reg_was_tainted)
# now we know where to insert a payload
```

```
[+] Buffer: aaaabaaacaaadaaaeaaafaaagaaahaaaiaaajaaakaaalaaama
[=]	write(fd = 0x1, buf = 0x804d160, count = 0x3f) = 0x3f
registers:
AH: 0
AL: 63
CH: 209
CL: 96
DH: 0
DL: 63
BH: 0
BL: 1
AX: 63
CX: 53600
DX: 63
BX: 1
SP: 51100
BP: 32128
SI: 53600
DI: 63
IP: 44114
EAX: 63
ECX: 134533472
EDX: 63
EBX: 1
ESP: 2146682780
EBP: 2003402112
ESI: 134533472
EDI: 63
EIP: 75213906
CR0: 17
CR1: 0
CR2: 0
CR3: 0
CR4: 0
CR8: 0
ST0: 0
ST1: 0
ST2: 0
ST3: 0
ST4: 0
ST5: 0
ST6: 0
ST7: 0
EF: 68
CS: 27
SS: 40
DS: 40
ES: 40
FS: 0
GS: 99
[x]	CPU Context:
[x]	ah	: 0x0
[x]	al	: 0x3f
[x]	ch	: 0x0
[x]	cl	: 0x0
[x]	dh	: 0x88
[x]	dl	: 0x90
[x]	bh	: 0x61
[x]	bl	: 0x61
[x]	ax	: 0x3f
[x]	cx	: 0x0
[x]	dx	: 0x8890
[x]	bx	: 0x6161
[x]	sp	: 0xce20
[x]	bp	: 0x6161
[x]	si	: 0x7000
[x]	di	: 0x0
[x]	ip	: 0x6161
[x]	eax	: 0x3f
[x]	ecx	: 0x0
[x]	edx	: 0x77698890
[x]	ebx	: 0x61696161
[x]	esp	: 0x7ff3ce20
[x]	ebp	: 0x616a6161
[x]	esi	: 0x77697000
[x]	edi	: 0x0
[x]	eip	: 0x616b6161
[x]	cr0	: 0x11
[x]	cr1	: 0x0
[x]	cr2	: 0x0
[x]	cr3	: 0x0
[x]	cr4	: 0x0
[x]	cr8	: 0x0
[x]	st0	: 0x0
[x]	st1	: 0x0
[x]	st2	: 0x0
[x]	st3	: 0x0
[x]	st4	: 0x0
[x]	st5	: 0x0
[x]	st6	: 0x0
[x]	st7	: 0x0
[x]	ef	: 0x4
[x]	cs	: 0x1b
[x]	ss	: 0x28
[x]	ds	: 0x28
[x]	es	: 0x28
[x]	fs	: 0x0
[x]	gs	: 0x63
[x]	PC = 0x616b6161 (unreachable)

[=]	Memory map:
[=]	Start      End        Perm    Label          Image
[=]	00030000 - 00031000   rwx     [GDT]          exploit_this
[=]	047ba000 - 047e2000   rwx     /home/student/synthesis/examples/rootfs/x86_linux/lib/ld-linux.so.2   exploit_this
[=]	08048000 - 08049000   r-x     exploit_this   exploit_this
[=]	08049000 - 0804b000   rw-     exploit_this   exploit_this
[=]	0804b000 - 0804d000   rwx     [hook_mem]     
[=]	0804d000 - 0806e000   rwx     [brk]          
[=]	0806e000 - 0806f000   rwx     [brk]          
[=]	774bf000 - 7769b000   rwx     [syscall_mmap2]   
[=]	774bf000 - 7769b000   rwx     [mmap2] /home/student/synthesis/reverse/../examples/rootfs/x86_linux/lib/libc.so.6   
[=]	77695000 - 77698000   rwx     [mmap2] /home/student/synthesis/reverse/../examples/rootfs/x86_linux/lib/libc.so.6   
[=]	7769b000 - 7769d000   rwx     [syscall_mmap2]   
[=]	7ff0d000 - 7ff3d000   rwx     [stack]        
  File "reverse.py", line 34, in <module>
    ql.run()
  File "/usr/local/lib/python3.6/dist-packages/qiling/core.py", line 707, in run
    self.os.run()
  File "/usr/local/lib/python3.6/dist-packages/qiling/os/linux/linux.py", line 139, in run
    self.ql.emu_start(self.ql.loader.elf_entry, self.exit_point, self.ql.timeout, self.ql.count)
  File "/usr/local/lib/python3.6/dist-packages/qiling/core.py", line 867, in emu_start
    self.uc.emu_start(begin, end, timeout, count)
  File "/usr/local/lib/python3.6/dist-packages/unicorn/unicorn.py", line 465, in emu_start
    raise UcError(status)
unicorn.unicorn.UcError: Invalid memory fetch (UC_ERR_FETCH_UNMAPPED)
```

You can see in the core file dumped during the crash, the EIP value 0x616b6161 is my buffer overflow, but in the last hooked handler the EIP isnt set to that, so I never have access to a crashed instance of the registers.

Is there something else I need to do to prevent the python code from throwing an error when the emulated program crashes after an invalid memory fetch? I thought just hooking the invalid fetch would work but that hook doesn't even seem to trigger.

TheZ3ro:

- [X] This PR introduces a new function/api for Qiling Framework.

### Coding convention?

### Extra tests?

- [X] Tests will be added after some discussion and review.

### Changelog?

### Target branch?

### One last thing

-----

This PR add an option to stop emulation whenever the next instruction is fetched from unmapped memory.
This usually happen with a buffer overflow and it's useful to debug what is the state of the registers and the memory.
Fix #988 

If this feature is useful I can add a test case for it.


qiling/core.py
```diff
@@ -685,6 +687,22 @@ def _exit_trap(ql):
...
 
+        # Stop when running to an unmapped IP
+        if self.stop_options.unmapped_ip:
+            def _check_ip(ql, access, address, size, value):
+                # make sure we got here for the right reason
+                if access == uc.UC_MEM_FETCH_UNMAPPED:
+                    self.log.info(f'Process returned from diverted execution flow to 0x{address:x}')
+                    # to gracefully recover from memory read/write error, you have to memory map the invalid page before you return true
+                    # https://github.com/unicorn-engine/unicorn/issues/1484#issuecomment-970727708
+                    self.mem.map(address//0x1000*0x1000, 0x1000, info='[BufferOverflow guard]')
```
elicn, line 698: Suggesting to use `mem.align` and `mem.pagesize` to maintain compatibility.

TheZ3ro: I was trying using `ql.mem.align` but it end up allocating memory after the address pointed in the instruction pointer:

```
[=]     Process returned from diverted execution flow to 0x61616168
[=]     Allocated aligned memory at 0x61617000
```

```python
self.log.info(f'Process returned from diverted execution flow to 0x{address:x}')
addr = self.mem.align(address, 0x1000)
self.log.info(f'Allocated aligned memory at 0x{addr:x}')
self.mem.map(addr, 0x1000, info='[BufferOverflow guard]')
```

And thus Unicorn complains

elicn: You're right, `mem.align` rounds the address up.. What about this one:
`self.mem.map(address & ~(self.mem.pagesize - 1), self.mem.pagesize, info='[guard]')`

TheZ3ro: maybe we can add an optional parameter to `mem.align` in order to specify if the address we are passing should be inside the aligned memory and thus doing floor rounding. This could be used often.

I like your proposal for the rounding

elicn: I don't think letting an optional parameter decide the rounding direction is a good idea.. an optional parameter may be used to slightly change the behavior of the function, not change it entirely.

Normally an `align` method would mean aligning the value by roudning it **down**, so I think it may be best to change the implementation in the `mem` module to have it rounding down one would expect. It looks like there are only a few usages of the `mem.align` currently, so it will be easy to revisit its usages and provide them with an alternate method of rounding up.

If it's OK with you, let's leave the `mem.align` method intact for now, and once I fix its functionality (to round down instead of up), we'll use it as intended. What do you think?

TheZ3ro: Make sense, I will exclude the `mem.align` commit from the PR

xwings:
Hi,

there is a conflict. can you please fix it ?

PR #1067, opened 2022-01-17
qiling-next

elicn:
#### Constructor arguments
| Removed | Alternative
| :--- | :---
`stdin`  | `ql.os.stdin`
`stdout` | `ql.os.stdout`
`stderr` | `ql.os.stderr`

| Added | Notes
| :--- | :---
`**kwargs` | Allow specifying misc switches (e.g. `thumb=True`)

#### Methods and properties
| Removed | Alternative
| :--- | :---
`ql.assembler` | `ql.arch.assembler`
`ql.disassembler` | `ql.arch.disassembler`
`ql.create_assembler` | `ql.arch.assembler`
`ql.create_disassembler` | `ql.arch.disassembler`
`ql.reg` | `ql.arch.regs`
`ql.archbit` | `ql.arch.bits`
`ql.archendian` | `ql.arch.endian`
`ql.pointersize` | `ql.arch.pointersize`
`ql.platform_os` | `ql.host.os`
`ql.platform_arch` | `ql.host.arch`

#### Methods and properties
| Removed | Alternative
| :--- | :---
`ql.arch.init_uc` | `ql.arch.uc`
`ql.arch.set_pc` | `ql.arch.regs.arch_pc`
`ql.arch.set_sp` | `ql.arch.regs.arch_sp`
`ql.arch.get_sp` | `ql.arch.regs.arch_sp`
`ql.arch.create_disassembler` | `ql.arch.disassembler`
`ql.arch.create_assembler` | `ql.arch.assembler`
`ql.arch.get_pc` | `ql.arch.effective_pc` (`QlArchARM` derivatives only)

#### Other changes
- Thumb mode is configurable on init and properly reflected in runtime (ARM)
- Endianess is configurable on init and properly reflected in runtime (ARM, MIPS)

#### Methods and properties
| Removed | Alternative
| :--- | :---
`ql.arcg.regs.ql` | N/A
`ql.arcg.regs.expand_mapping` | N/A
`ql.arcg.regs.msr` | `ql.arch.msr` (`QlArchIntel` derivatives only)
`ql.arcg.regs.bit` | `ql.arch.reg_bits` (`QlArchX8664` only)
`ql.arcg.regs.register_sp` | N/A
`ql.arcg.regs.register_pc` | N/A


xwings:
If we change to ql.arch.regs , mem need to mems too ?

elicn:
The question is not clear to me.
Are you taking about a specific change? What the memory module has to do with `ql.arch.regs`..?

xwings:
ah sorry, lets me elaborate more.

it was ql.reg (without s) and if we change to ql.arch.regs (with s). all other components need to go with s ? like ql.mems (with s). or there is a standard where which is is singular and plural?

elicn:
I don't think there's a namig convention here.. I guess it is because "regs" is an abbreviation for "registers" which is in plural.
_memory_, _architecture_ and other names are not in plural, so they are not abbreviated as such.

xwings:
> I don't think there's a namig convention here.. I guess it is because "regs" is an abbreviation for "registers" which is in plural. _memory_, _architecture_ and other names are not in plural, so they are not abbreviated as such.

cool, that make sense.

qiling/arch/arch.py
```diff
@@ -3,30 +3,44 @@
...
 
-        self._disasm: Optional[Cs] = None
-        self._asm: Optional[Ks] = None
+    @property
+    @abstractmethod
+    def uc(self) -> Uc:
+        """Get unicorn instance bound to arch.
+        """
+
+        pass
```
wtdcode, line 31: I suggest raising a NotImplementedError here. Same below.

qiling/arch/arch.py
```diff
@@ -37,10 +51,10 @@ def stack_push(self, value: int) -> int:
         Returns: the top of stack after pushing the value
         """
 
-        self.ql.reg.arch_sp -= self.ql.pointersize
-        self.ql.mem.write(self.ql.reg.arch_sp, self.ql.pack(value))
+        self.regs.arch_sp -= self.pointersize
```
wtdcode, line 54: Is it possible that some arch has a stack implementation that grows up? Not urgent for now, just some random thoughts, maybe we can leave a TODO here?

qiling/arch/arm.py
```diff
@@ -3,107 +3,111 @@
...
         self.arm_get_tls_addr = 0xFFFF0FE0
 
-    # get initialized unicorn engine
-    def get_init_uc(self) -> Uc:
-        if self.ql.archendian == QL_ENDIAN.EB:
-            mode = UC_MODE_ARM + UC_MODE_BIG_ENDIAN
-
-        elif self.ql.archtype == QL_ARCH.ARM_THUMB:
-            mode = UC_MODE_THUMB
+    @cached_property
```
wtdcode, line 29: `cached_property` requires python >= 3.8. Same below.

qiling/core.py
```diff
@@ -48,15 +48,11 @@ def __init__(
             filter = None,
             stop_on_stackpointer = False,
             stop_on_exit_trap = False,
-            stdin=None,
-            stdout=None,
-            stderr=None,
+            **kwargs
```
wtdcode, line 51: Personally, I prefer `stdin`, `stdout`, `stderr` to be passed as arguments, since it looks more intuitive to me.

qiling/arch/arch.py
```diff
@@ -3,30 +3,44 @@
...
+        """
+
+        pass
 
-    # ql.init_Uc - initialized unicorn engine
     @property
-    def init_uc(self) -> Uc:
-        return self.get_init_uc()
+    @abstractmethod
+    def regs(self) -> QlRegisterManager:
```
wtdcode, line 35: I suggest `ql.arch.pc` and `ql.arch.sp` to replace `ql.arch.regs.arch_pc` and `ql.arch.regs.arch_sp` because the latter is too verbose. Maybe an alias?

qiling/arch/arch.py
elicn: The `@abstractmethod` decorator does that.

qiling/arch/arch.py
elicn: I am really struggling with these stack accessors, as they clearly not belong to `QlArch`, but I don't know exactly where to put them. Anyway, I agree that this behavior should be overload-able.

qiling/arch/arm.py
elicn: Indeed. As far I remember Python 3.8 is the minimal version required for Qiling to work (based on our tests suite).

qiling/core.py
elicn: That is debateable, I suppose.
Qiling constructor has too many arguments, of which some are quite obscure (e.g. log filtering, etc.). My goal is to simplify the constructor arguments as much as I can, and let the user assign directly whatever properties they may like to use (like the OS standard streams).

In the same manner, the `**kwargs` is there to "throw aside" all the obscure switches and not let them pollute the constructor... though I agree it might not be a super-elegant way to do that.

qiling/arch/arch.py
elicn: Possible, but depends on whether we want to allow aliases. Aliases may provide more flexibility, but also come with code duplication and additional dependecies. Even if this one may be simple to implement, it counters the _simplicity_ guideline, IMHO, as it provides more than one way to do something.

qiling/arch/arch.py
wtdcode: I agree. I suppose it should go to `ql.mem`?

qiling/arch/arm.py
wtdcode: Yes, that's true. But there is still a small number of users (more than 5%) using 3.6 or 3.7 (https://pypistats.org/packages/qiling). I suppose it's the only feature that needs python>=3.8 so there is not too much work to do for compatibility with older python. [This answer](https://stackoverflow.com/questions/4037481/caching-class-attributes-in-python) may provide some insight about the compatible code.

qiling/arch/arch.py
wtdcode: Yes, alias is not a good way. Maybe we can eliminate `ql.arch.regs.arch_pc` and `ql.arch.regs.arch_sp` totally?

qiling/core.py
wtdcode: Fine, make sense.

qiling/arch/arch.py
xwings: i think in x86 16 bit there is actually a register call sp. that is the reason why we use arch_sp before this. (i think, i dont really remember)

qiling/arch/arch.py
elicn: This is one of the options.. but it makes `mem` depend on `arch`, which does not sit well with me.
Another option would be `QlOs` that may maintain a stack object providing these methods. Although stack does not belong to OS but to a process, we don't have an object for process - and the closest one we have is `QlOS`.
The last reasonable option would be `Qiling`, which is not a good idea for several reasons..

I did say it's a struggle (:

qiling/arch/arm.py
elicn: Thanks for the stats..! I didn't know we have these.
The stats show there is a non-negligble amount of users that still use Python 3.6 so I agree we should keep supporting it. However, as of Dec 2021 Python 3.6 has reached end-of-life [see [here](https://www.python.org/dev/peps/pep-0494/#lifespan)] and will no longer receive bugfixes or security updates.
Do we really want to keep supporting it after all...?

qiling/arch/arch.py
elicn: ARM, MIPS and RISC-V have registers named `pc` and `sp` (:
But if we put them on `QlArch`, there will be both `ql.arch.pc` and `ql.arch.regs.pc`.
No question we need a generic method to access the architectural IP/ PC, the question it makes most sense to put that accessor and how it will be called.

aquynh:
If you keep pushing the Python version, users need to keep updating their
own code just because they use qiling, which makes no sense.

elicn:
That is a fair point, however keep in mind that this re-design already breaks any existing code out there that does more than simple hooking. That could be a good opportunity to officially bump the version up.

Just a thought.

aquynh:
So you force users to update their Python just for some minor "features" in
the code that nobody is even aware of. Why is that worth it?

qiling/arch/arm.py
kabeor: Actually there is simple solution for 3.6 or 3.7. 
Plz check https://github.com/qilingframework/qiling/blob/750988e8eed42215bce5ee9531bc098002ebb2c8/qiling/arch/evm/vm/computation.py#L20-L24
I used to deal with this.

kabeor: Actually there is simple solution for 3.6 or 3.7. 
Plz check https://github.com/qilingframework/qiling/blob/750988e8eed42215bce5ee9531bc098002ebb2c8/qiling/arch/evm/vm/computation.py#L20-L24
I used to deal with this.

elicn: Sure, threre is always a way to work around these things.
It just started a discussion about whether that would make a good opportunity to bump up the requires Python version, along with its new features.

kabeor: Yes I know, this is something worth thinking about.

qiling/os/posix/syscall/unistd.py
```diff
@@ -432,8 +432,18 @@ def __read_str_array(addr: int) -> Iterator[str]:
     if ql.code:
         return
 
-    ql._uc = ql.arch.init_uc
-    QlCoreHooks.__init__(ql, ql._uc)
+    # recreate cached uc
+    del ql.arch.uc
```
wtdcode, line 436: Why uc is re-generated here?

elicn: The original code did that on `execve` not entirely sure why (perhaps that was a quick way to discard everything and start fresh..?). I just followed it.

wtdcode: But the two code behaves differently if `ql._uc` and `ql.arch.init_ic` points to the same object.

elicn: I'm not sure that I follow you here.. The old code regenerated a new Uc instance by calling `init_uc` (which actually was used only here and on Qiling init), it wasn't a cached property.
The new code has Uc cached, and the `del` statement is how you force it to be re-created next time you retrieve it.
Both assign Qiling with a new Uc instance, it is just done a bit differently.

PR #1073, opened 2022-01-20
ql.afl_fuzz API

wtdcode:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR adds a new API `ql.afl_fuzz`, which basically is a wrapper for unicornafl to reduce some duplicate code in fuzzers.

Note: Since unicornafl now is distributed via pypi regularly, I add the dependency in setup.py.

### DIFF
--- a/examples/fuzzing/dlink_dir815/dir815_mips32el_linux.py
+++ b/examples/fuzzing/dlink_dir815/dir815_mips32el_linux.py
@@ -7,12 +7,6 @@
 
 import os,sys
 
-# This is new. Instead of unicorn, we import unicornafl. It's the same Uc with some new `afl_` functions
-import unicornafl
-
-# Make sure Qiling uses our patched unicorn instead of it's own, second so without instrumentation!
-unicornafl.monkeypatch()
-
 sys.path.append("../../..")
 from qiling import *
 from qiling.const import QL_VERBOSE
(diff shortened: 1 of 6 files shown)


wtdcode:
> I like the idea of masking off the fuzzing setup from the user and provide them with a cleaner interface. However, it looks like the fuzzer and its setup do not depend much on Qiling and its state. What do you think about moving that out from core into a separated object (e.g. something like `ql.fuzzer` ) or make it an extension?
> 
> I think this cold be expanded even more with useful fuzzing API, but I don't want them to "mess" Qiling core.

`ql.afl.fuzz` seems fine since I'm also trying to integrate more fuzzers.

elicn:
Cool! Didn't know about the other fuzzers.
In that case, I think it would be better to refer to the fuzzer object genericly as `ql.fuzzer` and init it with the instance you want, e.g. `QlFuzzerAFL`. All fuzzer classes are expected to implement the same interface to let the user switch fuzzers seamlessly.

Would you like to design that interface?

wtdcode:
> Cool! Didn't know about the other fuzzers. In that case, I think it would be better to refer to the fuzzer object genericly as `ql.fuzzer` and init it with the instance you want, e.g. `QlFuzzerAFL`. All fuzzer classes are expected to implement the same interface to let the user switch fuzzers seamlessly.
> 
> Would you like to design that interface?

Fuzzers work differently so I'm afraid it makes no sense and almost impossible to "let the user switch fuzzers seamlessly".

elicn:
I didn't mean switching fuzzers on the fly, which indeed does not make sense.
My idea is to have the same Qiling program be able to pick either of the fuzzers without having to change the code, since they are all complying to the same API.

wtdcode:
> I didn't mean switching fuzzers on the fly, which indeed does not make sense. My idea is to have the same Qiling program be able to pick either of the fuzzers without having to change the code, since they are all complying to the same API.

That's where it's almost impossible. It's almost impossible to pick a fuzzers without changing any code (in fact, usually it means a big rewrite).

wtdcode:
To be exact, `aflpp` works as grandparent-parent-child, 3 level model while qiling program has full control of the `parent` (started by `uc_afl_fuzz`) and `child` (the actual fuzz-ed code).

However, `libfuzzer` works total differently. To a certain extent, you only have full control of the `child`, but not `parent`, which means the same script won't work anyway. And there is much more difference between fuzzers users __should__ be aware.

I agree there is some possibility to achieve almost the same API between fuzzers but that really doesn't make too much sense. I prefer something like `ql.afl.fuzz` and `ql.libfuzzer.fuzz` (I don't think this API can be implemented, though, due to the way libfuzzer works).

elicn:
Thanks for clarifying this.
In that case I believe that making them Qiling extensions would be a cleaner approach then having them an integral part of Qiling core.

wtdcode:
Closed due to #1075

PR #1071, opened 2022-01-19
Implement buffered stream

wtdcode:
Add close() and closed() to SimpleStreamBase

Set seekable to false for SimpleStringBuffer

### Which kind of PR do you create?

- [x] This PR only contains minor fixes.
- [ ] This PR contains major feature update.
- [ ] This PR introduces a new function/api for Qiling Framework.

### Coding convention?

- [x] The new code conforms to Qiling Framework naming convention.
- [x] The imports are arranged properly.
- [x] Essential comments are added.
- [x] The reference of the new code is pointed out.

### Extra tests?

- [x] No extra tests are needed for this PR.
- [ ] I have added enough tests for this PR.
- [ ] Tests will be added after some discussion and review.

### Changelog?

- [x] This PR doesn't need to update Changelog.
- [ ] Changelog will be updated after some proper review.
- [ ] Changelog has been updated in my PR.

### Target branch?

- [x] The target branch is dev branch.

### One last thing

- [x] I have read the [contribution guide](https://docs.qiling.io/en/latest/contribution/)

-----

This implements `SimpleBufferedStream` and adds some methods for `SimpleStreamBase` and `SimpleStringBuffer`.

PR #1075, opened 2022-01-23
Fuzzing evovled

wtdcode:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR takes over #1073 and #1071.

- Update unicornafl support in extensions.
- Integrate libfuzzer by fuzzercorn.
- Fix pipe seek implementation.
- Add a buffered stream implementation.


wtdcode:
Once being merged, the docs & changelog should also be updated.

qiling/extensions/pipe.py
```diff
@@ -48,11 +48,28 @@ def writable(self) -> bool:
     def readable(self) -> bool:
         return True
 
+    def seek(self, offset: int, origin: int) -> int:
+        # Imitate os.lseek
+        raise OSError("Illega Seek")
```
elicn, line 53: `seek` should not be implemented on a non-seekable streams.
Suggesting to remove this and let the caller check for `seekable`. If not seekable, the syscall should return `-ESPIPE` (as opposed to other errors).

qiling/extensions/pipe.py
```diff
@@ -48,11 +48,28 @@ def writable(self) -> bool:
...
+
+    def lseek(self, offset: int, origin: int) -> int:
```
elicn, line 55: `lseek` is not part of the `TextIO` interface, suggesting to remove this and let the caller use `seek` (after checking that it is indeed `seekable`)

qiling/extensions/pipe.py
```diff
@@ -84,3 +101,82 @@ def flush(self) -> None:
...
+
+class SimpleBufferedStream(TextIO):
+    """Simple buffered IO.
+    """
+
+    def __init__(self):
+        self.buff = bytearray()
+        self.cur = 0
+
+    def lseek(self, offset: int, origin: int) -> int:
```
elicn, line 113: Same here, `lseek` is not part of the interface; could use `seek` instead.

qiling/extensions/pipe.py
```diff
@@ -84,3 +101,82 @@ def flush(self) -> None:
...
+        while not (ret.endswith(b'\n') or len(ret) == limit):
+            ret.extend(self.read(1))
+
+        return bytes(ret)
+
+    def write(self, s: bytes) -> int:
+        if self.cur == 0:
+            self.buff.extend(s)
+        else:
+            self.buff[:self.cur].extend(s)
```
elicn, line 160: This statement has no effect.

qiling/extensions/pipe.py
wtdcode: > `seek` should not be implemented on a non-seekable streams. Suggesting to remove this and let the caller check for `seekable`. If not seekable, the syscall should return `-ESPIPE` (as opposed to other errors).

Then how about other methods like `fileno`, `fstat` and `fnctl`? My idea is that, this imitates how `sys.stdin` really works and a developer doesn't need to care about whether it is `sys.stdin` or some other objects we mocked. And "seek should not be implemented on a non-seekable streams." doesn't make sense to me. Even though `sys.stdin` is not seekable, it still can be seek-ed.

qiling/extensions/pipe.py
```diff
@@ -84,3 +101,82 @@ def flush(self) -> None:
...
+        return True
+
+    def readable(self) -> bool:
+        return True
+
+    def seekable(self) -> bool:
+        return True
+    
+    # This method simply extends internal buffer with self.cur untouched.
+    def extend(self, s: bytes) -> int:
```
elicn, line 179: This is also not part of the `TextIO` interface, and it is unclear where it is actually being used.

wtdcode: For example, when you would like to append some bytes in the script, using `write` would change the current position which is easy to cause errors hard to notice.

qiling/extensions/pipe.py
wtdcode: There are too many things that have to be checked if we do this in this way, for instance, `readable`, `writable` etc. My idea is just to keep the same behavior as `sys.stdin`.

qiling/extensions/pipe.py
wtdcode: Fixed.

qiling/extensions/pipe.py
elicn: Not sure that I follow on this.. The idea is to provide a clean Pythonic interface to the stream.
IMOH, adding all those POSIX / UNIX aliases would mess it up.

qiling/extensions/pipe.py
wtdcode: Fine, I agree we can change the implementation of lseek syscall to use seek instead,

qiling/extensions/pipe.py
wtdcode: > Not sure that I follow on this.. The idea is to provide a clean Pythonic interface to the stream. IMOH, adding all those POSIX / UNIX aliases would mess it up.

Yes, `lseek` can be removed, but I insist raising an exception instead of checking `seekable` before calling `seek`.

qiling/extensions/pipe.py
elicn: I agree: each syscall should check that it applies to the target fd before it does anything.
`write` should check first fd is writeable, `read` should check first fd is readable, and so on.. there are specific errors just for that.

That implementation would be much closer to the spec.

qiling/extensions/pipe.py
elicn: I am not sure that justifies breaking the interface. The idea behind the `Simple` name prefix is to follow the expected interface in the simplest manner. Any addition that breaks this should be implemented elsewhere; those little additions and hacks accumulate quickly to a `DefinitleyNotSimple` class...

wtdcode: > I am not sure that justifies breaking the interface. The idea behind the `Simple` name prefix is to follow the expected interface in the simplest manner. Any addition that breaks this should be implemented elsewhere; those little additions and hacks accumulate quickly to a `DefinitleyNotSimple` class...

Okay, make sense.

wtdcode: Removed.

qiling/extensions/pipe.py
wtdcode: Removed.

qiling/extensions/pipe.py
wtdcode: Removed.

qiling/extensions/pipe.py
wtdcode: Fine, now we don't need to check since an exception will be raised like other streams.

elicn (approved):

xwings (approved):

PR #1084, opened 2022-01-27
Add close all logs option for verbose

kabeor:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----


qiling/core.py
```diff
@@ -547,6 +551,8 @@ def verbose(self):
     @verbose.setter
     def verbose(self, v):
         self._verbose = v
+        if self._verbose == -1:
+            self.log.disabled = True
```
wtdcode, line 555: `Logger.disabled` is undocumented attributed. Use `Logger.setLevel` instead. Besides, the if statement shall go to `ql_resolve_logger_level`.

qiling/const.py
```diff
@@ -36,11 +36,12 @@ class QL_OS(IntEnum):
     BLOB = 210
 
 class QL_VERBOSE(IntEnum):
-    OFF = 0
-    DEFAULT = 1
-    DEBUG = 4
-    DISASM = 10
-    DUMP = 20
+    OFF = -1    # turn off all the output
+    MINIMUM = 0 # output only warnings
```
wtdcode, line 40: This is causing a break change. Keep the value of `OFF` and assign another value for `-1` instead.

elicn:
I find `OFF` and `DISABLED` to be confusing.. and neither of them actually turns off logging altogether.
What about changing the existing `OFF` (not really off but show only warnings) to be `WARNING` and have `OFF` really disabling the logging (like `console=False` does)?

kabeor:
like @wtdcode said “This is causing a break change.”

elicn:
Can you elaborate what exactly gets broken, and because of what..?

kabeor:
people's project which based on our code need change their verbose to adapt ours

xwings:
Make sure you update the documents too

xwings (approved):

xwings (approved):

PR #1086, opened 2022-02-01
Qdb improvements

ucgJhe:
1. add basic support for x86
2. fix minor bugs

no assembly listing before current address for now
demo

### DIFF
--- a/qiling/debugger/qdb/utils.py
+++ b/qiling/debugger/qdb/utils.py
@@ -36,6 +36,15 @@ def dump_regs(ql: Qiling) -> Mapping[str, int]:
                 "r12", "sp", "lr", "pc",
                 )
 
+    elif ql.archtype == QL_ARCH.X86:
+
+        _reg_order = (
+                "eax", "ebx", "ecx", "edx",
+                "esp", "ebp", "esi", "edi",
+                "eip", "ss", "cs", "ds", "es",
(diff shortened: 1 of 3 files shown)


xwings:
hi,

already PR to rootfs ?

qiling/debugger/qdb/utils.py
```diff
@@ -121,6 +131,17 @@ def get_cpsr(bits: int) -> (bool, bool, bool, bool):
             )
 
 
+def get_x86_eflags(bits: int) -> (bool, bool, bool, bool, bool, bool):
```
elicn, line 134: This is not valid annotation. You may want to use `Dict[str, bool]` instead.

elicn:
Adding an architectures is a good opportunity to think of a way to generalize the API and consider implementing this as an OOP. That could make additional architecture support a lot easier and less messy..

ucgJhe:
> hi,
> 
> already PR to rootfs ?

it's an accident, fixed by following commit.

qiling/debugger/qdb/utils.py
ucgJhe: thx for the hint, fixed in following commit.

xwings:
Happy New Year!

xwings (approved):

cla7aye15I4nd (approved):

PR #1105, opened 2022-03-04
Use multitask unicorn for mcu

wtdcode:
This in fact implemented timeclock interrupt for Qiling(Unicorn) and thus can be used in mcu and multithreaded.

Still work in progress but any review is highly welcome.


cla7aye15I4nd:
I think MCU might not be a good scenario for multitask unicorn because usually MCUs are single core, but I think this would be a better solution for emulating multithreading programs and kernels (xv6)

Issue #1109, opened 2022-03-09
Migration to pyproject.toml? (PEP 517, 518 support)


nullableVoidPtr:
I can do this in a PR, but I wanted to ask if there were any objections before I get started on this.

As of recently, [`setup.py` has been considered deprecated](https://pip.pypa.io/en/stable/reference/build-system/setup-py/) in favor of a [build-system independent format](https://www.python.org/dev/peps/pep-0517/#source-trees).
Seeing as qiling only uses `setuptools` anyways, the plan to migrate the package to PEP 518 is to specify only `setuptools` in the `pyproject.toml` file, alongside dependencies, license, and other metadata.
This setup would either shrink or remove `setup.py` entirely; setuptools itself does not support editable installs without a `setup.py`, but that has a [workaround in `pip`](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#editable-installation), although this mandates a minimum pip version of 21.3.

elicn:
Keeping up to date with the standard is always good, but we have leave some room for slower user adoption (meaning not moving forward too fast). Specifically, this project made a decision to move to Python 3.8 just recently, so we have to make sure this change doesn't break anything.

That being said, I think this is a great idea.

xwings:
Close for now.

We updated the codebase for Qiling and Unicorn since this issue being posted.

Feel free to try the latest version.

PR #1111, opened 2022-03-09
Migrate to PEP 518

nullableVoidPtr:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR makes the `qiling` package removes the `setup.py` script and replaces it with the [declarative `setup.cfg` file](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and ensures that it is compatible with the new `pyproject.toml` meta build system specified in [PEP 518](https://www.python.org/dev/peps/pep-0518/), as discussed in #1109.

Most of the package metadata including dependencies are included in anticipation of [PEP 621/631](https://www.python.org/dev/peps/pep-0621/) support in `setuptools` (see https://github.com/pypa/setuptools/issues/1688 and https://github.com/pypa/setuptools/pull/2970). Alternatively, Qiling can use https://github.com/pypa/flit) to rely entirely on the one TOML file, and I believe `pip` would work just the same.


nullableVoidPtr:
Ah, sorry if I didn't make it clear.

The list of runtime dependencies is the same, but it is duplicated twice to conform to a finalized PEP, which would eventually be supported in `setuptools`.

I brought up `filt` as an *optional alternative*, which maintainers could **choose** to use, but for now, in order to make this PR work with the current `setuptools` version (what everyone uses), I have specified dependencies in both the standard project TOML, and the declarative setuptools config file. This would get rid of the deprecated `setup.py` process.

There should be little/no change to how contributors develop.

xwings:
@elicn what do you think ?

elicn:
I think that keeping up to date is always a good idea; see my full feddback at #1109.
Since I don't deal with the release and distribution processes I don't know whether this is going to add more burden or not. I guess it is for you to consider and decide.

xwings:
> Ah, sorry if I didn't make it clear.
> 
> The list of runtime dependencies is the same, but it is duplicated twice to conform to a finalized PEP, which would eventually be supported in `setuptools`.
> 
> I brought up `filt` as an _optional alternative_, which maintainers could **choose** to use, but for now, in order to make this PR work with the current `setuptools` version (what everyone uses), I have specified dependencies in both the standard project TOML, and the declarative setuptools config file. This would get rid of the deprecated `setup.py` process.
> 
> There should be little/no change to how contributors develop.

I still dont understand where they are two files. one is TOML and one setuptools ?

nullableVoidPtr:
> I still dont understand where they are two files. one is TOML and one setuptools ?

TOML (the PEP standard) is preferred as it would allow support for other build systems.
Initially, when I started this PR, `setuptools` [did not support TOML](https://github.com/pypa/setuptools/pull/2970), just it's own `setup.cfg`, but [now it does experimentally](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) (see: pypa/setuptools#3214). I included both files at the time of writing expecting that it would be supported soon.

So we have several options:
- Choose to keep one of the two files: 
  - `pyproject.toml` which is a new standard only just _experimentally_ supported in very recent versions of `setuptools>=61.0.0`, or
  - `setup.cfg` which works for only  `setuptools`, but is considered stable and the successor to the legacy `setup.py` method
- Keep both files, allowing Qiling developers to safely package and develop on `setuptools`, but also allows use of alternative systems like `poetry` or `flit`
- Close this PR and maintain the legacy `setup.py`.

Sorry if it still doesn't make sense; Python building and packaging is a bit of a mess 😅

wtdcode:
> > I still dont understand where they are two files. one is TOML and one setuptools ?
> 
> TOML (the PEP standard) is preferred as it would allow support for other build systems. Initially, when I started this PR, `setuptools` [did not support TOML](https://github.com/pypa/setuptools/pull/2970), just it's own `setup.cfg`, but [now it does experimentally](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) (see: [pypa/setuptools#3214](https://github.com/pypa/setuptools/issues/3214)). I included both files at the time of writing expecting that it would be supported soon.
> 
> So we have several options:
> 
> * Choose to keep one of the two files:
>   
>   * `pyproject.toml` which is a new standard only just _experimentally_ supported in very recent versions of `setuptools>=61.0.0`, or
>   * `setup.cfg` which works for only  `setuptools`, but is considered stable and the successor to the legacy `setup.py` method
> * Keep both files, allowing Qiling developers to safely package and develop on `setuptools`, but also allows use of alternative systems like `poetry` or `flit`
> * Close this PR and maintain the legacy `setup.py`.
> 
> Sorry if it still doesn't make sense; Python building and packaging is a bit of a mess 😅

It's always a big mess but funny things are that everything "just" happens to work.

Personally, I prefer `poetry` and it looks good to me.

.github/workflows/pythonpublish.yml
```diff
@@ -13,11 +13,10 @@ jobs:
         python-version: '3.x'
     - name: Install dependencies
       run: |
-        pip install setuptools wheel
+        pip install setuptools wheel build
     - name: Build distribution 📦
       run: |
-        pip install .
-        python setup.py sdist bdist_wheel
+        python -m build
```
wtdcode, line 19: Did you test this already? e.g. on a testpypi?

setup.py
```diff
@@ -1,102 +0,0 @@
-#!/usr/bin/env python3
```
wtdcode, line 1: Could we just leave setup.py here for compatibility? Users might fallback to this file if anything is wrong and we can safely remove it after we receive enough feedback.

nullableVoidPtr: Sure, we could note that users could use `pip install . --no-use-pep517` to force use of setup.py.

.github/workflows/pythonpublish.yml
nullableVoidPtr: sdist generation was one of the tests I had to resolve, and IIRC it worked; I'll test locally properly with nektos/act.

nullableVoidPtr:
It appears that a zip distribution is currently used as a dependency in the develop branch; I'll need to look into how to resolve this with PEP 518.

.github/workflows/pythonpublish.yml
nullableVoidPtr: Thanks for making me upload to TestPyPI as it had identified an issue with the `pyproject.toml` with the project URLs. I have uploaded it successfully. However, I have identified another problem, more in the other review thread.

setup.py
nullableVoidPtr: Ironically, it seems that including both `setup.py` and `pyproject.toml` may cause headaches, as when `setuptools.build_meta` is called, it parses `pyproject.toml` first, but it runs additional code, *including the execution of setup.py (with a default stub if non-existent, which unfortunately does not apply in the current state of the PR)* to retrieve metadata and other information.

I'll file an issue with `setuptools`, and most likely their solution would just be a warning during build, but the easy fix is to replace `setup.py` with the `setup.cfg`, or to remove `setup.py` entirely.

xwings:
@wtdcode @kabeor @elicn  there is some update and what do u guys think.

xwings:
@nullableVoidPtr there is a conflict to solve :)

kabeor:
It definitely make package info more readable, I think it should be merged once the conflict is solved.

nullableVoidPtr:
Hi all,

Totally forgot about this PR, life got in the way.
@kabeor @xwings, am I still to maintain both the TOML and the old setup.py, or should Qiling use pyproject.toml only?

xwings:
One question,

we save and read the version number from setup.py

This will be a problem if we remove from setup.py

nullableVoidPtr:
> we save and read the version number from setup.py
Ah right, what with the `dev` branch tagging. I currently don't see any code in-tree, and it should be fairly easy to automatically edit a TOML file, but I understand if it's a part of others' workflow and if it may be difficult to change.

Okay, so one option is to have `project.dynamic = ['version']`, and further something like `version = {attr = "setup.VERSION"}`. Probably not the best way to go about it, but it might work.

If the need for the one version is not needed in `setup.py`, then this could be just `__version__` in `qiling/__init__py`, removing the need for the `importlib` dependency there, though it's up to you.

FWIW, this diff works wonders:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 636daf17..023d426f 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,7 @@
 [project]
 name = "qiling"
-version = "1.4.5-dev"
 description = "Qiling is an advanced binary emulation framework that cross-platform-architecture"
+dynamic = ["version"]
 readme = "README.md"
 requires-python = ">=3.8"
 license = {text = "GPL-2.0-only"}
@@ -69,7 +69,7 @@ changelog = "https://github.com/qilingframework/qiling/blob/master/ChangeLog"
 qltool = "qltool:main"
 
 [build-system]
-requires = ["setuptools>=61.0.0", "wheel"]
+requires = ["setuptools>=61.0.0", "setuptools_scm[toml]>=6.2", "wheel"]
 build-backend = "setuptools.build_meta"
 
 [tool.setuptools.package-data]
@@ -78,3 +78,4 @@ qiling = ["profiles/*.ql"]
 "qiling.os.uefi" = ["guids.csv"]
 "qiling.arch.evm.analysis" = ["signatures.json"]
 
+[tool.setuptools_scm]
```
On my machine, this builds version `1.4.5.dev279+g0dcf436c.d20221021` on my machine, but may need tweaking to fit Qiling's purposes

xwings:
I guess there is a problem between qiling and qltools. that is the reason why importlib is there.

and having version number in 2 or 3 files is not a good way to do it.

@elicn what do you think ?

elicn:
I like @nullableVoidPtr's approach and would happily drop the `importlib` thing from `__init__.py`.

However, I didn't understand what is the problem with Qiling / qltool.
Can't they both rely on what's in `__init__.py` ?

xwings:
I dont remember why it all move to importlib, it was in __init__.py all this while until someone made some changes to it ?

maybe @wtdcode knows ?

wtdcode:
The importlib is introduced by @nullableVoidPtr .

I moved the version to a single file and get it by manually `exec` because previously Qiling setup.py will assume Qiling is already installed and retrieve Qiling version by import, which violates some setup.py assumptions.

xwings:
@wtdcode so, this PR is good to go ? some still need some fix ?

wtdcode:
I will again suggest @nullableVoidPtr testing the latest dev branch on the test PyPI.

Also, I doubt if our release CI still works after this change, which should be checked on test PyPI, too.

Issue #1113, opened 2022-03-14
Error: mapping needed but failed when emulate firmware wnap320


newthis:
***Describe the bug**
When emulate wnap320 firmware whichi can be downloaded from here
(https://github.com/zyw-200/FirmAFL/blob/master/firmware/WNAP320_V3.0.5.0.zip),
the qiling framework 1.4.2 throws mmap failure exception, 

[+]	mmap - mapping needed for 0x77706000
[x]	Syscall ERROR: ql_syscall_mmap DEBUG: Error: mapping needed but failed
Traceback (most recent call last):
  File "/home/test/PycharmProjects/QLearn/venv/lib/python3.8/site-packages/qiling/os/posix/syscall/mman.py", line 115, in syscall_mmap_impl
    ql.mem.map(mmap_base, mmap_size, prot, "[syscall_%s]" % api_name)
  File "/home/test/PycharmProjects/QLearn/venv/lib/python3.8/site-packages/qiling/os/memory.py", line 519, in map
    raise QlMemoryMappedError('Requested memory is unavailable')
qiling.exception.QlMemoryMappedError: Requested memory is unavailable

**Sample Code**
```python

 def fetch_syscall_parameter(ql: Qiling, addr: int, mlen: int, prot: int, flags: int, fd: int, pgoffset: int):
    msg = "mmap_len***: {}, addr: {}".format(hex(mlen), hex(addr))
    ql.log.debug(msg)

 project_root = "/home/test/repos/qiling/examples/rootfs/wnap320/squashfs-root"
 path = [project_root + "/sbin/lighttpd", "-f", "/etc/lighttpd.conf"]
 ql = Qiling(path, project_root, verbose=QL_VERBOSE.DEBUG, multithread=False)
 ql.set_syscall("mmap", fetch_syscall_parameter, intercept=QL_INTERCEPT.ENTER)
 ql.run()
```

**Expected behavior**
It seems that the byte order is incorrectly reversed for mmap length parameter in qiling.
The stream_open at address 0x42BB80 in lighttpd caculate /etc/lighttpd.conf
size should be 0x00001f9e, however when parameter is hooked in fetch_syscall_parameter,
the output is 0x9e1f0000, which leads to mmap failure.

**Additional context**
environemnt: python3.8 , qiling 1.4.2,
you can extract the firmware filesystem using binwalk.

sigeryang:
```
$ file ./sbin/lighttpd
[...]

Issue #1099, opened 2022-02-24
Unable to debug mips program


bagelbyte:
Hello there :dango: 

***Describe the bug**
when i run the `examples/hello_mips32el_linux_debug.py` program, I am unable step through the program.

**Sample Code**
run `python3 hello_mips32el_linux_debug.py`, and then in a separate terminal run

`gdb-multiarch`
> set architecture mips
> target remote localhost:9999
> b main

**Expected behavior**
I should be able to go to the next instruction, or have the program break when the `main` breakpoint is hit. 

**Screenshots**
In this example, I'm running qiling from docker, with the image built from the current dev branch. however I have also verified the same operation on two separate pieces of hardware, both running ubuntu 20.04, with qiling running outside of docker. 

```
root@useruser:~/examples# python3 hello_mips32el_linux_debug.py
[+]	Profile: Default
[+]	Mapped 0x400000-0x48d000
[+]	Mapped 0x49d000-0x4a2000
[+]	mem_start : 0x400000
[+]	mem_end   : 0x4a2000
[+]	mmap_address is : 0x774bf000
[=]	gdb> Stop at entry point: 0x4004c0
[=]	gdb> Listening on 127.0.0.1:9999
```

and then for gdb
```
➜  examples git:(dev) ✗ gdb-multiarch -q
(gdb) set architecture mips
The target architecture is assumed to be mips
(gdb) target remote localhost:9999
Remote debugging using localhost:9999
Reading /qiling/examples/rootfs/mips32el_linux/bin/mips32el_hello_static from remote target...
warning: File transfers from remote targets can be slow. Use "set sysroot" to access files locally instead.
Reading /qiling/examples/rootfs/mips32el_linux/bin/mips32el_hello_static from remote target...
Reading symbols from target:/qiling/examples/rootfs/mips32el_linux/bin/mips32el_hello_static...
(No debugging symbols found in target:/qiling/examples/rootfs/mips32el_linux/bin/mips32el_hello_static)
warning: unable to open /proc file '/proc/42000/task/42000/maps'
0xc0044000 in ?? ()
(gdb) ni
Cannot access memory at address 0xc0044000
```

also if i were to set a breakpoint on `main` and continue from gdb, it would just skip over the main breakpoint

**Additional context**
the machines this has been tested on are both x86_64

KA2010:
If you find solution, please share it, I have same issue.

sigeryang:
[...]

PR #1114, opened 2022-03-15
Improve support for MIPS / ARM big endian

sigeryang:

### Extra tests?

### Target branch?

### One last thing

-----

It solves some issues:

#1099: Fix mistakes in GDB address conversion
#1113: Add `stat` structs for MIPS big endian

and

* improves support for MIPS / ARM big endian
* adds tests and examples for mips32eb & armeb
* updates arm kernel helpers (might be necessary for armeb)
* allows timespec64 in 32 bit architectures

The support for ARM big endian depends on unicorn-engine/unicorn@e3d0a33ab83c9dd65134da0c0a10a6a91173a08c


qiling/os/posix/syscall/stat.py
```diff
@@ -319,6 +319,57 @@ class LinuxMips64Stat(ctypes.Structure):
 
     _pack_ = 8
 
+class LinuxMips32EBStat(ctypes.BigEndianStructure):
```
wtdcode, line 322: hmmm, where does this structs come from? Could you add a few docs?

sigeryang: It is the same with MIPS little endian, which was already documented at

https://github.com/qilingframework/qiling/blob/2b47ec76fcb7fb64cc9c3f562ce64968f4b0303d/qiling/os/posix/syscall/stat.py#L187-L320

Since there is no `#if defined(__ARMEB__)`-ish thing in this file, it indicates that both MIPS little endian and big endian share the same structs.

elicn:
We currently do not support MIPS BE.
Providing that support should be done with a proper Arch class and perhaps some more additions. Adding `if` statements all around only to support a few specific features is not how we want this to be.

xwings:
> We currently do not support MIPS BE. Providing that support should be done with a proper Arch class and perhaps some more additions. Adding `if` statements all around only to support a few specific features is not how we want this to be.

i think we do support MIPS BE and not ARM BE ? 
I am not sure.

xwings (approved):

cla7aye15I4nd (approved):

kabeor (approved):

PR #1121, opened 2022-03-28
fix(example): mistakes in fuzz_x8664_linux binary

chinggg:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

IMO, `fuzz.c` intended to read from `stdin` to buf and copy it to buf2, then print buf2 as string.
But there are two mistakes which do not affect the fuzzing result, I think they are typos:
- `*buf = getc(stdin) == 'A'` does not reads char, but the result of comparison
- `printf(buf2)` does not print buf2 as string

So I fix `fuzz.c` then compile it with `gcc -g` on Ubuntu 20.04 + gcc9.3, and modify the according address of `main` and `stack_chk` in `fuzz_x8664_linux.py`, it works as before.

If you think the fix is correct, the `arm_qnx` example may need fix as well.


examples/fuzzing/linux_x8664/fuzz.c
```diff
@@ -10,13 +10,13 @@ int fun(int i)
...
 
-    while (*buf = getc(stdin) == 'A')
+    while ((*buf = getc(stdin)) == 'A')
     {
         buf[i++] = *buf;
     }
 
     strncpy(buf2, buf, i);
-    printf(buf2);
+    printf("%s", buf2);
```
elicn, line 19: This is actually fixing an _Uncontrolled Format String_ vulnerability that could be triggered if `buf` was not limited only to `'A'` chars. Anyway, consider changing that to `puts(buf2)` that is functionally identical but much more efficient where there is no format string involved.

elicn (approved):
Good catch!

examples/fuzzing/linux_x8664/fuzz.c
chinggg: Thanks! This has been changed.

xwings:
Hey, thanks and welcome to Qiling !

PR #1140, opened 2022-04-25
Introduce PowerPC architecture to qiling

bet4it:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

The test binaries are provided in qilingframework/rootfs#29


xwings:
There is still some error in CI after adding the file.

qiling/arch/ppc_const.py
```diff
@@ -0,0 +1,128 @@
...
+    "f25": UC_PPC_REG_FPR25,
+    "f26": UC_PPC_REG_FPR26,
+    "f27": UC_PPC_REG_FPR27,
+    "f28": UC_PPC_REG_FPR28,
+    "f29": UC_PPC_REG_FPR29,
+    "f30": UC_PPC_REG_FPR30,
+    "f31": UC_PPC_REG_FPR31,
+}
+
+class MSR(IntEnum):
```
elicn, line 85: I think that `Flag` ot `IntFlag` would be a better fit here (see [here](https://docs.python.org/3.8/library/enum.html#enum.Flag)).

elicn (approved):
Great contribution!
I am curious whether extending Qiling with a new architecture was an easy task, or not..?

qiling/arch/ppc_const.py
bet4it: You could search `IntEnum` in the code. There are many usages of `IntEnum`, but none of `Flag` or `IntFlag`. So if we really want to change it, that could be another PR.

xwings (approved):

xwings:
Amazing!

xwings:
Thanks !

PR #1144, opened 2022-05-01
fix(memory): remove misused region bound check of unmap_all

### LINKED ISSUE #1141: `ql.mem.unmap_all` does not unmap all memory regions
**Describe the bug**
Mapping a memory region starting at address 0 and then calling `ql.mem.unmap_all` does not unmap the region.

**Sample Code**
```python
import qiling
from qiling.const import QL_ARCH

ql = qiling.Qiling(code=b"\0", archtype=QL_ARCH.ARM, ostype='linux')
ql.mem.map(addr=0, size=0x1000)
ql.mem.unmap_all()
assert ql.mem.map_info == [], ql.mem.map_info 
```

**Actual behavior**
```
$ python main.py
  File "main.py", line 7, in <module>
    assert ql.mem.map_info == [], ql.mem.map_info 
AssertionError: [(0, 4096, 7, '[mapped]', False)]
```

**Expected behavior**
No assertion error and the entire memory space is reclaimed (as the docstring says).

**Additional context**
This region is not unmapped because `begin` is 0, which causes the `begin and end` guard in the below code to be `False`. I am not sure why this check is even there - I think the `if` should be removed, and `self.unmap` should always be executed.

https://github.com/qilingframework/qiling/blob/59d31c22525f03cc38d61a7ea9908faa2d079d22/qiling/os/memory.py#L417-L423

chinggg:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

If `self.ql.uc.mem_regions()` works properly, there is no need to check `begin` and `and`. Moreover, `begin=0` is a false negative.

closes #1141

### DIFF
--- a/qiling/os/memory.py
+++ b/qiling/os/memory.py
@@ -419,8 +419,7 @@ class QlMemoryManager:
         """
 
         for begin, end, _ in self.ql.uc.mem_regions():
-            if begin and end:
-                self.unmap(begin, end - begin + 1)
+            self.unmap(begin, end - begin + 1)
 
     def is_available(self, addr: int, size: int) -> bool:
         """Query whether the memory range starting at `addr` and is of length of `size` bytes


elicn (approved):

wtdcode (approved):

wtdcode:
Thanks and welcome ; )

PR #1148, opened 2022-05-04
New and improved gdbserver

elicn:
Changes Highlights:
 - Complete re-design of the gdbserver
   - Generic design with minimal arch-specific code
   - Decoupled connection handler from frontend
   - Flexible registers information: info is read from XML files instead of being hardcoded
 - Serial protocol now supports escaping and run-length encoding
 - More supported gdb features
 - Many bugs and instability issues were fixes

TODO:
 - More testing
 - More architectures
   - RISC
   - RISC64
   - PowerPC


xwings:
Add in PowerPC too ?

elicn:
Ready to review.

kabeor (approved):

qiling/debugger/gdb/gdb.py
```diff
@@ -3,811 +3,973 @@
...
-            self.send_raw('$%s#%.2x' % (msg, self.checksum(msg)))
+            data = packet['data']
+            read_csum = int(packet['checksum'], 16)
+            calc_csum = GdbSerialConn.checksum(data)
+
+            if read_csum != calc_csum:
+                raise IOError(f'checksum error: expected {calc_csum:02x} but got {read_csum:02x}')
+
+            # follow gdbserver debug output format
+            self.log.debug(f'getpkt ("{GdbSerialConn.__printable_prefix(data).decode(ENCODING)}");')
```
wtdcode, line 849: This is too verbose, could we have a more verbose logging level? Sometimes users turn on debug level just to see some extra information of the emulation not the packets.

qiling/debugger/gdb/gdb.py
```diff
@@ -3,811 +3,973 @@
...
         else:
-            self.clientsocket.send(b'$%s#%.2x' % (msg, self.checksum(msg)))
-            self.netout.flush()
+            data = GdbSerialConn.escape(data)
+            data = GdbSerialConn.rle_encode(data)
+
+            packet = b'$' + data + b'#' + f'{GdbSerialConn.checksum(data):02x}'.encode()
+
+        # follow gdbserver debug output format
+        self.log.debug(f'putpkt ("{GdbSerialConn.__printable_prefix(data).decode(ENCODING)}");')
```
wtdcode, line 880: This, too

qiling/debugger/gdb/gdb.py
elicn: I agree, but at the same time I wanted to follow gdbserver debugging messages - as the comment there says.
The previous version of the server also printed out a message upon each incoming packet (though it didn't emit the payload..). Adding an excessive verbosity level is a great idea.. but what do you suggest to do in the meantime?

wtdcode:
I fail to get a pull request to your fork so I leave it in my fork. See https://github.com/wtdcode/qiling/commit/715695f0066d347a3fad3dbd0babece3911ba038 which is in https://github.com/wtdcode/qiling/tree/gdb-improv

qiling/debugger/gdb/gdb.py
wtdcode: I suggest adding a `QL_VERBOSE.TRACE` with a value of 5 (`logging.DEBUG` is 10) or so. This might not the goal of this PR, but I think it's a good time to do it. ;p

elicn:
> I fail to get a pull request to your fork so I leave it in my fork. See [wtdcode@715695f](https://github.com/wtdcode/qiling/commit/715695f0066d347a3fad3dbd0babece3911ba038) which is in https://github.com/wtdcode/qiling/tree/gdb-improv

I would go with mapping procfs files through `fs_mapper` instead. That would be cleaner and safer.
Actually, I was planning on creating some of the commonly used procfs and vfs files for Linux, but that is for another PR..

wtdcode:
> > I fail to get a pull request to your fork so I leave it in my fork. See [wtdcode@715695f](https://github.com/wtdcode/qiling/commit/715695f0066d347a3fad3dbd0babece3911ba038) which is in https://github.com/wtdcode/qiling/tree/gdb-improv
> 
> I would go with mapping procfs files through `fs_mapper` instead. That would be cleaner and safer. Actually, I was planning on creating some of the commonly used procfs and vfs files for Linux, but that is for another PR..

That's different. I don't want to mess up with users' `fs_mapper` and that's why I add it here.

wtdcode:
> > I fail to get a pull request to your fork so I leave it in my fork. See [wtdcode@715695f](https://github.com/wtdcode/qiling/commit/715695f0066d347a3fad3dbd0babece3911ba038) which is in https://github.com/wtdcode/qiling/tree/gdb-improv
> 
> I would go with mapping procfs files through `fs_mapper` instead. That would be cleaner and safer. Actually, I was planning on creating some of the commonly used procfs and vfs files for Linux, but that is for another PR..

1. Users might hijack `/proc/42000/map` in any way, and it might be surprising to break gdbserver.
2. Note `42000` in the path, which is a gdbserver magic value, I don't think you would like to add this to users Qiling instance.

elicn:
There are already several hardcoded procfs paths embedded in the code, and it feels quite messy [not to mention that most of them are just poorly emulated..]. Concentrating all of them at one point could be easier for the users find and replace with their own functionality - should they need to. In that context `fs_mapper` makes perfect sense to me, since we strive for maximum flexibility, including letting the users shoot themseves i the foot.
This is also another step towards a virtualized filesystem with minimal side effects on the hosting system [though this is still a far goal..]

wtdcode:
> There are already several hardcoded procfs paths embedded in the code, and it feels quite messy [not to mention that most of them are just poorly emulated..]. Concentrating all of them at one point could be easier for the users find and replace with their own functionality - should they need to. In that context `fs_mapper` makes perfect sense to me, since we strive for maximum flexibility, including letting the users shoot themseves in the foot.
> This is also another step towards a virtualized filesystem with minimal side effects on the hosting system [though this is still a far goal..]

My point is that, this map hijacking belongs to gdbserver itself, so it has nothing to do with the emulated program. **Note, these two things are different**, the one of gdbserver is the real one of unicorn’s while the other is emulated or hijacked by Qiling.

This in fact really serves for “maximum flexibility” to allow users to hijack mapping on emulated filesystems while not breaking the gdbserver’s viewing real unicorn’s mappings.

To be honest, I can’t come up with any scenario that gdbserver must return the hijacked mappings.

elicn:
Good point.

qiling/debugger/gdb/gdb.py
elicn: What is  your suggestion until we do?
Disable the getpkt / putpkt messages, or leave it as-is?

wtdcode: The debug messages are indeed helpful so I would like to leave it as-is. I would draft another PR for `QL_VERBOSE.TRACE`.

elicn: I have something a bit different in mind for further verbosity levels.
Let's discuss that before you start working on this.

wtdcode (approved):

wtdcode:
Looks like my commits for gdbserver mappings are not merged. Would you like to review it in another PR?

qiling/debugger/gdb/utils.py
```diff
@@ -3,117 +3,76 @@
...
-            if self.ql.arch.type == QL_ARCH.ARM:
-                if self.ql.arch.is_thumb:
-                    address += 1
+    def resume_emu(self, address: Optional[int] = None, steps: int = 0):
+        if address is None:
+            address = self.ql.arch.regs.arch_pc
 
-            self.current_address = address
+        if self.ql.arch.type == QL_ARCH.ARM and self.ql.arch.is_thumb:
+            address += 1
```
wtdcode, line 73: `|=1` seems more reasonable?

wtdcode:
```
diff --git a/qiling/debugger/gdb/utils.py b/qiling/debugger/gdb/utils.py
index 30bba47e..8008f6ba 100644
--- a/qiling/debugger/gdb/utils.py
+++ b/qiling/debugger/gdb/utils.py
@@ -74,5 +74,5 @@ class QlGdbUtils:
 
         op = f'stepping {steps} instructions' if steps else 'resuming'
         self.ql.log.info(f'{PROMPT} {op} from {address:#x}')
-
+        self.ql.uc.ctl_flush_tb()
         self.ql.emu_start(address, self.exit_point, count=steps)
```

The diff to fix singlestep on next Unicorn release.

elicn:
@wtdcode 
I suspect that pushing additional commits after I rebased my origin branch would break.
Let me apply additional the changes in a different PR.

PR #1150, opened 2022-05-07
Rewrite package data reading

nullableVoidPtr:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

Using `__file__` to load internal resources is not recommended, as it may not work with import hooks or Python Eggs. Here, I am using the standard `pkgutil` (a wrapper around another standard library, `importlib.resources`) to retrieve load the data.

Note: `pkgutil.get_data` unfortunately returns `bytes` not strings or a file object, so I have to decode it first before passing it on. The alternative here is to use `importlib.resources.files` which returns a `Path` object (so, `ilr.files("qiling") / "profiles" / f"{os}.ql"`) which you can then call `open()` to return a `TextIO`, but it is only supported in 3.9.

### DIFF
--- a/.github/workflows/giteesync.yml
+++ b/.github/workflows/giteesync.yml
@@ -5,6 +5,7 @@
  jobs:
    deploy:
      runs-on: ubuntu-latest
+     if: github.repository_owner == 'qilingframework'
      steps:
      - uses: actions/checkout@v2
        with:
--- a/CREDITS.md
+++ b/CREDITS.md
@@ -54,6 +54,7 @@
 - danielmoos
 - sigeryang
 - bet4it
+- nullableVoidPtr
 
 
 #### Legacy Core Developers
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,5 +1,0 @@
-recursive-include qiling/debugger/gdb/xml *
-recursive-include qiling/extensions/windows_sdk/defs *
-recursive-include qiling/profiles *
-include qiling/os/uefi/guids.csv
-include qiling/arch/evm/analysis/signatures.json
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,3 +2,4 @@ recursive-include qiling/debugger/gdb/xml *
 recursive-include qiling/extensions/windows_sdk/defs *
 recursive-include qiling/profiles *
 include qiling/os/uefi/guids.csv
+include qiling/arch/evm/analysis/signatures.json
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,4 +2,3 @@ recursive-include qiling/debugger/gdb/xml *
 recursive-include qiling/extensions/windows_sdk/defs *
 recursive-include qiling/profiles *
 include qiling/os/uefi/guids.csv
-include qiling/os/qnx/syspage.bin
--- a/qiling/arch/evm/analysis/signatures.py
+++ b/qiling/arch/evm/analysis/signatures.py
@@ -1,4 +1,4 @@
-import os
+import pkgutil
 import re
 import logging
 import json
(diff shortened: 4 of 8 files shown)


elicn:
I agree that `__file__` may not be the right way to go, but neither is `pkgutil`. We could use `inspect.getfile` (see [here](https://docs.python.org/3.8/library/inspect.html#inspect.getfile)), which is the preferred way to replace `__file__` and requires less code modifications.

nullableVoidPtr:
> but neither is pkgutil

Would it be because of how the code would need to be restructured? Yeah, it's not ideal, but if either 3.9 is supported (unlikely I think) or if the backported `importlib_resources` packages is also added as a dependency (also unlikely), this is the way to do it (besides using the complicated `ResourceReader`)

> `inspect.getfile` 
As I understand it, it's basically a drop-in replacement for `__file__`, which can then be passed onto `os` functions, which may not be aware of non-standard loaders. `importlib` (and subsequently pkgutil) are, and can handle file loading for you.

elicn:
`inspect.getfile` is a good replacement for `os.path.abspath(__file__)`, since it just replaces the usage of `__file__` without requiring too much modifications. It follows the object passed to it (class, method, etc.) and returns the filename where it was defined. I find it very comfortable when you need to retreive the filename in which a specific class was defined - or a local class, if you want the `__file__` equivalent.

 I tend to prefer `pkgutil.get_data` less since it mixes the filename resolution and content retrieval in one statement and relies on exceptions to indicate that the resouce was not found. Also, as far as I understand it requires the code to be appear as a package (not too bad for itself, but will perhaps require annoying empty `__init__.py` files only for this purpose).

I didn't follow your comment on the non-standard loaders; do we really need this, or this is just a 'nice to have' feature we gain there..? To me it looks like the latter (.. unless I am missing it?) and not a strong reasoning for itself.

nullableVoidPtr:
> I didn't follow your comment on the non-standard loaders; do we really need this, or this is just a 'nice to have' feature we gain there..? To me it looks like the latter (.. unless I am missing it?) and not a strong reasoning for itself.

More of a nice-to-have, don't worry too much about it. There is a weird-but-standard(?) case in where __file__ and `inspect.getfile` does work with Python's `ZipImporter`, but not when the path is then passed onto `open` (`open('pkg.zip/__init__.py')`).

> I tend to prefer pkgutil.get_data less since it mixes the filename resolution and content retrieval in one statement and relies on exceptions to indicate that the resouce was not found.

Good call, and good catch! `get_data` actually returns None, but it seems that my leftover code from using `importlib.resources.files` snuck in. It was far more convenient (returning a Path-like object), but it failed on CI as 3.8 didn't support it.

> Also, as far as I understand it requires the code to be appear as a package (not too bad for itself, but will perhaps require annoying empty __init__.py files only for this purpose).

`pkgutil.get_data` can actually "dive deeper" into packages, that is, it can go into a non-package subdirectory that is physically located within another package (see the diffs to the gdb code where it relies on the arch and further the xfercmd file). This is something that 

Nevertheless, I can update this PR to use `inspect.getfiles` shortly.

nullableVoidPtr:
@elicn: made changes you wanted but wanted to bring this up:
It seems that `inspect.getfile` is a drop-in replacement for `__file__` because it is semantically the same.
If you pass a frame object to `inspect.getfile`, it is the same as `current_frame_function.co_filename`.
Alternatively, if you pass a class object to `inspect.getfile`, it is the same as `module_that_defined_class.__file__`.
(Again, properly accounting for import hooks is just a nice-to-have.)

I've also elected to use the `Path` class to traverse into the package dir. Let me know what you think.

elicn:
Thanks for that; I always appreciate a good technical discussion (: 

A few petty notes though:
 - Since `currentframe` may be somewhat problematic, I prefer to use a local class or method as an anchor
 - Where only a path is needed without any filesystem shenanigans, I prefer `PurePath` rather than `Path` [even if the file needs to be opened]
 - The changes in `gdb` are about to become irrelevant as I recently re-wrote the entire module

elicn (approved):

xwings (approved):

kabeor (approved):

kabeor:
Cool, thanks!

PR #1151, opened 2022-05-08
Support memory slicing

wtdcode:

### Coding convention?

### Extra tests?

### Changelog?

- [] This PR doesn't need to update Changelog.

### Target branch?

### One last thing

-----

This is one of several efforts to make the current wrapper more pythonic to fix https://github.com/unicorn-engine/unicorn/issues/201.

This PR makes `ql.mem` support slicing.

### DIFF
--- a/qiling/os/memory.py
+++ b/qiling/os/memory.py
@@ -3,6 +3,7 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
+from ctypes import Union
 import os, re
 from typing import Any, Callable, List, MutableSequence, Optional, Sequence, Tuple
 
@@ -62,6 +63,29 @@ class QlMemoryManager:
 
     def __write_string(self, addr: int, s: str, encoding: str):
         self.write(addr, bytes(s, encoding) + b'\x00')
+
+    def __getitem__(self, key: Union[slice, int]) -> bytearray:
+        if isinstance(key, slice):
+            start = key.start
+            stop = key.stop
+            step = key.step
+
+            if step and step != 1:
+                # step != 1 means we have to do copy, don't allow it.
+                raise IndexError("Only support slicing continous memory")
+
+            return self.ql.mem.read(start, stop - start)
+        elif isinstance(key, int):
+            return self.ql.mem.read(key, 1)
+        else:
+            raise KeyError("Wrong type of key")
+    
+    def __setitem__(self, key: Union[slice, int], value: Union[bytes, bytearray]):
+        if isinstance(key, int):
+            self.ql.mem.write(key, value)
+        else:
+            # Slicing doesn't make sense in writing.
+            raise KeyError("Wrong type of key")
 
     # TODO: this is an obsolete utility method that should not be used anymore
     # and here for backward compatibility. use QlOsUtils.read_cstring instead


qiling/os/memory.py
```diff
@@ -63,6 +63,29 @@ def __read_string(self, addr: int) -> str:
...
+        if isinstance(key, slice):
+            start = key.start
+            stop = key.stop
+            step = key.step
+
+            if step is not None and step != 1:
+                # step != 1 means we have to do copy, don't allow it.
+                raise IndexError("Only support slicing continous memory")
+
+            return self.ql.mem.read(start, stop - start)
```
elicn, line 76: Probably needs a check that `start` is not greater than `stop`, or we end up with a negative read size.
Perhaps something like `size = max(0, stop - start)` would do, but I don't remember how `read` behaves with a zero-length read size. Either way, the `bytearray` semantics defines a return value of an empty `bytearray` in such a case.

qiling/os/memory.py
```diff
@@ -63,6 +63,29 @@ def __read_string(self, addr: int) -> str:
...
+        elif isinstance(key, int):
+            return self.ql.mem.read(key, 1)
```
elicn, line 78: Since `mem.read` returns a `bytearray` it would be more intuitive to follow `bytearray` semantics here as well.
Perhaps return an integer value for a single index, like bytearrays do?

qiling/os/memory.py
```diff
@@ -63,6 +63,29 @@ def __read_string(self, addr: int) -> str:
...
+        else:
+            raise KeyError("Wrong type of key")
+    
+    def __setitem__(self, key: Union[slice, int], value: Union[bytes, bytearray]):
+        if isinstance(key, int):
```
elicn, line 83: Writing something like `ql.mem[0x1234] = b'hello world'` seems weird to me, as I would expect an integer key to represent a single byte, and not a starting location. That said, specifying a slice would not make sense either..

qiling/os/memory.py
```diff
@@ -63,6 +63,29 @@ def __read_string(self, addr: int) -> str:
...
+            self.ql.mem.write(key, value)
+        else:
+            # Slicing doesn't make sense in writing.
```
elicn, line 86: That actually could make sense if we followed the `bytearray` semantics.
Following it would not be entirely possible, since it allows expanding and deleting slots - which we can't have with memory, but we can still overwrite the given slice area with the given bytes; if the slice size doesn't match the length of the data, data could be cropped if larger, or repeated if smaller.

qiling/os/memory.py
```diff
@@ -63,6 +63,29 @@ def __read_string(self, addr: int) -> str:
...
+                # step != 1 means we have to do copy, don't allow it.
+                raise IndexError("Only support slicing continous memory")
+
+            return self.ql.mem.read(start, stop - start)
+        elif isinstance(key, int):
+            return self.ql.mem.read(key, 1)
+        else:
+            raise KeyError("Wrong type of key")
+    
+    def __setitem__(self, key: Union[slice, int], value: Union[bytes, bytearray]):
```
elicn, line 82: As far as I remember `mem.write` accepts only `bytes` and will fail on `bytearray` data,

elicn (commented):
Great idea!

qiling/os/memory.py
wtdcode: No, ctypes will marshal most bytes like objects for `uc_mem_write` though I didn't find some relevant docs.

qiling/os/memory.py
wtdcode: If `size=0`, the buffer is `bytearray(b'')` and left untouched in `uc_mem_read`

qiling/os/memory.py
wtdcode: cropped is accepted while repeat might lead to performance overhead.

qiling/os/memory.py
wtdcode: Okay... you are correct. `bytearray` is not acceptable.

qiling/os/memory.py
wtdcode: done

qiling/os/memory.py
wtdcode: now, it looks like `ql.mem[0x123] = 12`, which looks like `bytearray`

qiling/os/memory.py
wtdcode: done

qiling/os/memory.py
wtdcode: done

qiling/os/memory.py
elicn: `bytearray` would allow you to do something like `ba[3] = b'hello'`, and by that insert the data while expanding the array. That is not applicable to memory, so I honestly don't know what would make more sense here. Both options look wierd to me..

qiling/os/memory.py
```diff
@@ -63,6 +63,43 @@ def __read_string(self, addr: int) -> str:
...
+            stop = key.stop
+            step = key.step
+
+            if step is not None and step != 1:
+                raise IndexError("Only support slicing continous memory")
+            
+            if start is None:
+                raise IndexError("The start of memory is not supplied")
+            
+            if len(value) > stop - start:
```
elicn, line 96: What happens when data length is less than slice length? I thought you are going to throw an excpetion if sizes do not match (i.e. when `len(value) != (stop - start)`)

elicn: What about not enforcing the stop value? Where slice `stop` is not specified, it would mean either "end of memory" (or a "don't care", if you don't like the idea of spanning till the end of memory)

For example (note the missing stop value):
```python
ql.mem[addr:] = b'hello'
```
Pesonally I would prefer leaving the `int` key behavior to be replacing a single byte in the specified location, because that way the behavior of something like `ql.mem[addr] = data` is not ambiguous. Requiring the user to specify a slice with an empty `stop` would make it clearer. All that being said, both approaches make sense and acceptable.

wtdcode: This looks great and intuitive!

wtdcode: Done

qiling/os/memory.py
wtdcode: Resolved by `ba[3:] = b"hello"`.

qiling/os/memory.py
```diff
@@ -63,6 +63,43 @@ def __read_string(self, addr: int) -> str:
...
+            stop = key.stop
+            step = key.step
+
+            if step is not None and step != 1:
+                # step != 1 means we have to do copy, don't allow it.
+                raise IndexError("Only support slicing continous memory")
+
+            return self.ql.mem.read(start, max(0, stop - start))
+        elif isinstance(key, int):
+            return self.ql.mem.read(key, 1)[0]
```
LukeSerne, line 78: This line returns an `int`, not `bytearray`, which is somewhat counterintuitive. I think you should remove the `[0]`. Also, I'm not sure why this code uses `self.ql.mem` instead of `self`. In this class, they reference / point to the same thing, right?
```suggestion
            return self.read(key, 1)
```

wtdcode: > This line returns an `int`, not `bytearray`, which is somewhat counterintuitive. I think you should remove the `[0]`. 

Because we (try to) provide the same interface as `bytearray`, e.g.:

```
In [1]: a = bytearray(b"\x01\x02")

In [2]: type(a[0])
Out[2]: int

In [3]:
```

> Also, I'm not sure why this code uses `self.ql.mem` instead of `self`. In this class, they reference / point to the same thing, right?

Good point.

PR #1153, opened 2022-05-10
Raw shellcode

wtdcode:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR is based on my previous mem slicing PR #1151, which is also one of several phases to make Qiling a bit more pythonic and easy to use.

The motivation of this PR is that, the current implementation is rather not flexible for shellcode usage, e.g.

1. The shellcode must be specified during `Qiling.__init__`.
2. The shellcode can't be changed anymore, or many our internal functions would break like `ql.run`.

In general, the previous design assumes that only one shellcode is written and never changed and thus this PR improves flexibility.

### DIFF
--- a/examples/shellcode_run.py
+++ b/examples/shellcode_run.py
@@ -25,6 +25,14 @@ if __name__ == "__main__":
     ql = Qiling(code=ARM64_LIN, archtype="arm64", ostype="linux", verbose=QL_VERBOSE.DEBUG)
     ql.run()
 
+    print("\nCreate shellcode after creating Qiling instance")
+    ql = Qiling(archtype="arm64", ostype="linux", verbose=QL_VERBOSE.DEBUG)
+    addr = ql.mem.map_anywhere(0x1000)
+    ql.mem[addr:] = ARM64_LIN
+    stack = ql.mem.map_anywhere(0x1000)
+    ql.arch.regs.arch_sp = stack
+    ql.run(begin=addr, end=addr + len(ARM64_LIN))
+
     print("\nLinux ARM 32bit Shellcode")
     ql = Qiling(code=ARM_LIN, archtype="arm", ostype="linux", verbose=QL_VERBOSE.DEBUG)
     ql.run()
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -88,14 +88,14 @@ class Qiling(QlCoreHooks, QlCoreStructs):
         # argv setup #
         ##############
         if argv is None:
-            argv = ['qilingcode']
+            argv = []
 
         elif not os.path.exists(argv[0]):
             raise QlErrorFileNotFound(f'Target binary not found: "{argv[0]}"')
 
         self._argv = argv
-        self._path = self.argv[0]
-        self._targetname = os.path.basename(self.path)
+        self._path = self.argv[0] if len(self.argv) != 0 else ""
+        self._targetname = os.path.basename(self.path) if self.path is not None else ""
 
         ################
         # rootfs setup #
--- a/qiling/debugger/gdb/gdb.py
+++ b/qiling/debugger/gdb/gdb.py
@@ -54,7 +54,7 @@ class QlGdb(QlDebugger, object):
         if self.ql.baremetal:
             load_address = self.ql.loader.load_address
             exit_point = load_address + os.path.getsize(ql.path)
-        elif self.ql.code:
+        elif self.ql.code or len(self.ql.argv) == 0:
             load_address = self.ql.os.entry_point
             exit_point = load_address + len(ql.code)
         else:
@@ -503,7 +503,7 @@ class QlGdb(QlDebugger, object):
                         self.send("l" + file_contents)
 
                 elif subcmd.startswith('Xfer:auxv:read::'):
-                    if self.ql.code:
+                    if self.ql.code or len(self.ql.argv) == 0:
                         return
 
                     if self.ql.os.type in (QL_OS.LINUX, QL_OS.FREEBSD):
--- a/qiling/loader/blob.py
+++ b/qiling/loader/blob.py
@@ -14,6 +14,11 @@ class QlLoaderBLOB(QlLoader):
         self.load_address = 0
 
     def run(self):
+
+        # Shellcode case.
+        if len(self.ql.argv) == 0:
+            return
+
         self.load_address = self.ql.os.entry_point      # for consistency
 
         self.ql.mem.map(self.ql.os.entry_point, self.ql.os.code_ram_size, info="[code]")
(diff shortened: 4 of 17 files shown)


wtdcode:
I'm also going to test against #1148 after it gets merged.

elicn:
There are several things I don't understand about this PR:
1. Do we really need flexibility around loading the shellcode? Why not just set it as part of init as we do today?
2. As far as I see it there is always `ql.argv` or `ql.code`, but never both. Though I agree that the handling of "shellcode mode" vs. "loaded file mode" is a bit cumbersome, these changes seem to make that even more complicated with a additional checks. What am I missing?
3. Wouldn't asserting there is either `ql.argv` or `ql.code` (but not both) suffice? Once od them must exist; what is the meaning of `else` in:
```python
if self.ql.code:
  # shellcode mode
elif len(self.ql.argv) == 0:
  # loaded file mode
else:
  # ??
```

wtdcode:
> There are several things I don't understand about this PR:
> 
> 1. Do we really need flexibility around loading the shellcode? Why not just set it as part of init as we do today?
> 2. As far as I see it there is always `ql.argv` or `ql.code`, but never both. Though I agree that the handling of "shellcode mode" vs. "loaded file mode" is a bit cumbersome, these changes seem to make that even more complicated with a additional checks. What am I missing?
> 3. Wouldn't asserting there is either `ql.argv` or `ql.code` (but not both) suffice? Once od them must exist; what is the meaning of `else` in:
> 
> ```python
> if self.ql.code:
>   # shellcode mode
> elif len(self.ql.argv) == 0:
>   # loaded file mode
> else:
>   # ??
> ```

> 1. Do we really need flexibility around loading the shellcode? Why not just set it as part of init as we do today?

Say, you have a shellcode and after `ql.run` you move it to elsewhere or changes its length and you call `ql.run` again, then it breaks.

> 2. As far as I see it there is always ql.argv or ql.code, but never both. Though I agree that the handling of "shellcode mode" vs. "loaded file mode" is a bit cumbersome, these changes seem to make that even more complicated with a additional checks. What am I missing?
> 3. Wouldn't asserting there is either ql.argv or ql.code (but not both) suffice? Once od them must exist; what is the meaning of else in:

The test for `len(self.ql.argv) == 0` doesn't mean `loaded file mode`. It means something like `Qiling(code=None)`.

elicn:
> Say, you have a shellcode and after ql.run you move it to elsewhere or changes its length and you call ql.run again, then it breaks.

Is this a real usecase..? I guess that all other scenarios will break as well since there are some manual adjustments to be made anyway if you want to re-run the same Qiling instance (e.g. clearing hooks, etc.)

> The test for len(self.ql.argv) == 0 doesn't mean loaded file mode. It means something like Qiling(code=None).

My bad, I actually meant to ask what is the difference between `if self.ql.code` and `elif len(self.ql.argv) == 0` where one of them must exist. I'm not sure what `else` means in that context..

wtdcode:
> > Say, you have a shellcode and after ql.run you move it to elsewhere or changes its length and you call ql.run again, then it breaks.
> 
> Is this a real usecase..? I guess that all other scenarios will break as well since there are some manual adjustments to be made anyway if you want to re-run the same Qiling instance (e.g. clearing hooks, etc.)

Of course, consider something like Self Modified Code emulation. Of course, I'm not talking of current implementation can't do but it's quite not flexible. For other scenarios you mentioned, in most cases, we don't need such flexibility.

Just take a really simple scenario as an example:

1. The shellcode produces some new code.
2. I would like to execute the code with the existing `Qiling` instance. Unfortunately, it's impossible.

Another scenario as I mentioned already, the shellcode might not be available as the time of creating the `Qiling` instance.

> 
> > The test for len(self.ql.argv) == 0 doesn't mean loaded file mode. It means something like Qiling(code=None).
> 
> My bad, I actually meant to ask what is the difference between `if self.ql.code` and `elif len(self.ql.argv) == 0` where one of them must exist. I'm not sure what `else` means in that context..

- `self.ql.code` is set => We got a shellcode in `Qiling.__init__` so we help users to setup code space and possible stacks,
-  `self.ql.code` is not set while `self.ql.argv` is empty. Users are wiling to provide shellcode afterwards, do nothing.
-  `else`. Users provide a file and this is the most common case.

Again, let me clarify my motivation a bit more:

Using raw Unicorn bindings is painful and full of tedious work while Qiling sheld a light on this by providing numerous utilities. However, the assumption that the shellcode is already available (and some other internal assumptions as mentioned) as the time of creating `Qiling` instance doesn't always hold when interacting with shellcode. In this case, I would like `Qiling` to provide the maximum flexibility with various helpers (imagine debugging a shellcode through gdb to help understand better).

elicn:
I think I am starting to understand the direction you're going, but I am still not sure why this cannot happen with the current implementation. A shellcode provided at initialization time has a certain size and Qiling copies it to memory, and... that is pretty much it.
For the specific unique case of an unknown shellcode (.. which honestly I still struggle to understand), why can't someone just initialize Qiling with a nopsled shellcode [large enough to make room for any shellcode they may choose] and then overwrite it in memory with a new one? At the end of the day the content of `ql.code` becomes meaningless as soon as it is copied to memory. Then all that matters is what is in that memory location.

Sorry for bothering, but I am trying to understand the justification here. The code around shellcode handling is already messy, and that PR makes it even more messy than it is today. I had some preliminary thoughts about redesigning the way Core, OS and Loader are interacting with each other, by introducing a concept of Process objects (in short, the Loader is static, it loads a Process and had it over to the OS. All the specific process context is saved there instead of being stored on Loader and OS). That would decouple Qiling from a specific program (or shellcode), and let the user load multiple processes on the same Qiling instance.

I would use this usecase as an additional incentive to drive this re-design.

wtdcode:
> I think I am starting to understand the direction you're going, but I am still not sure why this cannot happen with the current implementation. A shellcode provided at initialization time has a certain size and Qiling copies it to memory, and... that is pretty much it.

But still, I would like to somehow support `Qiling(code=None)`. At least, I think it's a **valid** use case and it should not be too difficult to support.

>At the end of the day the content of `ql.code` becomes meaningless as soon as it is copied to memory. Then all that matters is what is in that memory location.

Incorrect. The length of `ql.code` is used in `ql.run`, though can be fixed without messing up more.

> Sorry for bothering, but I am trying to understand the justification here. The code around shellcode handling is already messy, and that PR makes it even more messy than it is today. I had some preliminary thoughts about redesigning the way Core, OS and Loader are interacting with each other, by introducing a concept of Process objects (in short, the Loader is static, it loads a Process and had it over to the OS. All the specific process context is saved there instead of being stored on Loader and OS). That would decouple Qiling from a specific program (or shellcode), and let the user load multiple processes on the same Qiling instance.

Any estimation on this? This looks cleaner indeed.

PR #1172, opened 2022-06-14
feat: add r2 extension

chinggg:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR introduces radare2 extension and adds an example `hello_r2.py` to show its basic usage.

My main concerns are:

- the organization of `R2`'s members
  - `dataclass` vs `namedtuple`
- how many r2 output details should be kept
  - section
  - string
  - function
  - flag
  - xref

Feel free to review it and give your advice.


qiling/extensions/r2/r2.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __init__(self, **kwargs):
+        names = set([f.name for f in fields(self)])
+        for k, v in kwargs.items():
+            if k in names:
+                setattr(self, k, v)
+
+
+class R2:
+    def __init__(self, ql: Qiling):
```
elicn, line 112: Suggesting to allow the user to specify r2 evars to allow more analysis flexibility.
That could be done perhaps by accepting a dictionary of keys and values (e.g. `{'anal.hasnext' : 'true', 'anal.depth' : 5}` or any other reasonable format. Consider also allowing an "r2 init script" to let the user run a few r2 commands before starting the analysis.

examples/extensions/r2/hello_r2.py
```diff
@@ -0,0 +1,30 @@
...
+
+def my_sandbox(path, rootfs):
+    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEFAULT)
+    r2 = R2(ql)
+    assert(ql.loader.images[0].base == r2.baddr)
+    addrs = ql.mem.search(b'Hello world!')
+    addr = r2.strings['Hello world!'].vaddr
+    assert(addr == addrs[0])
+    ql.mem.write(addr, b"No hello, Bye!\x00")
+    ql.hook_address(func, r2.functions['main'].offset)
```
elicn, line 26: Does `offset` contains the address or offset..? If only offset, I'll be surprised if it worked.

qiling/extensions/r2/r2.py
```diff
@@ -0,0 +1,156 @@
...
+
+    def __init__(self, **kwargs):
+        names = set([f.name for f in fields(self)])
+        for k, v in kwargs.items():
+            if k in names:
+                setattr(self, k, v)
+
+
+@dataclass(unsafe_hash=True)
+class Section:
```
elicn, line 29: Unless there is a fields post-processing or any additional functionality, these classes could be `typing.NamedTuple`.

elicn (commented):
Thanks for the contribution! This functionality looks very cool.

I think this is a nice PoC that could be even cooler if it didn't take r2 data "as-is" and provided some post-processing, like turning all offsets into a vaddrs.

Anyway, I've made additional comments there.

examples/extensions/r2/hello_r2.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def func(ql: Qiling, *args, **kwargs):
+    ql.os.stdout.write(b"=====hooked main=====!\n")
+    return
+
+def my_sandbox(path, rootfs):
+    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEFAULT)
+    r2 = R2(ql)
+    assert(ql.loader.images[0].base == r2.baddr)
```
wtdcode, line 21: I don't think `assert` is a good style for examples. They should be something in tests.

setup.py
```diff
@@ -42,6 +42,9 @@
     "fuzz" : [
         "unicornafl>=2.0.0;platform_system=='Windows'",
         "fuzzercorn>=0.0.1;platform_system=='Linux'"
+    ],
+    "SCA" : [
+        "r2libr"
```
wtdcode, line 47: @elicn maybe we could add `r2libr` as an essential dep? My thought roughly is:

- First of all, unlike the previously discussed `udbserver`, `r2libr` is maintained by me, in other words, by Qiling team directly.
- `r2libr` could be tightly integrated into our loader.
- The `r2libr` could help us polish some old API like `set_api` with the additional information.
- There seems no use case that users really would like to use QIling without `r2libr`?

qiling/extensions/r2/r2.py
```diff
@@ -0,0 +1,156 @@
...
+        names = set([f.name for f in fields(self)])
+        for k, v in kwargs.items():
+            if k in names:
+                setattr(self, k, v)
+
+
+class R2:
+    def __init__(self, ql: Qiling):
+        super().__init__()
+        path = ql.path.encode()
```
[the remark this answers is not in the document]
wtdcode: Good point, I remember r2 supports writing code to analyze?

setup.py
wtdcode: Also note, `radare2` sometimes introduces break changes. To specify `r2libr` please use something like `r2libr==5.7.0`.

qiling/extensions/r2/r2.py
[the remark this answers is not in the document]
elicn: Yes, you could load r2 with an empty buffer but you'll have to specify `-a` and `-b` for architecture and bits.
Shellcode content can be inserted with `wx`.

examples/extensions/r2/hello_r2.py
elicn: Agree; I think a comment saying "expecting _x_ to be equal to _y_" is more appropriate here.

setup.py
trufae: Breaking changes should only happen when the major and minor numbers change. only patch releases are abi stable. so if you have r2libr-5.7.0 you can use r2-5.7.0, but also 5.7.4 or 5.7.8 without runtime problems. That relaxes a little the stress to maintain updates and also makes it easy to transition between abi-breaking versions

examples/extensions/r2/hello_r2.py
chinggg: I think `offset` here means address. r2's command `aflj` use `offset` as the key name in JSON, and I don't find an elegant way to specify field name in dataclass.

qiling/extensions/r2/r2.py
chinggg: Yes. Now these classes have no special post-processing stuff. I use dataclass because it allows custom init function so I can pass all dict items got from r2 to construct the class without specifying fields.

Ref:
dataclass vs namedtuple: https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases
create namedtuple only use a subset of arguments: https://stackoverflow.com/questions/17622419/creating-a-namedtuple-object-using-only-a-subset-of-arguments-passed

qiling/extensions/r2/r2.py
chinggg: R2 has its own type of config variables, I think it's similar to qiling's profile, which is for different os though. We can extend it later.

qiling/extensions/r2/r2.py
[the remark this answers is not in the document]
chinggg: Thanks! I have implemented "shellcode mode" using command `wx HEX`.
But when I try to use the ctype API `libr.r_io.r_io_write_at(self._r2c.contents.io, loadaddr, code, size)`  it complains `ctypes.ArgumentError: argument 1: <class 'TypeError'>: expected LP_struct_r_io_t instance instead of LP_struct_r_io_t`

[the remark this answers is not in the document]
chinggg: Also there is currently no public util to accurately translate `QlArch` enum to arch and bits.

qiling/extensions/r2/r2.py
```diff
@@ -12,21 +12,27 @@
...
-class Function:
+class R2Data:
+    def __init__(self, **kwargs):
+        names = set([f.name for f in fields(self)])
+        for k, v in kwargs.items():
+            if k in names:
+                setattr(self, k, v)
+
+
+@dataclass(unsafe_hash=True)
```
elicn, line 24: There is no need to decorate `R2Data` subclasses with `@dataclass` anymore, they all inherit that from `R2Data`.

qiling/extensions/r2/r2.py
```diff
@@ -12,21 +12,27 @@
...
     offset: int
     size: int
     signature: str
 
     def __init__(self, **kwargs):
-        names = set([f.name for f in fields(self)])
-        for k, v in kwargs.items():
-            if k in names:
-                setattr(self, k, v)
+        super().__init__(**kwargs)
```
elicn, line 32: When all what the `__init__` method does is to call its `super` `__init__` method with no special processing, you could just omit it altogether. In other words: the subclasses `__init__` methods can be removed.

examples/extensions/r2/hello_r2.py
```diff
@@ -18,13 +19,27 @@ def func(ql: Qiling, *args, **kwargs):
...
+
+    # get function address and hook it
     ql.hook_address(func, r2.functions['main'].offset)
     ql.run()
 
 if __name__ == "__main__":
     my_sandbox(["rootfs/x86_windows/bin/x86_hello.exe"], "rootfs/x86_windows")
+
+    # test shellcode mode
+    ARM64_LIN = unhexlify('420002ca210080d2400080d2c81880d2010000d4e60300aa01020010020280d2681980d2010000d4410080d2420002cae00306aa080380d2010000d4210400f165ffff54e0000010420002ca210001caa81b80d2010000d4020004d27f0000012f62696e2f736800')
```
elicn, line 39: the `binascii` module is obsolete; you can use `bytes.fromhex` directly without the need of an extra module.

qiling/extensions/r2/r2.py
```diff
@@ -102,14 +103,35 @@ class SymbolBind(str, Enum):
 
 
 class R2:
-    def __init__(self, ql: Qiling):
+    def __init__(self, ql: Qiling, baseaddr=(1 << 64) - 1, loadaddr=0):
```
elicn, line 106: a base address of 0xffffffffffffffff ? are you sure..?

qiling/extensions/r2/r2.py
```diff
@@ -102,14 +103,35 @@ class SymbolBind(str, Enum):
...
+        libr.r_core.r_core_bin_load(self._r2c, path, self.baseaddr)
+        cmd = f'wx {self.ql.code.hex()}'
+        self._cmd(cmd)
+
+    def _setup_file(self):
+        path = self.ql.path.encode()
+        fh = libr.r_core.r_core_file_open(self._r2c, path, UC_PROT_READ | UC_PROT_EXEC, self.loadaddr)
+        libr.r_core.r_core_bin_load(self._r2c, path, self.baseaddr)
+
+    def _cmd(self, cmd: str) -> str:
```
elicn, line 134: Consider adding a matching `cmdj` method to handle `j`-suffixed commands, instead of calling `json.loads` every time.

qiling/extensions/r2/r2.py
chinggg: At first I only declare `__init__()` on the base class, but the fields of subclasses will not work. I don't know if there is anything special with dataclass.

examples/extensions/r2/hello_r2.py
chinggg: Just copy it from our example `shellcode_run.py`, will change it.

qiling/extensions/r2/r2.py
chinggg: I guess it's a default value hardcoded in r2 when calling `r_core_bin_load` but I am not sure. I have debugged functions `r_core_bin_load` and `r_core_file_open` to see the exact value of `baseaddr` and `loadaddr`. Maybe I should left some comments about them.

See https://github.com/radareorg/radare2/blob/master/libr/main/radare2.c#L427

wtdcode: > a base address of 0xffffffffffffffff ? are you sure..?

Yes, that's the default value indeed.

elicn: So `-1` is a special case there..?
I don't see any other explanation for picking this value as default.

qiling/extensions/r2/r2.py
```diff
@@ -0,0 +1,162 @@
...
+        r = libr.r_core.r_core_cmd_str(
+            self._r2c, ctypes.create_string_buffer(cmd.encode("utf-8")))
+        return ctypes.string_at(r).decode('utf-8')
+
+    def _cmdj(self, cmd: str) -> list[dict]:
```
elicn, line 131: `list` and `dict` type annotations became available starting from Python 3.9.
We officially (?) support 3.8, so we'll have to use `typing.List` and `typing.Dict` respectively.

setup.py
elicn: IMHO, I would refrain from defining something as a prerequisit unless it is absolutley required; that is, the program cannot operate without it. Plugging r2 into Qiling sounds like a great idea, where Qiling can benefit a lot from r2 static analysis.

Integrating r2 static analysis into Qiling requires some thinking of how to do it right and consistent across all modules. I really don't want it to be integrated into some modules and not in others. Until r2 is integrated across all Qiling modules, I believe we can wait with this prerequisit.

wtdcode: Yes, I understand your concern so let me illustrate my motivation a bit more:

1. Radare2 helps reduce redundant code and improves robustness, e.g. ELF loader, PE loader, even disassembler etc.
2. Indeed, I strongly agree with you that r2 should be, some sort of `everywhere` in Qiling. I hope to achieve this by using r2 to build a few basic utils or modules. An example I showed before is to identify function boundaries. Imagine you can `set_api` on a binary without any symbol tables, which is really useful and I don't think users will reject it.

Based on these two points, I think r2 will become a prerequisite someday but also I agree we should wait until we really have to do so.

chinggg:
Have we reached an agreement about how to integrate r2? Or we need a PoC of using r2 everywhere in qiling at first?

elicn:
> Have we reached an agreement about how to integrate r2? Or we need a PoC of using r2 everywhere in qiling at first?

IMHO, r2 should be optional for now, and not a requirement (i.e. revert the last commit).

chinggg:
I am trying to implement a loader of `minidump` using r2. If r2 can help load different file formats, it may be considered as a requirement.

qiling/extensions/r2/r2.py
```diff
@@ -121,6 +121,11 @@ def _cmd(self, cmd: str) -> str:
 
     def _cmdj(self, cmd: str) -> List[Dict]:
         return json.loads(self._cmd(cmd))
+    
+    def read(self, addr: int, size: int) -> bytes:
+        self._cmd(f"s {addr}")
+        hexstr = self._cmd(f"p8 {size}")
```
elicn, line 127: I would have consolidated the two r2 commands into one: `self._cmd(f"p8 {size} @ {addr}")`

elicn:
> I am trying to implement a loader of `minidump` using r2. If r2 can help load different file formats, it may be considered as a requirement.

As mentioned above, till Qiling becomes dependent of radare2 (by design, not opportunistically), radare2 should not appear as a required dependency. If someone wants to analyze a specific file format that its loading relies on radare2, they can install radare2 for that purpose (similar to EVM and its dependencies).

elicn:
Leveraging radare2 static analysis capabilities to enhance Qiling emulation is a great idea, especially strings and flags. I implemented a similar resolving capabilities to my enhanced `trace` module [see [here](https://twitter.com/EliCohenNehemia/status/1380464269272186882/photo/1)], but that relied on an address-to-symbol mapping which I got by parsing map files. Back then I was thinking about enabling PDB and DWARF parsing in Qiling, but never got to it..

All that being said, it is esential to keep in mind that radare2, as powerful as it might be, is a _static analysis_ tool , whereas Qiling _dynamically_ emulates code. Though some of radare2 analysis may be useful in a dynamic environment, its analysis may break or become irrelevant when dealing with dynamically loaded libraries, self modifying code, and basically everything that is not static.

My understanding is that radare2 may be useful to extract debugging information and strings, but I am not sure about the rest. We'll still have to load dynamic libraries, resolve imports, etc.

Like I mentioned in the past, `QlOS` and `QlLoader` are a mess and they need to be re-designed to better support processes, threads, etc. If there is a design plan behind `minidump` along with design changes to the Loader, it would be great if you could share it to get feedback. Please note that linking to an existing design might not be sufficient (ex: "see pwntols loader"); we need to see how it is going to be plugged into Qiling with all necessary adjustments.

chinggg:
@wtdcode I have made a simple PoC of address to flag mapping (by calling `at` function instead of mapping every address). 
But I don't fully understand using xref to `view current backtrace`, I think It does not mean to give possible backtraces (cannot be accurate) statically since we want to bridge it with qiling instead of just static analysis.
I cannot find existing `backrace` in qiling, I guess it maybe a kind of code hook. Currently I am not sure about the implementation.

qiling/extensions/r2/r2.py
```diff
@@ -0,0 +1,227 @@
...
+    def refrom(self, addr: int) -> Optional[Xref]:
+        return self.xrefs.get(addr)
+
+    def refto(self, addr: int) -> List[Xref]:
+        return [xref for xref in self.xrefs.values() if xref.addr == addr]
+
+    def enable_trace(self, mode='full'):
+        # simple map from addr to flag name, cannot resolve addresses in the middle
+        self.ql.loader.symsmap = {flag.offset: flag.name for flag in self.flags}
+        if mode == 'full':
```
wtdcode, line 209: Use an enumeration instead.

chinggg: This is just a PoC, the trace extension made by @elicn need a `symsmap` to resolve addresses, I have to made further modification on the trace extension.

qiling/extensions/trace.py
```diff
@@ -2,14 +2,16 @@
 
 # More info, please refer to https://github.com/qilingframework/qiling/pull/765
 
+from __future__ import annotations
```
elicn, line 5: I don't think there is a circular dependency between `trace` and `Qiling`, so I don't think this is actually needed here.

qiling/extensions/r2/r2.py
```diff
@@ -111,25 +115,33 @@ def __lt__(self, other):
...
 
     def _setup_code(self, code: bytes):
         path = f'malloc://{len(code)}'.encode()
         fh = libr.r_core.r_core_file_open(self._r2c, path, UC_PROT_ALL, self.loadaddr)
         libr.r_core.r_core_bin_load(self._r2c, path, self.baseaddr)
         self._cmd(f'wx {code.hex()}')
         # set architecture and bits for r2 asm
-        self._cmd(f"e,asm.arch={self.ql.arch.type.name.lower().removesuffix('64')},asm.bits={self.ql.arch.bits}")
+        try:
+            self._cmd(f"e,asm.arch={self.ql.arch.type.name.lower().removesuffix('64')},asm.bits={self.ql.arch.bits}")
```
elicn, line 142: `removesuffix` was introduced in Python 3.9 and we officially support 3.8.
That said, I think this is a hacky and instable way to map Qiling arch names to Radare2 arch names.

qiling/core.py
```diff
@@ -181,6 +183,10 @@ def __init__(
         if self.baremetal:
             self._hw = select_component('hw', 'hw')(self)
 
+        # load r2 extension
+        if enable_r2:
+            self.r2 = R2(self)
```
elicn, line 188: I think integrating r2 into Qiling at this point is a **very bad idea**.
This should stay as extention, and by definition extention cannot be an integral part of the core - but only an optional / experimental feature.

qiling/arch/utils.py
```diff
@@ -53,6 +53,11 @@ def disassembler(self, ql: Qiling, address: int, size: int):
 
         for insn in ql.arch.disassembler.disasm(data, address):
             offset = insn.address - ba
+            try:
+                # use r2 to get fine-grained name and offset
+                name, offset = ql.r2.resolve(insn.address)
```
elicn, line 58: As mentioned elsewhere, r2 should not be an integral part of core, but assuming that it is - this is an ineffective way to do this. Better rely on the r2 enable flag rather than set an excpetion frame.

elicn (changes requested):
Turning r2 into an integral part of core is a bad idea, in my opinion.
As an extention, r2 analysis should stay an external add-on feature that may [or may not] be used with Qiling.

qiling/arch/utils.py
wtdcode: "this is an ineffective way to do this" is not correct. Note r2 flags have more information than we have for now. For example, Qiling can't retrieve cross reference by any means for now.

qiling/core.py
wtdcode: I agree that it shouldn't be an option in the constructor but instead I hope to make it used everywhere, with the higher level information obtained from r2.

qiling/arch/utils.py
chinggg: I think "ineffective" refers to the `try ... except` block. Anyway, if `r2` can be integrated into qiling, I think it's better to be a `property` so we can construct `R2` at need without specifying it at first.

qiling/core.py
chinggg: Yeah in this PR I also want to keep r2 as an extension without any modification of qiling core.

qiling/extensions/r2/r2.py
```diff
@@ -0,0 +1,241 @@
...
+        self._r2c = libr.r_core.r_core_new()
+        if ql.code:
+            self._setup_code(ql.code)
+        else:
+            self._setup_file(ql.path)
+
+    def _qlarch2r(self, archname: str) -> str:
+        archname = archname.lower()
+        # str.removesuffix() is not available until Python 3.9
+        return archname[:-2] if archname.endswith('64') else archname
```
wtdcode, line 137: Why not do a mapping from enum to r2 arch type?

chinggg: QlArch
```python
class QL_ARCH(IntEnum):
    X86 = 101
    X8664 = 102
    ARM = 103
    ARM64 = 105
    MIPS = 106
    A8086 = 107
    EVM = 108
    CORTEX_M = 109
    RISCV = 110
    RISCV64 = 111
    PPC = 112
```

r2asm
```
[0x00000000]> e asm.arch=
bf            6502          8051          amd29k        arm.as        
arm           arm.winedbg   avr           cr16          dalvik        
dcpu16        gb            h8300         i4004         i8080         
java          lh5801        lm32          m680x         malbolge      
mcore         mcs96         mips          nios2         or1k          
pic           ppc.as        ppc           propeller     riscv         
rsp           sh            snes          sparc         s390          
tms320        tms320c64x    tricore       v810          v850          
vax           ws            x86.as        x86           x86.nasm      
x86.nz        xap           xcore         arm.gnu       cris          
m68k.gnu      lanai         mips.gnu      ppc.gnu       sparc.gnu     
s390.gnu      v850.gnu
```

Only `A8086`, `EVM`, `CORTEX_M` are not seen in r2, the others can all be converted by removing `64` suffix.

wtdcode:
@elicn I think this PR is fine to go and we would like another a few PRs about R2 application in Qiling internally.

wtdcode (approved):

elicn:
Looks good to me.
Can we just remove the `r2libr` from the dependencies and make it optional, like it used to be?

elicn (approved):

PR #1173, opened 2022-06-14
Maintenance PR

elicn:
Maintenance PR for security and code quality.

Highlights:
- Safer path semantics
  - Provide the emulated program with virtual paths rather than host paths, when possible (WIP)
  - Paths accessed by the emulated program must end up under rootfs to prevent messing up the host file system [that takes symlinks into considerations to prevent path traversals]
  - The emulated program must reside somewhere in rootfs
- Revised POSIX `filestruct` and `socket` modules for a better code quality
- Comments and typing annotations

### DIFF
--- a/qiling/os/posix/syscall/unistd.py
+++ b/qiling/os/posix/syscall/unistd.py
@@ -338,23 +338,25 @@ def ql_syscall_write(ql: Qiling, fd: int, buf: int, count: int):
 def ql_syscall_readlink(ql: Qiling, path_name: int, path_buff: int, path_buffsize: int):
     pathname = ql.os.utils.read_cstring(path_name)
     # pathname = str(pathname, 'utf-8', errors="ignore")
-    real_path = ql.os.path.transform_to_link_path(pathname)
-    relative_path = ql.os.path.transform_to_relative_path(pathname)
+    host_path = ql.os.path.virtual_to_host_path(pathname)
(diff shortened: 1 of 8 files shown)


qiling/os/posix/syscall/unistd.py
```diff
@@ -338,23 +338,25 @@ def ql_syscall_write(ql: Qiling, fd: int, buf: int, count: int):
...
-    real_path = ql.os.path.transform_to_link_path(pathname)
-    relative_path = ql.os.path.transform_to_relative_path(pathname)
+    host_path = ql.os.path.virtual_to_host_path(pathname)
+    virt_path = ql.os.path.virtual_abspath(pathname)
 
-    if not os.path.exists(real_path):
-        regreturn = -1
+    # cover procfs psaudo files first
+    # TODO: /proc/self/root, /proc/self/cwd
+    if virt_path == r'/proc/self/exe':
```
wtdcode, line 346: Maybe a separate file for such things instead of “hacky if”?

elicn: I agree this is ~hacky~ ugly, but that `if` was already there..
The elegant way to handle this would be a fully virtualized file system that could handle symlinks like this one.

elicn:
My local branch got toasted.. I'll open another PR instead.

PR #1174, opened 2022-06-14
Linux procfs PoC

elicn:
Introduce procfs to Linux OS
This is just a minimal PoC; looking for feedback.

Notes:
- The implementation of `/proc/self/exe` just mimics a symlink behvaior, and could be generalized if we had a full virtualized file system
- As long as Qiling doesn't have a full virtualized file system, it would be hard to maintain both "real" files and pseudo files seamelessly
- This somewhat repeats code patterns from `ql_file` and `QlFsMappedObject`, but till those modules are appropriately generalized, I figured it would be cleaner this way

### DIFF
--- a/qiling/debugger/gdb/gdb.py
+++ b/qiling/debugger/gdb/gdb.py
@@ -475,23 +475,15 @@ class QlGdb(QlDebugger):
                     return f'l{content}'
 
                 elif feature == 'auxv' and op == 'read':
-                    auxv_data = bytearray()
+                    try:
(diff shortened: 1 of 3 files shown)


wtdcode:
Regarding the repeat pattern between `ql_file` and `QlFsMappedObject`, the original motivation mostly is backward compatibility since `ql_file` was the only choice for users at that time and I didn’t want to break everything. Maybe we can safely merge the two base class nowadays.

elicn:
Guess there is no more further feedback; ready for review.

cla7aye15I4nd (approved):

PR #1195, opened 2022-07-19
Maintenance PR

### LINKED ISSUE #1192: Error loading kernel module nfnetlink.ko
***Describe the bug**
When executing a load to emulate the nfnetlink.ko kernel module. An error occurs. Presumably due to passing in the lkm_dynlinker function as a string 'SHN_ABS' rather than a numeric parameter (_section = elffile.get_section(_symbol['st_shndx']))

**Sample Code**
```
#!/usr/bin/python3

import sys
from qiling import *

def test(path, rootfs):
    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEBUG)

    print(f"ql.loader.load_address: {ql.loader.load_address}")
    ql.run(ql.loader.load_address + 0x1205, ql.loader.load_address + 0x126c)  

if __name__ == "__main__":
    test(["nfnetlink.ko"], 
    "rootfs/x8664_linux")
```

**Expected behavior**

**Screenshots**
изображение

**Kernel module**
[nfnetlink.zip](https://github.com/qilingframework/qiling/files/9101949/nfnetlink.zip)

elicn:
Highlights:
- Increase host security around Windows file access API
- Have some tests clean up file system residuals
- Remove obsolete settings from profiles
- Fixes #1192


qiling/os/windows/const.py
```diff
@@ -12,12 +12,16 @@
 ERROR_PATH_NOT_FOUND = 0x3
 ERROR_ACCESS_DENIED = 0x5
 ERROR_INVALID_HANDLE = 0x6
+ERROR_GEN_FAILURE = 0x1f
```
kabeor, line 15: I think here should be capitalized?

elicn: I don't think we have a consistent convention for that (there are many lowercase and upercase constans everywhere), neither PEP8 defines it. Is there are specific reason here?

kabeor (approved):

qiling/os/windows/const.py
kabeor: No, I just saw the variables that existed before is all uppercase values.

xwings (approved):

wtdcode (approved):

PR #1196, opened 2022-07-20
feat(r2): human-friendly disassembler

chinggg:

### Coding convention?

- [] The reference of the new code is pointed out.

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This is part of the further work discussed in #1172 
`r2` has `pD` to disassemble N bytes and `pd` to disassemble N instructions directly. Though the latter may be used more often, counting bytes is more compatible with the callback type `ql.hook_code()` expects, like `QlArchUtils.disassembler`.

Now there is no change in qiling core, so the code looks a bit ugly. If we can integrate r2 into qiling, the feature can be implemented with only a few lines and no attention of users.

### DIFF
--- a/examples/extensions/r2/hello_r2.py
+++ b/examples/extensions/r2/hello_r2.py
@@ -16,7 +16,7 @@ def func(ql: Qiling, *args, **kwargs):
     return
 
 def my_sandbox(path, rootfs):
-    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEBUG)
+    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEFAULT)
     r2 = R2(ql)
 
     # search bytes sequence using ql.mem.search
@@ -33,7 +33,8 @@ def my_sandbox(path, rootfs):
     # get function address and hook it
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
-    r2.enable_trace()
+    # r2.enable_trace()
+    r2.enable_disasm()
     ql.run()
 
 if __name__ == "__main__":
--- a/examples/extensions/r2/hello_r2.py
+++ b/examples/extensions/r2/hello_r2.py
@@ -34,8 +34,7 @@ def my_sandbox(path, rootfs):
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
     # r2.enable_trace()
-    # selectively disasm by regex searching flag name
-    r2.enable_disasm('main')
+    r2.enable_disasm()  # monky-patched disasm powered by r2
     ql.run()
 
 if __name__ == "__main__":
--- a/examples/extensions/r2/hello_r2.py
+++ b/examples/extensions/r2/hello_r2.py
@@ -16,7 +16,8 @@ def func(ql: Qiling, *args, **kwargs):
     return
 
 def my_sandbox(path, rootfs):
-    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEFAULT)
+    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DISASM)
+    # QL_VERBOSE.DISASM will be monkey-patched when r2 is available
     r2 = R2(ql)
 
     # search bytes sequence using ql.mem.search
@@ -34,7 +35,6 @@ def my_sandbox(path, rootfs):
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
     # r2.enable_trace()
-    r2.enable_disasm()  # monky-patched disasm powered by r2
     ql.run()
 
 if __name__ == "__main__":
--- a/examples/extensions/r2/hello_r2.py
+++ b/examples/extensions/r2/hello_r2.py
@@ -34,7 +34,8 @@ def my_sandbox(path, rootfs):
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
     # r2.enable_trace()
-    r2.enable_disasm()
+    # selectively disasm by regex searching flag name
+    r2.enable_disasm('main')
     ql.run()
 
 if __name__ == "__main__":
--- a/.github/workflows/build-ci.yml
+++ b/.github/workflows/build-ci.yml
@@ -80,7 +80,7 @@ jobs:
         cd ../qiling
         cd ../examples/rootfs/x86_linux/kernel && unzip -P infected m0hamed_rootkit.ko.zip
         cd ../../../../
-        pip3 install -e .[evm]
+        pip3 install -e .[evm,RE]
 
         if [ ${{ matrix.os }} == 'ubuntu-18.04' ] and [ ${{ matrix.python-version }} == '3.9' ]; then
           docker run -it --rm -v ${GITHUB_WORKSPACE}:/qiling qilingframework/qiling:dev bash -c "cd tests && ./test_onlinux.sh"
--- a/qiling/arch/utils.py
+++ b/qiling/arch/utils.py
@@ -73,7 +73,12 @@ class QlArchUtils:
             self._block_hook = None
 
         if verbosity >= QL_VERBOSE.DISASM:
-            self._disasm_hook = self.ql.hook_code(self.disassembler)
+            try:  # monkey patch disassembler
+                from qiling.extensions.r2 import R2
+                r2 = R2(self.ql)
+                self._disasm_hook = self.ql.hook_code(r2.disassembler)
+            except (ImportError, ModuleNotFoundError):
+                self._disasm_hook = self.ql.hook_code(self.disassembler)
 
             if verbosity >= QL_VERBOSE.DUMP:
                 self._block_hook = self.ql.hook_block(ql_hook_block_disasm)
--- a/qiling/extensions/r2/r2.py
+++ b/qiling/extensions/r2/r2.py
@@ -10,7 +10,7 @@ import re
 import libr
 from dataclasses import dataclass, field, fields
 from functools import cached_property, wraps
-from typing import TYPE_CHECKING, Dict, List, Literal, Pattern, Tuple, Union
+from typing import TYPE_CHECKING, Dict, List, Literal, Optional, Pattern, Tuple, Union
 from qiling.const import QL_ARCH
 from qiling.extensions import trace
 from unicorn import UC_PROT_NONE, UC_PROT_READ, UC_PROT_WRITE, UC_PROT_EXEC, UC_PROT_ALL
@@ -243,6 +243,18 @@ class R2:
         # minus 1 to find the corresponding flag
         flag = self.flags[idx - 1]
         return flag, addr - flag.offset
+
+    def where(self, name: str, offset: int=0) -> Optional[int]:
+        '''Given a name (+ offset), return its address or None'''
+        if '+' in name:  # func_name + offset
+            name, offset = name.split('+')
+            name = name.strip()
+            try:
+               offset = int(offset.strip(), 0)
+            except ValueError:
+                pass
+        func = self.functions.get(name)
+        return func.offset + offset if func else None
 
     def refrom(self, addr: int) -> List[Xref]:
         return [x for x in self.xrefs if x.fromaddr == addr]
--- a/qiling/extensions/r2/r2.py
+++ b/qiling/extensions/r2/r2.py
@@ -88,6 +88,7 @@ class Instruction(R2Data):
     opcode: str  # raw opcode
     disasm: str = ''  # flag resolved opcode
     bytes: bytes
+    type: str
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -269,12 +270,22 @@ class R2:
         insts = [Instruction(**dic) for dic in self._cmdj(f"pDj {size} @ {addr}")]
         return insts
 
-    def disassembler(self, ql: 'Qiling', addr: int, size: int, filt: Pattern[str]) -> None:
+    def disassembler(self, ql: 'Qiling', addr: int, size: int, filt: Pattern[str]=None) -> int:
+        '''A human-friendly disassembler powered by r2, can be used for hook_code
+            :param ql: Qiling instance
+            :param addr: start address for disassembly
+            :param size: size in bytes
+            :param filt: regex pattern to filter instructions
+            :return: progress of dissembler, should be equal to size if success
+        '''
         anibbles = ql.arch.bits // 4
         for inst in self.dis_nbytes(addr, size):
+            if inst.type.lower() == 'invalid':
+                return inst.offset - addr # stop disasm and return address progress
             flag, offset = self.at(inst.offset)
-            if filt.search(flag.name):
+            if filt is None or filt.search(flag.name):
                 ql.log.info(f'{inst.offset:0{anibbles}x} [{flag.name:20s} + {offset:#08x}] {inst.bytes.hex(" "):20s} {inst.disasm}')
+        return size
 
     def enable_disasm(self, filt_str: str=''):
         filt = re.compile(filt_str)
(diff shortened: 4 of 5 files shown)


elicn:
I am not sure why do we need disassembling capabilities based on radare2. Given that radare2 itself relies on Capstone for disassembly, I don't see a clear advantage here (unless we strive to base as many things as we can on top of radare2..?)

wtdcode:
> > We strive to base as many things as we can on top of radare2, not limited to symbol resolving
> 
> Where did that come from..? Not sure that I agree with this.
> 
> Anyway, this is indeed a nice disassembly feature. The only concern I can think of here is how a user may tweak radare2 analysis in order to get more accurate info where radare2 guesses wrong.

Mostly from cross-reference analysis.

elicn:
> Mostly from cross-reference analysis.

I am not following.

chinggg:
In the latest commits, I integrate r2 extension into qiling core again, but as `cached_property`. To demonstrate the advantage of introducing r2, I modify the breakpoint part of qdb, now it can set breakpoints using `main+3` instead of integer addresses only. 

image

The tests fail since `libr` is sitll optional dependency.

chinggg:
It seems I ignore the fact that r2 can only provide disasm for binary itself, but not the libraries, so it should fallback to the existing disassembler approach when disasm is not valid.

wtdcode:
You can try to create multiple r2 instances for libraries.

chinggg:
@wtdcode I have tried to use r2 on `msvcrt.dll`, it's extremely slow. I must wait for 10s+ for r2 to analyze the library. Moreover, since the library does not provide debug info, it's also a bit useless to see things like `fcn.0x40abcd`. So I make r2 return the address for disasembler to continue disasm manually.

Anyway, multiple r2 instances for libraries can be supported later if it is useful.

wtdcode:
> @wtdcode I have tried to use r2 on `msvcrt.dll`, it's extremely slow. I must wait for 10s+ for r2 to analyze a single library. Moreover, since the library does not provide debug info, it's also a bit useless to see things like `fcn.0x40abcd`. So I make r2 return the address for disasembler to continue disasm manually.
> 

`fcn.0x40abcd` is also useful in many cases instead of a numeric address, say you are moving it as a function pointer:

```
mov rax, fcn.0x40abcd
```

instead of 

```
mov rax, 0x40abcd
```

also this would be helpful if we would like a CFG.

For the load time, we could 

- analyze less on Windows, say `aa` instead of `aaaa`
- I'm not sure how much time it takes to load `msvcrt.dll` into unicorn. If that already takes a long time, I don't think we bother a bit longer.
- If the time is indeed too long for users, we can disable this feature on Windows emulation by default but enable it on other systems.

elicn:
Note that you cannot analyze the binaries and libraries from disk, because they might have been patched by Qiling (`ql.patch`). You should communicate a patched version to radare2.

chinggg:
@wtdcode @elicn  Thanks for mentioning **multiple instances** and **patched binaries**. I think we can solve it later but not here, since we really have to take many things into consideration, such as code organization and consistency, handling of extremely slow corner cases (`aa` libc.so takes me 10s+ on my PC, crafted binaries may hang for hours) and showcase examples.

This PR focuses on **enhanced disassembly**, and I want to get it merged quickly. Since it contains not only feature enhancement, but also bug fixes.

Now `QlArchUtils.disassembler` is smoothly enhanced by r2, you can test it with `hello_r2` example, it should print same amount of disassembly. So I think time is ripe to **make r2 required dependency**, or the tests will always fail.

elicn:
Wait. Let's not jump the gun here. There are few considerations I would like us all to take into account:
1. Implementing an extension is one thing, and integrate that into core and make it a dependency is totally different. Qiling does not widely rely on radare2 to perform analysis (yet). Moving toward this direction would require a lot of thinking and design decisions. Until we do, there is no justification to make radare2 an integral dependency for Qiling.
2. Though the enhanced disassembly is a cool feature, I don't think it is ready. There are too many "corner cases" which fail the analysis. Patched binaries are one, but self-modifying code and running code from stack (like every other exploit out there does) would also fail. This is exactly where _static_ meets _dynamic_, and we have to come up with the right design to work around that.
3. Qiling has too many "half baked" features that somebody introduced to the project once, and then left them there. Most of my efforts are dedicated to fixing [half-]broken implementations rather than introducing new features, so I must say I am very much against introducing another half-working feature (no matter how cool it is). Taking the time to think and design a robust solution is crucial.

All that being said, my opinion is that the radare2-based analysis should appear as an *extension* until we figure out how to leverage it properly. People would still be able to develop features on top of that extension (e.g. the enhanced disassembly) without turning it into a core feature. Once we figure out how to harness radare2 analysis and work around the issues, we would make it a core feature.

wtdcode:
> Wait. Let's not jump the gun here. There are few considerations I would like us all to take into account:
> 
> 1. Implementing an extension is one thing, and integrate that into core and make it a dependency is totally different. Qiling does not widely rely on radare2 to perform analysis (yet). Moving toward this direction would require a lot of thinking and design decisions. Until we do, there is no justification to make radare2 an integral dependency for Qiling.

The final goal of the project is indeed to make r2 being used fluently in Qiling. Say, we no long rely on various magic addresses or offsets, instead we will have a good reference system (provided by r2).

> 2. Though the enhanced disassembly is a cool feature, I don't think it is ready. There are too many "corner cases" which fail the analysis. Patched binaries are one, but self-modifying code and running code from stack (like every other exploit out there does) would also fail. This is exactly where _static_ meets _dynamic_, and we have to come up with the right design to work around that.

I agree with this point. @chinggg Last time I talked about this, maybe we really have to be able to sync r2 memory with Qiling (Unicorn) memory somehow. Pancake also once talked about it with me, a clean way is to write a new `r_io_plugin` for unicorn using the `uc_mem_map_ptr`. But for now, I think manually sync the memory is an acceptable choice.

> 3. Qiling has too many "half baked" features that somebody introduced to the project once, and then left them there. Most of my efforts are dedicated to fixing [half-]broken implementations rather than introducing new features, so I must say I am very much against introducing another half-working feature (no matter how cool it is). Taking the time to think and design a robust solution is crucial.
> 
> All that being said, my opinion is that the radare2-based analysis should appear as an _extension_ until we figure out how to leverage it properly. People would still be able to develop features on top of that extension (e.g. the enhanced disassembly) without turning it into a core feature. Once we figure out how to harness radare2 analysis and work around the issues, we would make it a core feature.

I agreed with this statement, but not fully. We should be careful about new features indeed but there will be tradeoff. Being an extension means not every user could notice and use it and then provide enough feedback. While introducing r2 means a very big conceptional change (remember we now have CFG, we can use flags instead of magic addresses), this needs more feedback and being an extension make it _hard_ to understand and use.

chinggg:
I agree that we really need well-design decisions when integrating r2 into qiling. As @wtdcode mentioned, syncing r2 memory with Qiling (Unicorn) memory is important for building our flag-address system, maybe we need to manipulate r2 binary manually.

> Qiling has too many "half baked" features that somebody introduced to the project once, and then left them there. Most of my efforts are dedicated to fixing [half-]broken implementations rather than introducing new features

Thanks for your efforts to keep the project robust and healthy! Maybe `qdb` is just one of the "half baked" features. I even wonder why it is not an external extension :joy_cat: Now I have made some commits to enhance qdb by utilizing flag-address ability provided by r2. Can you imagine that `qdb` only supports `b 0x400120` but not `b main+3`? But with r2 integrated, the latter can be implemented with just a few lines of code! (I admit that's not always the case, we will pay for the lack of caution). It's not easy to do so if `r2` is not part of qiling.

elicn:
>  agreed with this statement, but not fully. We should be careful about new features indeed but there will be tradeoff. Being an extension means not every user could notice and use it and then provide enough feedback. While introducing r2 means a very big conceptional change (remember we now have CFG, we can use flags instead of magic addresses), this needs more feedback and being an extension make it hard to understand and use.

radare2 support is still an experimental feature, and as such it should stay out of core. Making it an extension would be a great solution: it's there for everyone to experiment with, but also can be avoided if someone doesn't feel experimental and prefer stability.

You two try to convience me that radare2 analysis is needed and has great potential, and I'm not sure why: I already said it myself. I don't think we shouldn't have it, all I'm saying is that it is curently immature and needs to go through more thinking and deveopment before we can integrate radare2 analysis into Qiling.

For now the radare2-based analysis is a nice PoC; not something we can integrate to core.

chinggg:
Now r2 is sill **optional dependency** and **partially** integrated into qiling core only if `r2libr` is installed. Users can use qiling as usual if they don't install `r2libr`.

wtdcode:
> > agreed with this statement, but not fully. We should be careful about new features indeed but there will be tradeoff. Being an extension means not every user could notice and use it and then provide enough feedback. While introducing r2 means a very big conceptional change (remember we now have CFG, we can use flags instead of magic addresses), this needs more feedback and being an extension make it hard to understand and use.
> 
> radare2 support is still an experimental feature, and as such it should stay out of core. Making it an extension would be a great solution: it's there for everyone to experiment with, but also can be avoided if someone doesn't feel experimental and prefer stability.
> 
> You two try to convience me that radare2 analysis is needed and has great potential, and I'm not sure why: I already said it myself. I don't think we shouldn't have it, all I'm saying is that it is curently immature and needs to go through more thinking and deveopment before we can integrate radare2 analysis into Qiling.
> 
> For now the radare2-based analysis is a nice PoC; not something we can integrate to core.

Okay, make sense and I think we are sync-ed here. The latest commit will detect the ImportError which I think is good to go.

elicn:
To keep a clean code, this feature needs to be **completely detached** from core - just like other experimental features, that are not plugged-in to core at all. To let it take over the disasemble method, the module can monkeypatch the utils module once it is imported, instead of `try` / `catch` each and every asembly instruction.

chinggg:
Already detach r2 extension from qiling core, r2.disassembler now monkey patch `QlArchUtils.disassembler`

elicn:
That looks better, thanks.
My suggestion to monkeypatch the disassemble method meant to replace the default disasembly hook that is set when the user sets the verbosity level to `QL_VERBOSE.DISASM`. That could have take place when the new r2 module imported, and provide a transparent way to go with the new feature. 

Anyway, the current method is fine too.

chinggg:
Since we have approved the current change, can we merge this PR and move quickly for further work like backtrace, xref, CFG and memory consistency?

elicn:
@wtdcode, do you approve it with the latest chances?

wtdcode:
> @wtdcode, do you approve it with the latest chances?

Sorry too busy these days, will check today later.

examples/extensions/r2/hello_r2.py
```diff
@@ -33,7 +33,8 @@ def my_sandbox(path, rootfs):
     # get function address and hook it
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
-    r2.enable_trace()
+    # r2.enable_trace()
+    r2.enable_disasm()  # monky-patched disasm powered by r2
```
wtdcode, line 37: Why not follow eli's advice here?

> My suggestion to monkeypatch the disassemble method meant to replace the default disasembly hook that is set when the user sets the verbosity level to QL_VERBOSE.DISASM. That could have take place when the new r2 module imported, and provide a transparent way to go with the new feature.

chinggg: > That could have take place when the new r2 module imported, and provide a transparent way to go with the new feature. Anyway, the current method is fine too.

I cannot figure a smooth way to call r2 disassembler by setting the verbosity level to `QL_VERBOSE.DISASM` when r2 is not imported by qiling's core.

wtdcode: > > That could have take place when the new r2 module imported, and provide a transparent way to go with the new feature. Anyway, the current method is fine too.
> 
> I cannot figure a smooth way to call r2 disassembler by setting the verbosity level to `QL_VERBOSE.DISASM` when r2 is not imported by qiling's core.

I think @elicn means something like:

```python
if verbose == QL_VERBOSE.DISASM:
    # ....
    try:
        from qiling.extension.code import r2
    except ImportError:
        # Monkey patch
```

chinggg: Done. Now when `setup_output` is called (mostly during the initialization of qiling), a instance of R2 will be created, though r2 is not part of qiling's core (`ql.r2` is not possible now)

elicn: Ah, that is not exactly what I meant..
I meant to replace the `disassembler` method from _outside_, to keep Qiling "clean" from any radare2 references. You can replace the disassembler instance method with something like:
```python
ql.arch.utils.disassembler = another_method.__get__(ql.arch.utils)
```
That can be called within r2 module, and executed when it is loaded. You should not worry about whether the original `disasembler` method was already used for a hook. Just set the verbosity again to itself:
```python
ql.verbose = ql.verbose
```
so `setup_output` will be called and re-install the hook with the new disasembly method.

wtdcode: Then how could user enable r2 after r2 installed?

elicn: I don't understand the question.. can you be more specific?

wtdcode: If there is no import in core, how could monkey patch work? Or users have to import…? That doesn’t make too much sense as no one will really write an extra line…

elicn: Why not..? If someone wants to enable radare2-based disasembly, they should explicitly import the extension - just like any other extension. How is that different from the `trace` extension?

wtdcode: > Why not..? If someone wants to enable radare2-based disasembly, they should explicitly import the extension - just like any other extension. How is that different from the `trace` extension?

It's different. `trace` is more of a tool but `r2` will provide many APIs in parallel with some existing Qiling API. Users would always prefer an `out-of-box` solution and we really need some further feedback on r2 integration. I really don't think a few import lines in our files could make so much difference...

elicn: Are we talking about the disasembly functionality, or is this a wider discussion on radare2-based analysis..?
If the former, than this is just like any other tool - like you said. Anyway, this is only my opinion..

Just thinking out loud here: would it make more sense to create a dedicated branch for radare2-assisted Qiling, to let people give it a spin?

wtdcode: > Are we talking about the disassembly functionality, or is this a wider discussion on radare2-based analysis..?
If the former, than this is just like any other tool - like you said. Anyway, this is only my opinion..

> Just thinking out loud here: would it make more sense to create a dedicated branch for radare2-assisted Qiling, to let people give it a spin?

I refer to the wider discussion and I understand your concern. However, I think `pip install qiling[RE]` already provides a "dedicated branch" to a certain extent (thinking of how many users install EVM extensions till today...).

chinggg: @elicn Do you mean a dedicated branch in GitHub where I continuously add new features? Now I can only limited new work until the previous PR get merged since they are not totally independent.

elicn: I suggested to create a separated dev branch for radare2-assisted Qiling, where radare2 can be integrated without worrying about breaking anything or messing up the code. This is just a continuation of a previous discussion, in which we talked about whether to leave radare2 stuff out of core until Qiling gets re-architected to harness radare2 analysis.

All that being said, I approved this PR a few days ago and as far as I am concerned - it can be merged.

chinggg: Sure, I have made two more new commits including refactor and test to guarantee the API is stable, then this PR can be merged.

wtdcode (approved):

chinggg:
@wtdcode You approve the changes and close this PR, so you want me to work on a dedicated branch and submit a big PR?

elicn:
What happened..? I am wondering the same thing..

wtdcode:
@chinggg @elicn My bad, I intended to merge but close it wrongly. Sorry to busy these days.

PR #1197, opened 2022-07-21
Updates qltool to provide an interactive UI

Rohan-cod:
# Description
Used `PyInquirer` module to build an **interactive user interface** for `qltool` as opposed to the pure text-based interface of `argparse`.

### DIFF
--- a/qltool
+++ b/qltool
@@ -3,11 +3,25 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
-import argparse
 import os
-import sys
 import ast
 import pickle
+import re
+import click
+import six
+import json
+
+from PyInquirer import (Token, ValidationError, Validator, print_json,
+                        prompt, style_from_dict)
+from pyfiglet import figlet_format
+from rich import print
+from rich.console import Console
+from rich.table import Table
+from rich import box
+try:
+    from termcolor import colored
+except ImportError:
+    colored = None
 
 from unicorn import __version__ as uc_ver
 from qiling import __version__ as ql_ver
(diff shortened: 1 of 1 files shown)


aquynh:
Cool, can you post a screenshot?

Rohan-cod:
> Cool, can you post a screenshot?

Sure

Screenshot 2022-07-21 at 6 33 55 PM

elicn:
I am not familiar with this particular module; when does the TUI show up? only when executed without command line arguments (so it enters a "wizard mode")? or unconditionally?

I'm asking because I would like to maintain the possibility to run `qltool` without a TUI (like it was until this modification).

Rohan-cod:
> I am not familiar with this particular module; when does the TUI show up? only when executed without command line arguments (so it enters a "wizard mode")? or unconditionally?

This program doesn't accept any arguments. It's as you mentioned is only a TUI. When you run the program the UI will show up.

> I'm asking because I would like to maintain the possibility to run `qltool` without a TUI (like it was until this modification).

If you want to keep this functionaltiy provided via argparse, then maybe we can have a separate file for this, `qltool` for the program built using argparse and `qltui` for the TUI program? OR We can have them both in the same file and trigger the TUI based on the the length of the command line arguments, i.e if no command line arguments are supplied, show the TUI, if they are supplied, use argparse?

elicn:
I think that the best way would be separating the backend from its UI, to align with the GNU approach, but that is only my opinion.

Off the bat, I think that it would be great to have the UI defined in a separated module, where it can be called by `qltool` when an appropriate command line argument is specified (e.g. `--tui` or `--wizard`). The UI module would just construct the necessary arguments list and return them to `qltool`. In other words, the user could either specify a list of arguments (like they always have) or fire-up the TUI to walk them through the process - and then provide the result arguments list back to `qltool`. Once the UI module would be design to construct and return the args list, plugging it into `qltool` woud be easy.

What do you think?

Rohan-cod:
That sounds good to me. This way the user will have both the options and code will be manageable too. I'll give it some more thought to see if there's anything else that can be added/improved.

xwings:
Nice work and welcome to Qiling.

I agree we need to have a "swtich" and for now this will break the CI test due to the current for for qltool

So, i guess we fix the switch and test before we move on ?

elicn:
I thinks it would be easier to close this PR and open a new one when it is ready.

Rohan-cod:
> I thinks it would be easier to close this PR and open a new one when it is ready.

👍

elicn:
Just in case we need a cool banner ;)
```
    ██████     ███  ████   ███                     
  ███░░░░███  ░░░  ░░███  ░░░                      
 ███    ░░███ ████  ░███  ████  ████████    ███████
░███     ░███░░███  ░███ ░░███ ░░███░░███  ███░░███
░███   ██░███ ░███  ░███  ░███  ░███ ░███ ░███ ░███
░░███ ░░████  ░███  ░███  ░███  ░███ ░███ ░███ ░███
 ░░░██████░██ █████ █████ █████ ████ █████░░███████
   ░░░░░░ ░░ ░░░░░ ░░░░░ ░░░░░ ░░░░ ░░░░░  ░░░░░███
                                           ███ ░███
                                          ░░██████ 
                                           ░░░░░░  
```

Rohan-cod:
> Just in case we need a cool banner ;)
> 
> ```
>     ██████     ███  ████   ███                     
>   ███░░░░███  ░░░  ░░███  ░░░                      
>  ███    ░░███ ████  ░███  ████  ████████    ███████
> ░███     ░███░░███  ░███ ░░███ ░░███░░███  ███░░███
> ░███   ██░███ ░███  ░███  ░███  ░███ ░███ ░███ ░███
> ░░███ ░░████  ░███  ░███  ░███  ░███ ░███ ░███ ░███
>  ░░░██████░██ █████ █████ █████ ████ █████░░███████
>    ░░░░░░ ░░ ░░░░░ ░░░░░ ░░░░░ ░░░░ ░░░░░  ░░░░░███
>                                            ███ ░███
>                                           ░░██████ 
>                                            ░░░░░░  
> ```

Nicee 👌

PR #1210, opened 2022-08-17
Windows Maintenance PR

elicn:
This PR aggregates additional commits on top of older ones that were not merged, so it became a bit messy..

Highlights:
- Better consistency of kernel data structures
- Code quality improvements
- Introducing generic structure base class for easier data handling (resulted in many changes)
  - adopted only on Windows for now

### New struct module
A generic `ctypes` structure base class that provides intuitive accessors for easier data handling, simplifying data coherency (consistency with memory). The new `BaseStruct` class inherits all `ctypes.Structure` methods and properties, and provides additional ones on top of them.

Following are a few examples to show off the new capabilities:
```python
from qiling.os import struct

# define a new structure
class Point(struct.BaseStruct):
     _fields_ = [
         ('x', ctypes.c_uint32),
         ('y', ctypes.c_uint32)
     ]


# define a new structure whose alignment matches the emulated system native size
Struct = struct.get_aligned_struct(ql.arch.bits)

class AlignedPoint(Struct):
     _fields_ = [
         ('x', ctypes.c_uint32),
         ('y', ctypes.c_uint32)
     ]


# use this as target memory location for structure
ptr = 0x12340000

# create a structure, populate it and save to memory
obj = Point(x=10, y=20)
obj.save_to(ql.mem, ptr)


# load structure content from memory
obj = Point.load_from(ql.mem, ptr)
print(f'x = {obj.x}, y = {obj.y}')


# refer to a memory address as structure (similar to a C pointer).
# if content is modified, structure data is written back to memory when context exits.
with Point.ref(ql.mem, ptr) as obj:
    obj.x = 15
    obj.y = 25
# -> data changed and will be written back to memory


# same as above, but no data is modified this time.
# if content is unmodified, structure data is discarded when context exits.
with Point.ref(ql.mem, ptr) as obj:
    z = obj.y * 80 + obj.x
# -> data hasn't changed, so nothing is written back to memory


# refer to a memory address as volatile structure (similar to a volatile C pointer).
# fields are read straight from memory to reflect changes immediately: fields will be read and
# written to memory as they are accessed
obj = Point.volatile_ref(ql.mem, ptr)
obj.x = 17      # x value is flushed to memory as it is modified
# -> y value hasn't changed and will not be written-back.
# -> if needed, obj may be held for future direct access to memory
```


xwings (approved):

kabeor (approved):

qiling/os/struct.py
```diff
@@ -0,0 +1,244 @@
+from __future__ import annotations
```
wtdcode, line 1: Is it possible to rename `struct` to some other names? It shadows the standard library `struct` and might cause some confusion for our users. (`ql.os` name clash had already caused a few confusion before)

elicn: Couldn't come up with any reasonable alternative really.. any suggestions?

elicn:
Do not merge yet; a few commits to follow.

xwings:
Good to go ?

PR #1213, opened 2022-08-24
feat(r2): backtrace and control flow

chinggg:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

Now I have implemented callstack and backtrace, though the features have not been fully tested yet. I am currently working on deflat as a show case of control flow analysis ability


examples/extensions/r2/hello_r2.py
```diff
@@ -35,6 +35,7 @@ def my_sandbox(path, rootfs):
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
     # r2.enable_trace()
+    r2.bt(0x401906)
```
wtdcode, line 38: Why do we need an extra argument?

chinggg: The extra argument is the target address for setting the bt_hook, so we can print backtrace without `emu_error()`. I will test `_backtrace_fuzzy` without argument in emu_error()

qiling/extensions/r2/r2.py
```diff
@@ -268,6 +269,42 @@ def dis_nbytes(self, addr: int, size: int) -> List[Instruction]:
         insts = [Instruction(**dic) for dic in self._cmdj(f"pDj {size} @ {addr}")]
         return insts
 
+    def dis_ninsts(self, addr: int, n: int=1) -> List[Instruction]:
+        insts = [Instruction(**dic) for dic in self._cmdj(f"pdj {n} @ {addr}")]
+        return insts
+
+    def _backtrace_fuzzy(self, at: int = None, depth: int = 128) -> Optional[CallStack]:
```
wtdcode, line 276: More arch support

elicn:
I am failing to understand why you had to change the memory module..
Was there something missing or not behaving as expected?

wtdcode:
> I am failing to understand why you had to change the memory module.. Was there something missing or not behaving as expected?

By changing the memory module, Qiling and r2 will share the same memory and thus solve the previously mentioned memory sync problem.

elicn:
> By changing the memory module, Qiling and r2 will share the same memory and thus solve the previously mentioned memory sync problem.

Pfff.. thigs are getting more and more complicated. Having the memory content being managed in Python code is not a great idea.. I guess that is going to come with a great performance hit and tons of bugs. As a first exmaple, the memory range splitting is handling it wrong (ranges are splitted, but their data isn't).

wtdcode:
> > By changing the memory module, Qiling and r2 will share the same memory and thus solve the previously mentioned memory sync problem.
> 
> Pfff.. thigs are getting more and more complicated. Having the memory content being managed in Python code is not a great idea.. I guess that is going to come with a great performance hit and tons of bugs. As a first exmaple, the memory range splitting is handling it wrong (ranges are splitted, but their data isn't).

For performance, I don't think it will have any performance overhead as we are passing the raw pointer to Unicorn. The only difference is that now the memory is allocated by Python. @chinggg maybe you could do a small benchmark on `uc_mem_map` and `uc_mem_map_ptr` with `uc_mem_write/read`

For bugs, the memory splitting handling is buggy indeed (and probably results in the CI crash @chinggg) but can be fixed easily. I don't think it will introduce too many bugs as we have enough tests. Note even unicorn wrote the memory split on its own instead of using any qemu code.

chinggg:
I know little about Unicorn internal and cannot find methods to pass uc mem to r2, so we can either allocate memory in r2  or Python code. I admit it is weird to deal with memory management stuff in Python since that's the strength of Unicorn. But it seems to be more acceptable than to malloc in r2 after making PoC of both approaches. 
Now the memory is not handled well so all the CIs failed, but what is promising is that most simulation can still work without too many changes of code.
I have another concern for r2's functionality, since we just pass mapped memory instead of whole file, it may miss much information about symbols and file structures. But the static information is really what we want at first.

qiling/os/memory.py
```diff
@@ -142,12 +148,15 @@ def change_mapinfo(self, mem_s: int, mem_e: int, mem_p: Optional[int] = None, me
...
+            self.unmap(mem_s, mem_e - mem_s)
+            self.map_ptr(mem_s, mem_e - mem_s, mem_p, data)
+            self.add_mapinfo(mem_s, mem_e, mem_p, mem_info or tmp_map_info[3], tmp_map_info[4], data)
             return
 
         if mem_info is not None:
-            self.map_info[info_idx] = (tmp_map_info[0], tmp_map_info[1], tmp_map_info[2], mem_info, tmp_map_info[4])
+            self.map_info[info_idx] = (tmp_map_info[0], tmp_map_info[1], tmp_map_info[2], mem_info, tmp_map_info[4], tmp_map_info[5])
 
     def get_mapinfo(self) -> Sequence[Tuple[int, int, str, str, str]]:
```
wtdcode, line 161: type hint here

wtdcode, line 161: Data is missing.

chinggg: Thanks for pointing out this, though now there is no existing usage of `change_mapinfo` that passes data as argument

chinggg:
It seems now there leaves only one error in `test_elf_partial_linux_x8664`, but I cannot figure out why... Hoping someone can have a quick review on it so I can solve it easily.

qiling/os/memory.py
```diff
@@ -268,12 +280,12 @@ def save(self):
...
 
-        for lbound, ubound, perm, label, is_mmio in self.map_info:
+        for lbound, ubound, perm, label, is_mmio, data in self.map_info:
             if is_mmio:
                 mem_dict['mmio'].append((lbound, ubound, perm, label, *self.mmio_cbs[(lbound, ubound)]))
             else:
-                data = self.read(lbound, ubound - lbound)
-                mem_dict['ram'].append((lbound, ubound, perm, label, bytes(data)))
+                data = self.read(lbound, ubound - lbound)  # read instead of using data from map_info to avoid error
+                mem_dict['ram'].append((lbound, ubound, perm, label, data))
```
chinggg, line 288: Actually data in `map_info` will be different from the return of `uc.read`, which causes error in ql.mem.save/restore, so I made a hack here. But it seems to cover some unseen inconsistency between `map_info` and `uc.mem`, which could be source of other errorrs.

chinggg:
According to the CI testing result https://github.com/qilingframework/qiling/actions/runs/3065273326/jobs/4949204918, there seems to be 3 kind of error in `test_elf`:
- For binary of MIPS architecture, the map address is wrong when exceeding `0x80000000` https://github.com/unicorn-engine/unicorn/issues/1708
- `Memory region 3 0x47e1000 - 0x47e2000 != map_info[3] ld-linux.so.2`
- `Memory region 10 0x7fffb81c1000 - 0x7fffb81c3000 != map_info[10] [mmap] libc.so.6`

PR #1219, opened 2022-08-31
Add Windows Kernel structs

HackingFrogWithSunglasses:
As per https://github.com/qilingframework/qiling/pull/1217 this is the WIndows kernel structures PR

- [ * ] This PR introduces a new function/api for Qiling Framework.

### Coding convention?

- [ * ] The new code conforms to Qiling Framework naming convention.
- [ * ] The imports are arranged properly.
- [ * ] Essential comments are added.

### Extra tests?

- [ * ] Tests will be added after some discussion and review.

### Changelog?

- [ * ] Changelog has been updated in my PR.

### Target branch?

- [ * ] The target branch is dev branch.

### One last thing

- [ * ] I have read the [contribution guide](https://docs.qiling.io/en/latest/contribution/)

-----

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -7,7 +7,7 @@ import os, pefile, pickle, secrets, ntpath
 from typing import Any, Dict, MutableMapping, NamedTuple, Optional, Mapping, Sequence, Tuple, Union
 
 from unicorn import UcError
-from unicorn.x86_const import UC_X86_REG_CR4, UC_X86_REG_CR8
+from unicorn.x86_const import UC_X86_REG_CR4, UC_X86_REG_CR8, UC_X86_REG_GS
 
 from qiling import Qiling
 from qiling.arch.x86_const import FS_SEGMENT_ADDR, GS_SEGMENT_ADDR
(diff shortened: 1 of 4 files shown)


qiling/loader/pe.py
```diff
@@ -629,16 +629,152 @@ def init_ki_user_shared_data(self):
...
+
+        sysconf = self.ql.os.profile['SYSTEM']
+        osconf  = self.ql.os.profile[f'OS{self.ql.arch.bits}']
+
+        kthread_addr = osconf.getint('KTHREAD')
+        kthread_struct = KTHREAD
+        self.ql.mem.map(kthread_addr, self.ql.mem.align_up(kthread_struct.sizeof()), info='[kthread]')
+
+        # Initialize an instance with key fields
+        kthread_obj = kthread_struct.volatile_ref(self.ql.mem, kthread_addr)
```
elicn, line 695: This structure is initialized with magic numbers whose origin is unclear.
Is there a place [a Qilign object] we can take the values from? Consistency is key.

qiling/loader/pe.py
```diff
@@ -629,16 +629,152 @@ def init_ki_user_shared_data(self):
...
+        kthread_obj.ThreadLock      = 11
+        kthread_obj.PreviousMode    = 0
+        kthread_obj.InitialStack    = 0x1000
+        kthread_obj.StackBase       = 0x1500
+        kthread_obj.StackLimit      = 0x2000
+        kthread_obj.MiscFlags       |= 0x400
```
elicn, line 701: The value of `MiscFlags` is OR-ed with the existing one, but what is it...? That field was never initialized.

qiling/os/windows/api.py
```diff
@@ -8,7 +8,7 @@
 # See: https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types
 
 LONG   = PARAM_INTN
-ULONG  = PARAM_INTN
+ULONG  = PARAM_INT32
```
elicn, line 11: How come `ULONG` size is different from `LONG`'s ?
If Windows treats `LONG` as a 32-bit integer regardless of the arch native size, so both should be changed.

qiling/os/windows/structs.py
```diff
@@ -3,10 +3,12 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
+from calendar import c
```
elicn, line 6: Seems to be imported for no reason.

qiling/os/windows/structs.py
```diff
@@ -115,6 +117,35 @@ class PEB(Struct):
 
     return PEB
 
+class NT_TIB(struct.BaseStruct):
```
elicn, line 120: Here and on other new structures:
Are you sure the kd output is needed here? If possible, please refer to a URL where the sturcutre layout can be found rather than pasting it here.

qiling/os/windows/structs.py
```diff
@@ -115,6 +117,35 @@ class PEB(Struct):
...
+        +0x010 StackLimit       : Ptr64 Void
+        +0x018 SubSystemTib     : Ptr64 Void
+        +0x020 FiberData        : Ptr64 Void
+        +0x020 Version          : Uint4B
+        +0x028 ArbitraryUserPointer : Ptr64 Void
+        +0x030 Self             : Ptr64 _NT_TIB
+    '''
+
+    _fields_ = (
+        ('ExceptionList',           ctypes.c_void_p),
```
elicn, line 138: Here and on other new structures:
Using types like `c_void_p` causing the structure definition to be affected by the hosting system architecture, and not by the emulated system - as it should be. For example, if a 64-bit Windows binary is emulated on a 32-bit host, `c_void_p` will be 4 bytes long instead of the expected 8. For that reason we use `native_type` and `pointer_type` that are chosen according to the emulated arch.

Also, for simplicity we use simple types instead of pointers. ctypes pointers are assigned in a slightly different way, which makes it a bit more complicated, and since we do not reference 'real' memory here, we keep it simple. This is why we use `pointer_type` that is actually `native_type`.

qiling/os/windows/structs.py
```diff
@@ -238,6 +269,660 @@ class KUSER_SHARED_DATA(struct.BaseStruct):
...
+   +0x6ea0 PrcbPad140       : [508] Uint8B
+   +0x7e80 RequestMailbox   : [1] _REQUEST_MAILBOX
+    '''
+
+    # We need a native pointer so that we can write the address of _KTREAD structure
+    native_type = struct.get_native_type(64)
+    pointer_type = native_type
+
+    _fields_ = (
+        ('MxCsr',                       ctypes.c_ulong),
```
elicn, line 720: From the same reason mentioned above (avoiding structures definitions getting affected by the hosting arch), we should use only known size types, like `c_uint32` and `c_uint16` instead of `long` and `short`.

qiling/profiles/windows.ql
```diff
@@ -1,13 +1,17 @@
 [OS64]
-heap_address = 0x500000000
+heap_address = 0xfffff76000000000
```
elicn, line 2: Is there any specific reason for this change?

elicn (changes requested):
Thanks for the contribution.
As mentioned on the previous PR you opened, the additional structures are not defined in a correct way. Please see `BaseStruct` documentation and how other Windows structures are defined to get the idea.

qiling/loader/pe.py
HackingFrogWithSunglasses: `MiscFlags` can be removed in this instance, I was using this for my own purpose and it won't be useful for Qiling generically.

qiling/loader/pe.py
```diff
@@ -629,16 +629,152 @@ def init_ki_user_shared_data(self):
...
+        # Initialize and instance with a few key fields
+        kprcb_obj = kprcb_struct.volatile_ref(self.ql.mem, kprcb_addr)
+        kprcb_obj.CurrentThread = kthread_addr      # Writes _KTHREAD pointer to CurrentThread field
+        kprcb_obj.IdleThread = kthread_addr         # Writes _KTHREAD pointer to IdleThread field
+        kprcb_obj.ParentNode = knode_addr           # Writes _KNODE pointer to ParentNode field
+
+        # Writes KPRCB pointer into GS:[0x20]
+        self.ql.mem.write_ptr(0x6000020, kprcb_addr)
+
+        # @NOTE: Tests for writing structure pointers.
```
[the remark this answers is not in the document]
HackingFrogWithSunglasses: You can remove these debug statements, I just had those there for my local branch and forgot to remove them when I submitted PR.

qiling/profiles/windows.ql
HackingFrogWithSunglasses: No specific reason relevant to Qiling. Was for my own testing purposes.

qiling/profiles/windows.ql
```diff
@@ -1,13 +1,17 @@
...
 heap_size = 0x5000000
-stack_address = 0x7ffffffde000
+stack_address = 0xfffff77000000000
 stack_size = 0x40000
 image_address = 0x400000
 dll_address = 0x7ffff0000000
 entry_point = 0x140000000
-# KI_USER_SHARED_DATA = 0xfffff78000000000
-KI_USER_SHARED_DATA = 0x7ffe0000
+KI_USER_SHARED_DATA = 0xfffff78000000000
```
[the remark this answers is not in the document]
HackingFrogWithSunglasses: No specific reason, again for my own testing.

qiling/profiles/windows.ql
```diff
@@ -17,8 +21,12 @@ stack_size = 0x21000
 image_address = 0x400000
 dll_address = 0x10000000
 entry_point = 0x40000
-# KI_USER_SHARED_DATA = 0xffdf0000
-KI_USER_SHARED_DATA = 0x7ffe0000
+KI_USER_SHARED_DATA = 0xffdf0000
```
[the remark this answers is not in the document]
HackingFrogWithSunglasses: Same as above.

qiling/loader/pe.py
```diff
@@ -629,16 +629,152 @@ def init_ki_user_shared_data(self):
...
+        # @TODO: initialize StackBase & StackLimit
+        kpcr_obj = kpcr_struct.volatile_ref(self.ql.mem, kpcr_addr)
+        kpcr_obj.MajorVersion = sysconf.getint('majorVersion')
+        kpcr_obj.MinorVersion = sysconf.getint('minorVersion')
+        kpcr_obj.CurrentPrcb = kprcb_addr   # Writes _KPRCB pointer to CurrentPrcb field
+        kpcr_obj.Self = kpcr_addr           # Writes _KPCR pointer to Self field
+        kpcr_obj.Prcb = kprcb_addr          # Writes _KPRCB pointer to Prcb field
+
+        # Writes KPCR pointer into GS:[0x18]
+        self.ql.mem.write_ptr(0x6000018, kpcr_addr)
```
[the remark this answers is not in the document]
HackingFrogWithSunglasses: This is not a magic number. That is the exact offset for the KPRCB pointer in Windows KM. I agree that the constant + offset should be used, though.

qiling/os/windows/api.py
HackingFrogWithSunglasses: I did not mean to include this change in the PR, this can be removed.

qiling/os/windows/structs.py
HackingFrogWithSunglasses: Agree, can be removed.

HackingFrogWithSunglasses (commented):
I agree with the requested changes.

qiling/loader/pe.py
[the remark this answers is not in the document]
elicn: Great, that's what I meant.

HackingFrogWithSunglasses (commented):
:+1:

qiling/loader/pe.py
elicn: I actually referred to the `kthread_obj` fields values: 11, 0x1000, 0x1500, 0x2000?

qiling/os/windows/structs.py
elicn: Yes, I am aware.. if there is an unformal documentation you can link to, that's perfectly fine.

elicn (commented):
I took the liberty to edit some of the files and apply the fixes myself.
Also added a few responses.

HackingFrogWithSunglasses:
> 

That's absolutely fine as I don't have time myself to maintain this. Thanks!

PR #1227, opened 2022-09-10
allow user to build config from dictionary other than disk file

ucgJhe:
I find change `profile.ql` constantly sometimes can be a bit tedious. So, i think allower user build their config from dictionary other than reading config from disk file could be a good idea.


kabeor (approved):

qiling/utils.py
```diff
@@ -404,32 +404,37 @@ def select_os(ostype: QL_OS) -> QlClassInit['QlOs']:
...
 
+        # build config from dictionary, but not read from disk file
+        if isinstance(config_src, dict):
+            config.read_dict(config_src)
+
+        else:
+            qiling_home = Path(inspect.getfile(inspect.currentframe())).parent
+            os_profile = qiling_home / 'profiles' / f'{ostype.name.lower()}.ql'
+
+            profiles = [os_profile]
```
elicn, line 431: I don't know whether this is intentional or not, but when a filename is specified, the user-specified profile file will be appeneded to the default profile (that is, there are defaults and user may provide specific overrides). The new addition doesn't support that, so if a user provides a dictionary rather than a filename, then there are no defaults - just the mapping on the specified dictionary.

If we want to retain the same behavior also for diciotnary-based profile, then we need to extract this line (i.e. `profiles = [os_profile]`) out.

elicn (approved):

ucgJhe:
thx @elicn for helping me to figure out.

elicn (approved):

PR #1230, opened 2022-09-17
Adds a terminal user interface`qltui` for qltool

Rohan-cod:
# Description

Updated `qltool` to add an option to show a terminal user interface (`qltui`). 

## Qltui Functionality

- Runs on top of `qltool`.
- Accept data using a terminal interface for `run` and `code` commands.
- Returns an `Argparse Namespace` object back to `qltool` for execution.
- Interactive report viewer plus an option to save it as a pdf.

## How to Run?

```
./qltool qltui
```
## Images
<p>
Screenshot 2022-09-11 at 9 38 02 PM
Screenshot 2022-09-11 at 9 38 34 PM
Screenshot 2022-09-11 at 9 39 31 PM
Screenshot 2022-09-11 at 9 39 27 PM
Screenshot 2022-09-11 at 9 39 39 PM
Screenshot 2022-09-11 at 9 40 40 PM
Screenshot 2022-09-11 at 9 42 37 PM
</p>


qltui.py
```diff
@@ -0,0 +1,743 @@
...
+HEX_REGEX = r'^(0[xX])[a-fA-F0-9]+$'
+
+
+class Callback_Functions():
+    """
+    Callback Functions for Hook Operation
+    """
+
+    @staticmethod
+    def read_mem_2_args(ql, user_data):
```
elicn, line 59: Code duplication for `read_mem_*_args` seems to be avoidable by referring all methods to a single implementation. Anyway all args other than `user_data` are discarded.

A cleaner way to do that would be defining just one method that receives an arbitrary number of arguments, but expects a named `user_data` arg at the end. This way all methods would be consolidated into just one method, and calling them would be done by expending a `**kwargs` rather than just `*args` (at the bottom line of this file)

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+        ql.log.info(buff)
+
+    @staticmethod
+    def read_mem_6_args(ql, address=0, size=0, access=1, value=0, user_data=None):
+        buff = ql.mem.read(user_data["address"], user_data["bytes_size"])
+        ql.log.info(f"Hook was triggered at -> {user_data['address']}")
+        ql.log.info(buff)
+
+    @staticmethod
+    def read_reg_2_args(ql,user_data):
```
elicn, line 77: See comment about `read_mem_*_args` code duplication.

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+        ql.log.info(buff)
+
+    @staticmethod
+    def read_reg_6_args(ql, address=0, size=0, access=1, value=0, user_data=None):
+        buff = ql.reg.read(user_data["register_name"])
+        ql.log.info(f"Hook was triggered at -> {user_data['register_name']}")
+        ql.log.info(buff)
+
+    @staticmethod
+    def write_mem_2_args(ql, user_data):
```
elicn, line 95: See comment about `read_mem_*_args` code duplication.

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+        ql.log.info(buff)
+
+    @staticmethod
+    def write_mem_6_args(ql, address=0, size=0, access=1, value=0, user_data=None):
+        buff = ql.mem.write(user_data["address"], user_data["value"])
+        ql.log.info(f"Hook was triggered at -> {user_data['address']}")
+        ql.log.info(buff)
+
+    @staticmethod
+    def write_reg_2_args(ql,user_data):
```
elicn, line 113: See comment about `read_mem_*_args` code duplication.

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+        ql.log.info(buff)
+
+    @staticmethod
+    def write_reg_6_args(ql, address=0, size=0, access=1, value=0, user_data=None):
+        buff = ql.reg.write(user_data["register_name"], user_data["value"])
+        ql.log.info(f"Hook was triggered at -> {user_data['register_name']}")
+        ql.log.info(buff)
+
+    @staticmethod
+    def emu_start_2_args(ql, user_data):
```
elicn, line 131: See comment about `read_mem_*_args` code duplication.

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+    @staticmethod
+    def emu_start_4_args(ql, address=0, size=0, user_data=None):
+        ql.emu_start(begin=user_data["start"], end=user_data["end"])
+
+    @staticmethod
+    def emu_start_6_args(ql, address=0, size=0, access=1, value=0, user_data=None):
+        ql.emu_start(begin=user_data["start"], end=user_data["end"])
+
+    @staticmethod
+    def emu_stop_2_args(ql, user_data=None):
```
elicn, line 143: See comment about `read_mem_*_args` code duplication.

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+        ql.log.info('killer switch found, stopping')
+        ql.emu_stop()
+
+    @staticmethod
+    def emu_stop_6_args(ql, address=0, size=0, access=1, value=0, user_data=None):
+        ql.log.info('killer switch found, stopping')
+        ql.emu_stop()
+
+    @staticmethod
+    def save_2_args(ql, user_data=None):
```
elicn, line 158: See comment about `read_mem_*_args` code duplication.

qltui.py
```diff
@@ -0,0 +1,743 @@
...
+        content = f.read()
+
+    return content
+
+
+def log(string, color):
+    """
+    Function to beautify terminal output
+    """
+    if colored:
```
elicn, line 205: `colored` is constant throughout the emulation. Instead of testing for it every time a log line is about to be emitted, it could be determined once and choose between two methods to implement it (i.e. the `if` clause or the `else` one).

elicn (commented):
That looks aswesome..!
I like the banner and the overall design ;)

qltui.py
```diff
@@ -726,18 +660,27 @@ def hook(ql):
...
+    return hook_dictionary
 
-        else:
-            operation += '_2_args'
 
-        getattr(ql, hook_type)(getattr(Callback_Functions, operation), *args)
+def transform_syscalls(syscalls, keys=["address"], func=lambda x: hex(x)):
+    for i in syscalls:
+        try:
+            if type(i) == type({}):
```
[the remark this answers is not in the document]
Rohan-cod: Hi @elicn, `Iterable` is also picking up string that is not giving the correct result. I clubbed it using or to have just two conditionals.

qltui.py
```diff
@@ -56,114 +56,45 @@ class Callback_Functions():
     """
 
     @staticmethod
-    def read_mem_2_args(ql, user_data):
+    def read_mem(ql, *args):
```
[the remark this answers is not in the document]
Rohan-cod: If I add `user_data` after `*args`, it is required that it is passed as a `keyword-only` argument. But [here](https://github.com/qilingframework/qiling/blob/dev/qiling/core_hooks_types.py#L27) where the callback function is called, it is passed directly and not as `user_data = user_data`, which leads to this error

Screenshot 2022-10-01 at 4 35 04 PM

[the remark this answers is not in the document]
elicn: I didn't notice that these functions are called directly by Qiling hooks.. that makes sense.

qltui.py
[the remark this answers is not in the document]
elicn: Well, that wasn't for filtering bad arguments but for generalizing the condition.
If you go with the `isinstance` and explicit types, then you could write `isinstance(x, (type1, type2, type3)` instead of `or`-ing them. Anyway, I am fine with this either.

kabeor:
plz fix ci crash which cause by missing `json2html` module

elicn (approved):

kabeor (approved):

kabeor:
Thanks for your contribution.

PR #1244, opened 2022-10-13
feat: R2Qiling with refactored memory and de-flatten plugin

chinggg:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

The draft PR #1213 refactors `ql.mem` by changing its internal implementation, which cannot get agreement from all developers. So I keep the existing code almost intact and move the promising but unstable feature into `R2Qiling` and `R2Mem`. Users can just replace `Qiling` with `R2Qiling` to enjoy and  test the r2-powered Qiling.


qiling/os/memory.py
```diff
@@ -166,7 +166,7 @@ def __perms_mapping(ps: int) -> str:
 
             return ''.join(val if idx & ps else '-' for idx, val in perms_d.items())
 
-        def __process(lbound: int, ubound: int, perms: int, label: str, is_mmio: bool) -> Tuple[int, int, str, str, str]:
+        def __process(lbound: int, ubound: int, perms: int, label: str, is_mmio: bool, _data: bytearray = None) -> Tuple[int, int, str, str, str]:
```
wtdcode, line 169: Why do we need this?

chinggg: The wrapper `R2Mem` may call `get_mapinfo` then `__process` each entry in `map_info`, I add this optional argument since `map_info` in `R2Mem` has 6 elements instead of 5. I don't want to copy much code which is the same except for  # of parameters. Most importantly, **It will not affect the existing code**.

chinggg: @elicn Do you approve these changes on the existing code?

examples/extensions/r2/deflat_r2.py
```diff
@@ -0,0 +1,25 @@
...
+
+from qiling import Qiling
+from qiling.const import QL_VERBOSE
+from qiling.extensions.r2 import R2Qiling
+
+
+
+if __name__ == "__main__":
+    # a program obfuscated by OLLVM CFF flatten, which should print 4 when argv[1] is 1
+    ql = R2Qiling(['rootfs/x86_linux/bin/test_fla_argv', '1'], 'rootfs/x86_linux', verbose=QL_VERBOSE.DEFAULT)
```
wtdcode, line 18: Source code of this file and steps to compile.

chinggg:
```c
#include <stdio.h>
#include <stdlib.h>
                 
unsigned int target_function(unsigned int n)
{                
  unsigned int mod = n % 4;
  unsigned int result = 0;                                                                                                         
                 
  if (mod == 0) result = (n | 0xBAAAD0BF) * (2 ^ n);
                 
  else if (mod == 1) result = (n & 0xBAAAD0BF) * (3 + n);
                                                                                                                                   
  else if (mod == 2) result = (n ^ 0xBAAAD0BF) * (4 | n);                                                                          
                                                                                                                                   
  else result = (n + 0xBAAAD0BF) * (5 & n);                                                                                        
                                                                                                                                   
  return result;                                                                                                                   
}                                                                                                                                  
                                                                                                                                   
int main(int argc, char **argv) {                                                                                                  
   int n;                                                                                                                          
   if (argc < 2) {                                                                                                                 
      n = 0;                                                                                                                       
   } else {                                                                                                                        
      n = atoi(argv[1]); nptr:                                                                                                     
   }                                                                                                                               
   int val = target_function(n);                                                                                                   
   printf("%d\n", val); format:                                                                                                    
   return 0;                                                                                                                       
}
```

`clang -m32 -mllvm -fla test.c -o test_fla_argv`

chinggg: @wtdcode Should I upload the binary to rootfs or put source code in the comment of this example?

wtdcode: > @wtdcode Should I upload the binary to rootfs or put source code in the comment of this example?

`examples/src/linux`

examples/extensions/r2/deflat_r2.py
```diff
@@ -0,0 +1,26 @@
...
+from qiling import Qiling
+from qiling.const import QL_VERBOSE
+from qiling.extensions.r2 import R2Qiling
+
+
+
+if __name__ == "__main__":
+    # a program obfuscated by OLLVM CFF flatten, which should print 4 when argv[1] is 1
+    # see source code at examples/src/linux/fla_test.c
+    ql = R2Qiling(['rootfs/x86_linux/bin/test_fla_argv', '1'], 'rootfs/x86_linux', verbose=QL_VERBOSE.DEFAULT)
```
elicn, line 19: I think that removing the Qiling import and importing r2q as Qiling would be a cleaner approach; e.g.
```python
from qiling.extensions.r2 import R2Qiling as Qiling
# ...
ql = Qiling(...)
```

examples/extensions/r2/deflat_r2.py
```diff
@@ -0,0 +1,26 @@
...
+    r2 = ql.r2
+    # now we can use r2 parsed symbol name instead of address
+    fcn = r2.get_fcn_at(r2.where('target_function'))
+    print(fcn)
```
elicn, line 23: I [and possibly everyone who looks at this code the first time] have no idea what is going on and why you are doing what you are doing. Please *heavily comment* all steps, and what each of them is suppose to do.

printing a function?
deflat?
shell?
no idea.

examples/extensions/r2/hello_r2.py
```diff
@@ -35,6 +35,7 @@ def my_sandbox(path, rootfs):
     ql.hook_address(func, r2.functions['main'].offset)
     # enable trace powered by r2 symsmap
     # r2.enable_trace()
+    r2.set_backtrace(0x401906)
     ql.run()
 
 if __name__ == "__main__":
```
elicn, line 41: It looks like this file contains two differnt examples.
Please break them down to two files, or clarify how they are related to each other (in case trhey are).

examples/extensions/r2/mem_r2.py
```diff
@@ -0,0 +1,88 @@
+import sys
+from types import FrameType
+sys.path.append('..')
+
+from tests.test_elf import ELFTest
+from qiling import Qiling
+from qiling.const import QL_VERBOSE
+from qiling.extensions.r2 import R2
+
+def test_elf_linux_arm():
```
elicn, line 10: I am confused.. is this a test or an example?

examples/extensions/r2/mem_r2.py
```diff
@@ -0,0 +1,88 @@
...
+
+def addr_hook(ql: "Qiling"):
+  mapinfo = ql.mem.map_info
+  for i, mem_region in enumerate(ql.uc.mem_regions()):
+    if i != 8: continue
+    uc_mem = ql.mem.read(mem_region[0], mem_region[1] - mem_region[0] + 1)
+    with open('right.bin', 'wb') as f:
+      f.write(uc_mem)
+  
+if __name__ == '__main__':
```
elicn, line 71: Examples should be crystal clear, otherwise they are just confusing.
Please cleanup this file and leave it only with the relevant example code.

qiling/extensions/r2/__init__.py
```diff
@@ -1 +1,39 @@
+from qiling import Qiling
 from .r2 import R2
+from .mem import R2Mem
+from .deflat import R2Deflator
+
+from unicorn.unicorn_const import UC_PROT_READ, UC_PROT_WRITE, UC_PROT_EXEC
+
+
```
elicn, line 8: `__init__` should not contain funcionality logic directly, but only imports.
Better split the classes to different files, even if they are short.

elicn (changes requested):
Thanks for the effort you put in to this!
Please go over the code, clean and tidy it up, so it would be more intuitive to use out of the box.

xwings:
@wtdcode  are we ready to merge this ?

chinggg:
Currently example code like `r2.strings` is broken since r2 seems to be unaware of the whole binary.

xwings:
once this is merged and some minor bug fix. we can tag a release

wtdcode:
No, unfortunately it's not ready to merge yet and probably needs a few more revise.

chinggg:
It seems https://github.com/qilingframework/qiling/pull/1293 has made some improvements to memory management functions like `del_mapinfo`. I will take a look and update this branch.

chinggg:
I have tried to keep the r2 extension compatible with existing code, now there is no extra record of bytearray in `map_info`, not sure if there is other problems.

PR #1252, opened 2022-10-16
fix MIPS relocs

ucgJhe:
fix wrong MIPS reloc type.


xwings:
Interesting, just wondering does it break any binary ?

xwings (approved):

ucgJhe:
> 

it's not breaking binaries, but qiling couldn't intercept target function it should be able to.

before the fix:
image

after the fix:
image

kabeor (approved):

PR #1272, opened 2022-11-17
Feature/Fix some error in syscall fcntl and getsockopt 

owl129:
## Fix syscall fcntl
```
int fcntl(int fd, int cmd, ... /* arg */ );
```
when call syscall fcntl with cmd F_SETFL, file descriptor in args was set,  Which need mapping FIRST to support different OS/Arch.

## Feature add to getsockopt

Add level IPPROTO_TCP support for syscall getsockopt

Also fix exception handle in getsockopt

## Implements poll syscall

https://man7.org/linux/man-pages/man2/poll.2.html 
```
#include <poll.h>

int poll(struct pollfd *fds, nfds_t nfds, int timeout);
```
pollfd params struct
```
struct pollfd {
   int   fd;         /* file descriptor */
   short events;     /* requested events */
   short revents;    /* returned events */
};
```


xwings:
Hi @owl129 there is a conflict based on the recent merge. can you please fix it ?

qiling/os/posix/const_mapping.py
```diff
@@ -154,6 +154,17 @@ def socket_domain_mapping(p: int, archtype: QL_ARCH, ostype: QL_OS) -> str:
     return _constant_mapping(p, socket_domain_map)
 
 
+def socket_tcp_option_mapping(t: int, archtype: QL_ARCH, ostype: QL_OS) -> str:
```
elicn, line 157: Even if we want to keep it consistent with other similar methods, `ostype` is not used here any may be removed.

qiling/os/posix/syscall/poll.py
```diff
@@ -4,6 +4,34 @@
...
 
 from qiling import Qiling
+import select
 
 def ql_syscall_poll(ql: Qiling, fds: int, nfds: int, timeout: int):
-    return 0
\ No newline at end of file
+    fn_map = {}
+    try:
+        p = select.poll()
```
elicn, line 12: This is going to break on a Windows host, as it only supports `select`.
Perhaps we need to have two implementations, based on the hosting system.

qiling/os/posix/syscall/poll.py
```diff
@@ -4,6 +4,34 @@
...
+        for i in range(nfds):
+            fd = ql.mem.read_ptr(fds + i*8, 4)
```
elicn, line 14: Consider using a `BaseStruct` for that, instead of doing pointer arithmetics and reading memory.
Specifically, look at the `ref` method which woud be a good fir here.

elicn (changes requested):

qiling/os/posix/const_mapping.py
owl129: I not sure whether the option has diff on macos, like other methods. I'll just remove it first.

qiling/os/posix/syscall/poll.py
owl129: As i know, `select()` canot handle events like `poll()`, so i just limit the syscall implemention in Linux host. Do you have a better solution?

qiling/os/posix/syscall/poll.py
```diff
@@ -4,34 +4,48 @@
...
-            ql.mem.write_ptr(fds + idx*8 +6, revent, 2)
-        
-    except Exception as e:
-        ql.log.error(f'{e} {fds=}, {nfds=}, {timeout=}')
-        regreturn = -1
-            
-    return regreturn
+    base = struct.BaseStructEL if ql.arch.endian == QL_ENDIAN.EL else struct.BaseStructEB
+
+    class Pollfd(base):
```
elicn, line 16: Thanks for chaning that, it looks much better now.
Could you please move the structure definition to `os.posix.structs.py` where all other socket-related structures are already? There you'll also see a pattern of defining those structures.

elicn (changes requested):

qiling/os/posix/syscall/poll.py
elicn: No, that is exactly what I meant.

elicn (approved):

owl129:
@xwings can u help review this request？

xwings (approved):

xwings:
Cool and Thanks!

PR #1289, opened 2022-12-09
Qdb: add command show_args

ucgJhe:
show arguments for different architecure X86, X8664, MIPS, ARM, CORTEX_M for now

as for MIPS, it shows right value if certain register value modified by the delay slot instruction

image


xwings:
Nice!

kabeor (approved):

PR #1300, opened 2022-12-31
Clear memory before restore.

### LINKED ISSUE #1136: Restoring a state with MMIO causes an exception
**Describe the bug**
Mapping an MMIO region, then saving the qiling state and then restoring the qiling state causes an exception. See the script below.

**Sample Code**
```python
import qiling
from qiling.const import QL_ARCH

ql = qiling.Qiling(code=bytes(0x1000), archtype=QL_ARCH.ARM, ostype='linux')
ql.mem.map_mmio(addr=0, size=0x1000, read_cb=lambda *args: 0, write_cb=lambda *args: None)
qiling_state = ql.save()

# ...

ql.restore(qiling_state)
```
**Actual behavior**
```
$ python main.py 
  File "main.py", line 15, in <module>
    ql.restore(qiling_state)
  File "~/.local/lib/python3.8/site-packages/qiling/core.py", line 783, in restore
    self.mem.restore(saved_states["mem"])
  File "~/.local/lib/python3.8/site-packages/qiling/os/memory.py", line 275, in restore
    self.map_mmio(lbound, ubound - lbound, read_cb, write_cb, info=label)
  File "~/.local/lib/python3.8/site-packages/qiling/os/memory.py", line 547, in map_mmio
    self.ql.uc.mmio_map(addr, size, __mmio_read, read_cb, __mmio_write, write_cb)
  File "~/.local/lib/python3.8/site-packages/unicorn/unicorn.py", line 527, in mmio_map
    raise UcError(status)
unicorn.unicorn.UcError: Invalid memory mapping (UC_ERR_MAP)
```

**Expected behavior**
No exception.

**Additional context**
An exception occurs in unicorn because the MMIO memory region is mapped, while it is already mapped. I think the best way to handle this is to first unmap the existing memory region, and then map the MMIO region. I don't think simply ignoring the MMIO region if it overlaps with a region that is already mapped is possible, since the read/write handler might be different.

https://github.com/qilingframework/qiling/blob/573ce1fafeec66e010b818da102d9fe0bf733a79/qiling/os/memory.py#L291-L295

**Proposed patch**
```python
    for lbound, ubound, perms, label, read_cb, write_cb in mem_dict['mmio']: 
        self.ql.log.debug(f"restoring mmio range: {lbound:#08x} {ubound:#08x} {label}")

        size = lbound - ubound
        if self.is_mapped(lbound, size):
            # unmap whatever region is already there
            self.unmap(lbound, size)

        # map the MMIO region
        self.map_mmio(lbound, size, read_cb, write_cb, info=label)
```

LukeSerne:
This fixes #1136 by unmapping everything before applying the restore operation. This will make sure that after a restore, the memory state will be exactly the same as it was when the memory's `save` method was called.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/memory.py
+++ b/qiling/os/memory.py
@@ -286,6 +286,9 @@ class QlMemoryManager:
         """Restore saved memory content.
         """
 
+        # Unmap everything so we start with a clean memory state.
+        self.unmap_all()
+
         for lbound, ubound, perms, label, data in mem_dict['ram']:
             self.ql.log.debug(f'restoring memory range: {lbound:#08x} {ubound:#08x} {label}')
 


chinggg:
@wtdcode @elicn Do you think `mem.restore()` should have the semantic to clear all memory?

elicn:
That is a good question.
I tend to say "yes", because restore operations are usually used to overwrite an existing state with a previous one. Since we restore the entire memory map here (as opposed to restoring each memory region individually), I think this is expected.

LukeSerne:
What is the status of this PR? I see some tests are failing on ubuntu, but I don't really see why. Personally, I think that if the tests assume the old behaviour (where a `restore` would not clear the memory state), they should be changed.

PR #1305, opened 2023-01-16
add re_sigsupend sched_get_priority_max sched_get_priority_min

another1024:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/os/posix/syscall/sched.py
+++ b/qiling/os/posix/syscall/sched.py
@@ -136,3 +136,9 @@ def ql_syscall_sched_yield(ql: Qiling):
     ql.emu_stop()
     ql.os.thread_management.cur_thread.sched_cb = _sched_yield
     return 0
+    
+def ql_syscall_sched_get_priority_max(ql: Qiling,intpolicy: int):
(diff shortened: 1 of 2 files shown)


elicn:
Hi @another1024 and thanks for submitting the PR.
Non-functional system call stubs have little positive effect; in fact, they might cause problems because they provide a false sense of functionality where there is one. Would you consider implementing the syscall in a more meaningful / functional way?

another1024:
i can try to add function by meaningful  way, but It seems that the signal mechanism such as "sigaction_act" has not been processed, so if I modify this function, I may need to make some changes in thread.py. I am not sure whether such changes are acceptable, or I just need to save the mask somewhere.

another1024:
i can try to add function by meaningful way, but It seems that the signal mechanism such as "sigaction_act" has not been processed,and sched not work on coroutine,so if I modify this function, I may need to make some changes in thread.py and make signal send function. I am not sure whether such changes are acceptable, or I just need to save the mask somewhere.

PR #1358, opened 2023-06-01
Improved handling of export table for invalid export address and removed GandCrab workaround in GetProcAddress

clairelevin:
Binaries with an invalid export address table cause Qiling to crash when attempting to load exports. This fix ensures that pefile has loaded the export table before attempting to parse the exports.

Additionally, the hook for GetProcAddress contains a workaround for GandCrab that returns 0 when looking for RtlComputeCrc32. This causes other emulated binaries using the RtlComputeCrc32 API to fail, which makes me think this workaround was included unintentionally.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -553,6 +553,9 @@ class Process:
 
         # Do a full load if IMAGE_DIRECTORY_ENTRY_EXPORT is present so we can load the exports
         pe.full_load()
+        
+        # check whether loading of export table succeeded
+        if not hasattr(pe, 'DIRECTORY_ENTRY_EXPORT'): return
 
         iat = {}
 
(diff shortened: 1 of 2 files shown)


elicn:
Doesn't the check on line 551 help avoid this situation?

clairelevin:
The check on line 551 can fail if the export table exists, but is malformed. In that situation, the IMAGE_DIRECTORY_ENTRY_EXPORT attribute is present, but when pefile.full_load() is called, it can't parse the export table. Because of this, the DIRECTORY_ENTRY_EXPORT attribute never gets created, so Qiling crashes when it tries to access it.

elicn:
OK, then better document that just above that line (something like: "address a corner case where ...."), because it might look redundant for someone who is not familiar with that exact corner case.

Also, please break the `if` statement so the `return` appears on a different line.

clairelevin:
Ok, I've made the changes.

xwings:
@elicn are you ok with the updates

elicn (approved):

kabeor (approved):

kabeor:
Thanks, merged.

PR #1367, opened 2023-07-28
update return register accordingly for all arch

ucgJhe:
update return register accordingly, so setReturnAddress could work as expected.

### DIFF
--- a/qiling/cc/arm.py
+++ b/qiling/cc/arm.py
@@ -20,8 +20,7 @@ class QlArmBaseCC(QlCommonBaseCC):
         return 1
 
     def setReturnAddress(self, addr: int) -> None:
-        # TODO: do we need to update LR?
-        self.arch.stack_push(addr)
+        self.arch.regs.lr = addr
 
     def unwind(self, nslots: int) -> int:
         # TODO: cleanup?
--- a/qiling/cc/mips.py
+++ b/qiling/cc/mips.py
@@ -12,6 +12,9 @@ class mipso32(QlCommonBaseCC):
     _shadow = 4
     _retaddr_on_stack = False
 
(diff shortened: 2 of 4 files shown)


elicn:
I am wondering why these haven't failed so far.. @ucgJhe, did you encounter any failures that brought you to fix these?

Also, would appreciate another pair of eyes that are more familiar with those arch that I do.
@wtdcode, what do you think?

ucgJhe:
> I am wondering why these haven't failed so far.. @ucgJhe, did you encounter any failures that brought you to fix these?
> 
> Also, would appreciate another pair of eyes that are more familiar with those arch that I do. @wtdcode, what do you think?

Basically, I've been working on improving function call feature and happened to found this is not in the right place.

xwings:
> > I am wondering why these haven't failed so far.. @ucgJhe, did you encounter any failures that brought you to fix these?
> > Also, would appreciate another pair of eyes that are more familiar with those arch that I do. @wtdcode, what do you think?
> 
> Basically, I've been working on improving function call feature and happened to found this is not in the right place.

interesting. i think you should start to add in some small unit test. i remember u told me the test run too long. but again, bad test it ways better better then no test.

elicn (approved):

wtdcode:
> I am wondering why these haven't failed so far.. @ucgJhe, did you encounter any failures that brought you to fix these?
> 
> Also, would appreciate another pair of eyes that are more familiar with those arch that I do. @wtdcode, what do you think?

The comment seems left by me. IIRC, the reason behind the comment is that, depending on the context, the return address _may_ be on stack, say interrupt handlers. Though in most cases for userland code, LR is used instead.

xwings (approved):

PR #1376, opened 2023-08-24
Periodic maintenance PR

elicn:
Yet another maintenance PR including bug fixes and various improvements.

Highlights:
 - General
   - Refactored system calls ABI to be cleaner and more maintainable
   - qltool tweaks
   - Refreshed many tests
 - POSIX
   - Extracted msq and shm functionalities
   - Fixed fd args signed int handling

New Features:
 - Users may now specify a CPU model on Qiling init for Unicorn to use. Specifying a CPU model provides users with better control over the ISA supported by the underlying Unicorn instance.

```python
from qiling import Qiling
from qiling.arch.models import X86_CPU_MODEL

ql = Qiling(['my_program'], r'./examples/rootfs/x8664_linux', cputype=X86_CPU_MODEL.INTEL_SKYLAKE_CLIENT)
ql.run()
```


xwings (approved):

qiling/arch/models.py
```diff
@@ -0,0 +1,447 @@
+#!/usr/bin/env python3
+#
+# Cross Platform and Multi Architecture Advanced Binary Emulation Framework
+#
+
+from enum import Enum
+from typing import Union
+
+
+class X86_CPU_MODEL(Enum):
```
wtdcode, line 10: Is it possible to sync this automatically from Unicorn? Or is it blocked by #1629?

elicn: No, it is unrelated.
I did consider building the enumeration in runtime through reflection / introspection but that would miss the point of simplifying the usage by enabling IDE hints: When the enumeration is available in editing time the IDE can assist with autocompletion and linting. If the enumeration is built in runtime, the IDE doesn't know what to expect there.

I agree that enumerating everything in runtime is more maintainable and elegant (?), but I think usability comes first. See a past PR with a similar idea that got rejected: #581 

I think that the way it is now, backed up by tests, should cover the advantages of both methods.

wtdcode: > that would miss the point of simplifying the usage by enabling IDE hints

This can be done by adding the interfaces in unicorn. Like

```python
def cpus() -> [CPU1|CPU2|CPU3]
     # auto generated enums
     return ....
```

> I think that the way it is now, backed up by tests, should cover the advantages of both methods.

I agree. I'm also fine with this approach anyway.

elicn: That's a toally different story.
Unicorn doesn't use Enums, only integer values, and could improve a lot if it goes that path.

wtdcode: > That's a toally different story. Unicorn doesn't use Enums, only integer values, and could improve a lot if it goes that path.

Cool, I will try to do that in 2.1.0

wtdcode: I apologize for long delay as too many things happened last year. I promise that PR will go to 2.1.0 as I’m trying to work on it.

xwings (approved):

kabeor (approved):

PR #1380, opened 2023-09-05
fix bugs in handling argv/env with bytes

### LINKED ISSUE #1379: qiling framework has a potential issue when handling `env` parameters of type bytes.
Hello, the bug I found is very similar to issue #635. I guess that the developers have overlooked this issue again after refactoring the relevant code.

**Describe the bug**
In `qiling/loader/elf.py`, line 297, has following code
``` python
# write env
for k, v in env.items():
    new_stack = __push_str(new_stack, f'{k}={v}')
    elf_table.extend(self.ql.pack(new_stack))
```
while `__push_str` is in line 275
``` python
def __push_str(top: int, s: str) -> int:
    """Write a string to stack memory and adjust the top of stack accordingly.
    Top of stack remains aligned to pointer size
    """

    data = s.encode('latin') + b'\x00'
    top = self.ql.mem.align(top - len(data), self.ql.arch.pointersize)
    self.ql.mem.write(top, data)

    return top
```
If `k` and `v` in `env` is of type `bytes`, `f'{k}={v}'` will force convert the `bytes` `k` into `str`, which would cause bugs.
Additionally, there is only a `__push_str` function which can not handle case when the argv or env is of type bytes.

**Sample Code**
the `env` with `bytes` data is passed into `Qiling` like following:
```python
ql = Qiling(["x8664_freebsd/bin/x8664_hello_asm"], "x8664_freebsd", env = {b"key": b"\xff"})
ql.run()
```

**Screenshots**
image

**Additional context**
As you can see, `bytes` is converted into `str` and has new char `b`, `'` added into it, this is not we want.

It's worth noting that this bug doesn't cause the Python program to throw any exceptions, but the bug exactly exists.

I have create a PR #1380 to fix this bug (maybe a little dirty).

noobone123:
## Checklist

fix #1379 

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -282,13 +282,30 @@ class QlLoaderELF(QlLoader):
             self.ql.mem.write(top, data)
 
             return top
+        
+        def __push_bytes(top: int, s: bytes) -> int:
+            """
+            Write a bytes to stack memory and adjust the top of stack accordingly.
+            Top of stack remains aligned to pointer size
+            """
+
+            data = s + b"\x00"
+            top = self.ql.mem.align(top - len(data), self.ql.arch.pointersize)
+            self.ql.mem.write(top, data)
+
+            return top
 
         # write argc
         elf_table.extend(self.ql.pack(len(argv)))
 
         # write argv
         for s in argv:
-            new_stack = __push_str(new_stack, s)
+            if isinstance(s, str):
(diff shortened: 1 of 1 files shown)


elicn:
Implemented as part of #1376

PR #1389, opened 2023-09-23
Add EZCOV coverage support

TheZ3ro:

- [X] This PR introduces a new function/api for Qiling Framework.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR add support to the EZCOV coverage format as specified by the [Cartographer](https://github.com/nccgroup/Cartographer/blob/main/EZCOV.md) tool.
This is just a "copy" of the already existing DRCOV coverage format extension already present in Qiling.

### DIFF
--- a/qiling/extensions/coverage/formats/ezcov.py
+++ b/qiling/extensions/coverage/formats/ezcov.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+# 
+# Cross Platform and Multi Architecture Advanced Binary Emulation Framework
+#
+
+from os.path import basename
+
+from .base import QlBaseCoverage
+
+
+# Adapted from https://github.com/nccgroup/Cartographer/blob/main/EZCOV.md#coverage-data
+class bb_entry(dict):
+    def __init__(self, offset, size, mod_id):
+        self.offset = '0x{:08x}'.format(offset)
+        self.size = size
+        self.mod_id = f"[ {mod_id if mod_id is not None else ''} ]"
+
+    def csvline(self):
+        return f"{self.offset},{self.size},{self.mod_id}\n"
+
+class QlEzCoverage(QlBaseCoverage):
+    """
+    Collects emulated code coverage and formats it in accordance with the Cartographer Ghidra extension:
+    https://github.com/nccgroup/Cartographer/blob/main/EZCOV.md
+
+    The resulting output file can later be imported by coverage visualization tools such
+    as Cartographer: https://github.com/nccgroup/Cartographer/
+    """
+
+    FORMAT_NAME = "ezcov"
+
+    def __init__(self, ql):
+        super().__init__()
+        self.ql            = ql
+        self.ezcov_version = 1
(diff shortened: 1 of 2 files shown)


qiling/extensions/coverage/formats/ezcov.py
```diff
@@ -0,0 +1,59 @@
...
+    https://github.com/nccgroup/Cartographer/blob/main/EZCOV.md
+
+    The resulting output file can later be imported by coverage visualization tools such
+    as Cartographer: https://github.com/nccgroup/Cartographer/
+    """
+
+    FORMAT_NAME = "ezcov"
+
+    def __init__(self, ql):
+        super().__init__()
```
elicn, line 33: According to the latest changes, `__init__` should recieve a ql argument.
When this is done, the following line (setting ql to self) can be removed.

qiling/extensions/coverage/formats/ezcov.py
```diff
@@ -0,0 +1,59 @@
...
+        self.ql            = ql
+        self.ezcov_version = 1
+        self.ezcov_flavor  = 'ezcov'
+        self.basic_blocks  = []
+        self.bb_callback   = None
+
+    @staticmethod
+    def block_callback(ql, address, size, self):
+        for mod in ql.loader.images:
```
elicn, line 42: consider using `ql.loader.find_containing_image` instead of this loop.

qiling/extensions/coverage/formats/ezcov.py
```diff
@@ -0,0 +1,59 @@
...
+# Cross Platform and Multi Architecture Advanced Binary Emulation Framework
+#
+
+from os.path import basename
+
+from .base import QlBaseCoverage
+
+
+# Adapted from https://github.com/nccgroup/Cartographer/blob/main/EZCOV.md#coverage-data
+class bb_entry(dict):
```
elicn, line 12: Not sure why this class is inheriting `dict`. Is it being used as a dictionary somewhere else?
If this is only to allow arbitrary fields such as `offset`, `size` and `mod_id`, consider declaring that as a `namedtuple` instead.

qiling/extensions/coverage/formats/ezcov.py
```diff
@@ -0,0 +1,59 @@
...
+    def __init__(self, offset, size, mod_id):
```
elicn, line 13: This class is being created in runtime. Probably it would be better to make it as fast as possible (to minimize performance penalty) by moving the calculations / string manipulations to the `csvline` method, which is called once after the execution has ended.

elicn (changes requested):

TheZ3ro:
I force-pushed rebasing this PR on your latest commit in `dev` branch.
Should be good to go now. Let me know if I need to change something else 👍

elicn (approved):
Thanks for the contribution!

xwings (approved):

xwings:
thank you!

PR #1400, opened 2023-10-20
fix errno bug

ucgJhe:
this is actually related to #1341, after dig more deeper, I found that the musl loader will try to locate and map library according to the path described in the file `/etc/ld-musl-$ARCH-.path`, the behavior of the binary is not correct due to the `__do_open` always return -1 on syscall error, and it will fail the process mapping library during runtime.

1.  use proper errno instead of always return -1 in __do_open

### DIFF
--- a/qiling/os/posix/syscall/fcntl.py
+++ b/qiling/os/posix/syscall/fcntl.py
@@ -33,8 +33,8 @@ def __do_open(ql: Qiling, absvpath: str, flags: int, mode: int) -> int:
 
     try:
         ql.os.fd[idx] = ql.os.fs_mapper.open_ql_file(absvpath, flags, mode)
-    except QlSyscallError:
-        return -1
+    except QlSyscallError as err:
+        return -err.errno
 
     return idx
 


elicn (commented):
Proper code should test for negative value / non-negative value rather than looking for a specific one. Anyway, the `QlSyscallError` reflects the error code on the hosting environment; it might work if you emulate a Linux binary on top of a Linux host, but not if you do it on top of Windows.

For that to make sense, `QlSyscallError` should reflect a proper Linux error code rather than host-os error code.

ucgJhe:
close this for now until we come out with a better solution for all arch.

elicn:
Not sure why you closed it.
It is still an issue we should take care of.

ucgJhe:
> Not sure why you closed it. It is still an issue we should take care of.

just file an issue for tracking this.

DockTree:
In my opinion, write the path of the unfound file in /etc/ld-musl-$ARCH-.path can solve the error, however, at the same time, the different base address of stack or other mapping areas would be a cause of other errors.

ucgJhe:
> In my opinion, write the path of the unfound file in /etc/ld-musl-$ARCH-.path can solve the error, however, at the same time, the different base address of stack or other mapping areas would be a cause of other errors.

To create the unfound file could serve as a workaround for the related issue, similar to what we typically do in contexts where a symlink is broken. However, this errno issue leading to incorrect behavior still needs to be fixed, so that the error-catching mechanism can work as expected.

PR #1401, opened 2023-10-29
enhance del_mapinfo/change_mapinfo

HyperSine:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

I was trying to emulate a ELF packed by virbox protector recently. I found there were some error logs when handling syscall_mprotect. Because I cannot post the ELF file on github, here's the minimum code to reproduce:

```py
#!/usr/bin/env python3
import qiling
import qiling.os.posix.syscall

ql = qiling.Qiling([ '/bin/ls' ], rootfs = '/')

ql.mem.map(0xdead0000, 0x4000, 3, 'manual')
ql.mem.map(0xdead4000, 0x4000, 7, 'manual')
ql.mem.map(0xdead8000, 0x4000, 1, 'manual')

print('before:')
print('\n'.join(filter(lambda s: 'manual' in s, ql.mem.get_formatted_mapinfo())))

qiling.os.posix.syscall.ql_syscall_mprotect(ql, 0xdead2000, 0x4000, 0x1)

print('after:')
print('\n'.join(filter(lambda s: 'manual' in s, ql.mem.get_formatted_mapinfo())))
```
Output:
```bash
$ ./code.py 
before:
0000000000dead0000 - 0000000000dead4000   rw-     manual                 
0000000000dead4000 - 0000000000dead8000   rwx     manual                 
0000000000dead8000 - 0000000000deadc000   r--     manual                 
[x]     Cannot change mapinfo at 0xdead2000-0xdead6000
after:
0000000000dead0000 - 0000000000dead4000   rw-     manual                 
0000000000dead4000 - 0000000000dead8000   rwx     manual                 
0000000000dead8000 - 0000000000deadc000   r--     manual
```

After some investigation, I found that the function `change_mapinfo()` in `qiling/os/memory.py` could only change only one MapInfoEntry at a time and that MapInfoEntry must be fully contained by memory range [mem_s, mem_e). If multiple MapInfoEntry just overlap but all of them are not fully contained by the memory range, such error would appear. So I did some enhancement to change_mapinfo and other functions. Here is what this PR contains:
1. Add a function `find_mapinfo()` as it would be used in both `del_mapinfo()` and `change_mapinfo()`.
2. Use `find_mapinfo()` to calculate `overlap_ranges` in `del_mapinfo()`. It has less iterations so we can have better performance.
3. Use direct insert when add new entries in the end of `del_mapinfo()`. The new entries to be added are all parts of entries removed before. So just insert new entries at index from `i0`, no need to call `bisect.insort`.
4. Rewrite `change_mapinfo()`. Now it could change multiple overlapping entries' permissions/label at a time.
5. Swap the order of `mem_unmap()` and `del_mapinfo()` in `unmap()` function for better exception safety.

After applying this PR, the code above wound have correct output:
```bash
$ ./code.py
before
0000000000dead0000 - 0000000000dead4000   rw-     manual                 
0000000000dead4000 - 0000000000dead8000   rwx     manual                 
0000000000dead8000 - 0000000000deadc000   r--     manual                 
after:
0000000000dead0000 - 0000000000dead2000   rw-     manual                 
0000000000dead2000 - 0000000000dead6000   r--     manual                 
0000000000dead6000 - 0000000000dead8000   rwx     manual                 
0000000000dead8000 - 0000000000deadc000   r--     manual   
```

### DIFF
--- a/qiling/os/memory.py
+++ b/qiling/os/memory.py
@@ -4,6 +4,7 @@
 #
 
 import bisect
+import itertools
 import os
 import re
 from typing import Any, Callable, Iterator, List, Mapping, Optional, Pattern, Sequence, Tuple, Union
@@ -81,7 +82,8 @@ class QlMemoryManager:
         self.__write_string(addr, value, encoding)
 
     def add_mapinfo(self, mem_s: int, mem_e: int, mem_p: int, mem_info: str, is_mmio: bool = False):
-        """Add a new memory range to map.
+        """Add a new memory range to map. Caller must guarantee that
+        the map info to be added DOES NOT overlap with any existing map infos.
 
         Args:
             mem_s: memory range start
@@ -93,6 +95,36 @@ class QlMemoryManager:
 
         bisect.insort(self.map_info, (mem_s, mem_e, mem_p, mem_info, is_mmio))
 
+    def find_mapinfo(self, mem_s: int, mem_e: int) -> List[int]:
+        """find map infos that overlap with a memory range
+
+        Args:
+            mem_s: memory range start
+            mem_e: memory range end
+
+        Returns: indices of map infos that overlap with the memory range [mem_s, mem_e)
+        """
+        
+        retval = []
+
+        # index of first (lbound, ubound, ...) that is >= (mem_s,)
+        # i.e., lbound >= mem_s
+        idx = bisect.bisect_left(self.map_info, (mem_s,))
+
+        # check the previous one, its lbound is bound to be < mem_s
+        # so, check its ubound only
+        if idx > 0 and mem_s < self.map_info[idx - 1][1]:
+            retval.append(idx - 1)
+        
+        for i in range(idx, len(self.map_info)):
+            if self.map_info[i][0] < mem_e:
+                retval.append(i)
+            else:
+                # the lbounds of map infos left are all bound to be >= mem_e, so skip check
+                break
+        
+        return retval
+
     def del_mapinfo(self, mem_s: int, mem_e: int):
         """Subtract a memory range from map.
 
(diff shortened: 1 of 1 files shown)


xwings:
Cool, that make sense.

xwings:
Hi, 

The code seems to break the test. I need to revert it. Will you be able to make another PR ?

HyperSine:
> Hi,
> 
> The code seems to break the test. I need to revert it. Will you be able to make another PR ?

Can you tell me which test is broken?

xwings:
https://github.com/qilingframework/qiling/actions/runs/7208092719/job/19636291311

HyperSine:
I think the problem is not from my code. It is due to the bug from upstream: https://github.com/unicorn-engine/unicorn/issues/1877.

In `test_qdb_mips32el_hello`, `ql` would map memory above 0x90000000 because of `ql.loader.mmap_address`. But it would call `ql.mem.unmap_all()` later which relies on `ql.uc.mem_regions()`. The function `mem_regions()` doesn't return a correct result and causes the test failed. Specifically, 

1. Some method in ql calls `ql.mem.map(0x90000000, ...)`, but unicorn engine treats it as memory starting from 0x10000000 because of the bug I mentioned above. Our `ql.mem.map_info` will have the record starting from 0x90000000.
2. Later, some method in ql calls `ql.mem.unmap_all()`, which retrieves memory regions from unicorn. So there will be a memory region starting from 0x10000000, and `ql.mem.del_mapinfo(0x10000000, ...)` will be called. Our `ql.mem.map_info` doesn't have such record, so `del_mapinfo` does nothing. But unicorn will still unmap such memory region.
3. Later, some method in ql calls `ql.mem.save()`. It iterates `ql.mem.map_info` and tries to read the undeleted memory record starting from 0x90000000 which, in unicorn's perspective, is 0x10000000, and of course, is already unmapped. So a UcError is raised.

If you modify the test code `tests/test_qdb.py`:

```py
# from line 13
    def test_qdb_mips32el_hello(self):
        rootfs = "../examples/rootfs/mips32el_linux"
        path = rootfs + "/bin/mips32el_hello"

        ql = Qiling([path], rootfs)
        ql.loader.mmap_address = 0x20000000     # <-- add just this line
        ql.debugger = "qdb::rr:qdb_scripts/mips32el.qdb"
        ql.run()
        del ql
```
you will find my PR passes the test.

PR #1407, opened 2023-11-08
Qdb improvements

ucgJhe:
1. allow multi-breakpoints
we can now use some thing like `ql.debugger = "qdb:0xAAAA:0xBBBB:0xCCCC"` for setting up multiple breakpoints 

2. integrity of hooks and patches
with this commit, all hooks from ql.os should be able to triggered as expected

3. add simple filter for output of command show , also show both entry points of elf and ld if possible

image

### DIFF
--- a/qiling/debugger/qdb/qdb.py
+++ b/qiling/debugger/qdb/qdb.py
@@ -9,7 +9,7 @@ from typing import Optional, Tuple, Union, List
 from contextlib import contextmanager
 
 from qiling import Qiling
-from qiling.const import QL_OS, QL_ARCH, QL_ENDIAN, QL_STATE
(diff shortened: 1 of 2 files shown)


ucgJhe:
https://github.com/qilingframework/qiling/pull/1407/commits/6799389ed3098fb4c0ac5a4e5cef95ae907ca7be#diff-090cd519aa623cfcd98762678b4150bd7a46b3e559fac27eb84a7eec7a703c12R101-R105

The ql.mem.unmap_all will failed here due to the MIPS memory layout issue of Unicorn engine, but it's already fixed in the latest dev brach . So I simply ignore it for now.

xwings (approved):

Issue #1416, opened 2023-12-07
UEFI does not work with qdb


dotCirill:
The code (emulation of Combined DXE/SMM module):

```python
from qiling import Qiling

if __name__ == "__main__":
    ql = Qiling([
        "files/SuperSMM_DXE_module.efi"
    ], "files")
    ql.debugger = "qdb"


    ql.run()
```

On `1.4.6` release and current `master` branch it dies with this error:

```
... init logs ...
Qdb> *several n (next-steps)*
AttributeError: 'QlOsUefi' object has no attribute 'smm'
```

backtrace:

```py
# qiling/core.py
debugger(self) # self.os.run is below
# qiling/debugger/qdb/qdb.py
elf.dbg_hook(init_hook)
self.interactive()
self.cmdloop()
...
# qiling/core_hooks.py
hook.call(ql)
# qiling/core_hooks_types.py
self.callback(ql, *args)
# qiling/os/uefi/fncc.py
ql.os.call(pc, f, params, onenter, onexit)
self.fcall.call(func, proto, args, onenter, onexit, passthru)
# qiling/os/fcall.py
func(ql, pc, params)
# qiling/os/uefi/protocols/EfiSmmBase2Protocol.py
ql.log.debug(f'InSmram = {ql.os.smm.active}')
```

As I understand ql.os.run should be runned but it isn't.

On `dev` branch I have this backtrace:

```py
ql.run()
# qiling/core.py
debugger = debugger(self) 
# qiling/debugger/qdb/qdb.py
if addr_elf_entry := getattr(self.ql.loader, 'elf_entry'):

AttributeError: 'QlLoaderPE_UEFI' object has no attribute 'elf_entry'
```

ucgJhe:
Hi,  I think it should be `getattr(self.ql.loader, 'elf_entry', None)`, please let me know if this fix the issue.

dotCirill:
Hi, it seems to be OK with that

loki52501:
is this fixed?

dotCirill:
> is this fixed?

Yes qdb works after the fix

PR #1427, opened 2023-12-25
Fix getattr self.ql.loader default value

dotCirill:
https://github.com/qilingframework/qiling/issues/1416 fix


elicn (approved):
I suspect this implementation does not support `0` as a valid entry point, since both `0` and `None` are considered as `False`. We could change that to query `is not None` explicitly, but maybe that is good enough.

xwings (approved):

PR #1434, opened 2024-01-08
Remove EVM architecture

### LINKED ISSUE #1379: qiling framework has a potential issue when handling `env` parameters of type bytes.

mochaaP:
Remove EVM architecture along its accompanying tests and examples.


qiling/debugger/gdb/gdb.py
```diff
@@ -444,21 +444,20 @@ def handle_q(subcmd: str) -> Reply:
...
-                    # filesystem dependent features
-                    if hasattr(self.ql.os, 'path'):
-                        features += [
-                            'QSetWorkingDir+',
-                            'qXfer:auxv:read+',
-                            'qXfer:exec-file:read+'
-                        ]
-
                     # process dependent features
                     if hasattr(self.ql.os, 'pid'):
```
elicn, line 462: That one was also under `if not self.ql.interpreter`, so it should be re-indented appropairately.

qiling/utils.py
```diff
@@ -376,9 +375,7 @@ def __int_nothrow(v: str, /) -> Optional[int]:
 def select_arch(archtype: QL_ARCH, cputype: Optional[QL_CPU], endian: QL_ENDIAN, thumb: bool) -> QlClassInit['QlArch']:
     kwargs = {}
 
-    # skip cpu model for evm
-    if archtype is not QL_ARCH.EVM:
-        kwargs['cputype'] = cputype
+    kwargs['cputype'] = cputype
```
elicn, line 378: That can be moved to be part of the initialization two lines above, that is:
`kwargs = {'cputype': cputype}`

elicn (changes requested):

elicn (approved):

xwings (approved):

xwings:
This is amazing. Thanks !

PR #1439, opened 2024-01-23
Fixed DOS Disk Mapping

racerxdl:
This fixes the DOS Mappings and makes 8086 examples work again.

This is probably not the best way to fix them, but at least it works fine with current FS Mapper. Basically what it does is map 0x80 to C and beyond and 0x00 to A and beyond.

There is few limitations though: INT 13H actually allows up to 126 disks, this only supports disks from C to Z. INT13 also supports floppies in CDEF... letters, this PR only supports A and B as floppy letters.

Although this is not ideal, I don't think I ever saw something actually uses more than two floppies and few disks. At least this make qiling able to simulate DOS.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----


xwings:
@elicn any update on this ?

elicn:
Unfortunately not.
Was busy maintaining other parts in Qiling (:

PR #1456, opened 2024-03-06
Enable more flexible logging redirection

elicn:
Modified logger setup to allow more flexible logging redirection and multiplexing.

## Interface changes
The current interface in which `log_file` is used to optionally specify a filename to log into:
```python
ql = Qiling(argv, rootfs, log_file=r'/tmp/myprog.log')
```
Was changed into an optional variable that accepts a `Collection` of either filenames or open streams:
```python
ql = Qiling(argv, rootfs, log_devices=[sys.stderr, r'/tmp/myprog.log'])
```

### Notes
 - While the current design creates a file logger **in addition** to the normal `sys.stderr` logger, the new implementation replaces it. That is, if a list of log devices is specified there would be be no `sys.stderr` logger by default anymore, unless it is specified (as seen in the example above)
 - The default behavior (i.e. not specifying any `log_file` / `log_devices`) is maintained: only `sys.stderr` will be created
 - Specifying an empty `Collection`, such as `log_devices=[]`, would result in no logger (same as setting `console=False`)
 - For that reason the `console` argument has no use anymore. However the `console` argument was not removed to maintain compatibility and maybe will be used in the future for other purposes

## Advantages
It is now way easier to open multiple loggers and redirect them to any file, stream or socket. This is very useful when trying to separate the emulated program output from logging output, or enable logging while fuzzing (which is currently hidden "behind" AFL box).

In the following example the logging output is redirected to another terminal pane to separate program output from logging output:
```python
ql = Qiling(argv, rootfs, log_devices=['/dev/pts/6'])
```

image

### DIFF
--- a/qiling/core.py
+++ b/qiling/core.py
@@ -4,9 +4,11 @@
 #
 
 import os
+import sys
 import pickle
+
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, AnyStr, List, Mapping, MutableMapping, Optional, Sequence, Tuple, Union
+from typing import TYPE_CHECKING, Any, AnyStr, Collection, IO, List, Mapping, MutableMapping, Optional, Sequence, Tuple, Union
 
 # See https://stackoverflow.com/questions/39740632/python-type-hinting-without-cyclic-imports
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


xwings (approved):

xwings:
Cool idea!

kabeor (approved):

PR #1462, opened 2024-03-13
Fix and improve AFL fuzzing

elicn:
Highlights:
 - Revamped the AFL extension: cleaned unnecessary code and added more comments
 - Refreshed and fixed AFL linux example
 - Added `QlFuzzer` extension to simplify the implementation of Qiling-based custom fuzzers
   - The class is heavily documented to help demystifying fuzzing
   - Added a usage example

### DIFF
--- a/qiling/extensions/afl/qlfuzzer.py
+++ b/qiling/extensions/afl/qlfuzzer.py
@@ -1,0 +1,157 @@
+# @author: elicn
+
+import os
+
+from abc import ABC, abstractmethod
+from typing import Any, Collection, Dict, Optional, Sequence
+
+from qiling import Qiling
+from qiling.const import QL_VERBOSE
+from qiling.extensions import afl
+from qiling.os.const import POINTER
+
+
+class QlFuzzer(ABC):
+    """Simplify Qiling-based fuzzing.
+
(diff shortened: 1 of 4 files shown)


qiling/extensions/afl/qlfuzzer.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+
+        image = self.ql.loader.get_image_by_name(module or os.path.basename(self.ql.argv[0]), casefold=casefold)
+
+        if image is None:
+            raise KeyError(f'could not find a loaded module named "{module}"')
+
+        return image.base + offset
+
+    def setup(self, infilename: str, entry: int, exits: Collection[int], crashes: Optional[Collection[int]] = None) -> None:
```
wtdcode, line 122: It would be better if `infile` can be multiple files though our cases always use single one.

wtdcode (approved):

qiling/extensions/afl/qlfuzzer.py
elicn: How would you suggest to do that?
The `uc_afl_fuzz_custom` API in `unicornafl` takes just one filename.

kabeor (approved):

PR #1483, opened 2024-07-13
Qdb bug fix missing function `handle_bnj_arm`

ucgJhe:
fix miss-handled funciton `handle_bnj_arm`

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -15,7 +15,7 @@ jobs:
         run: pipx install poetry
       - name: Build distribution 📦
         run: poetry build -n
-      - uses: actions/upload-artifact@v2
+      - uses: actions/upload-artifact@v4
         with:
           path: ${{ github.workspace }}/dist/*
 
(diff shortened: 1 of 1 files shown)


wtdcode (approved):

qiling/debugger/qdb/branch_predictor/branch_predictor_arm.py
```diff
@@ -40,9 +43,9 @@ def get_cpsr(bits: int) -> (bool, bool, bool, bool):
                 bits & 0x80000000 != 0, # N, sign flag
                 )
 
-    def predict(self):
+    def predict(self, pref_addr: Optional[int] = None) -> BranchPredictor.Prophecy:
         prophecy = self.Prophecy()
-        cur_addr = self.cur_addr
+        cur_addr = pref_addr or self.cur_addr
```
elicn, line 48: Please note that the `or` operator will pick the latter not only when the former is `None`, but also when it is `0`. If `0` is a valid value for `pref_addr`, then consider using:

```python
cur_addr = self.cur_addr if pref_addr is None else pref_addr
```

ucgJhe: thx for the hint, i will correct this in the following fix for the issue

xwings: @ucgJhe please update the code ?

PR #1500, opened 2024-10-11
✨ feature: add new HookSwitch hook type.

Paul2803k:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### DIFF
--- a/qiling/core_hooks.py
+++ b/qiling/core_hooks.py
@@ -43,7 +43,7 @@ from unicorn.unicorn_const import (
     UC_HOOK_INSN_INVALID
 )
 
-from .core_hooks_types import Hook, HookAddr, HookIntr, HookRet
+from .core_hooks_types import Hook, HookAddr, HookIntr, HookRet, HookSwitch
 from .const import QL_HOOK_BLOCK
 from .exception import QlErrorCoreHook
 
@@ -397,6 +397,12 @@ class QlCoreHooks:
 
         return HookRet(self, hook_type, hook)
 
+    def ql_switch_hook(self, callback: Callable, user_data: Any = None, begin: int = 1, end: int = 0) -> HookRet:
+        hook = HookSwitch(callback, user_data, begin, end)
+        self._ql_hook(UC_HOOK_CODE, hook)
+
+        return HookRet(self, UC_HOOK_CODE, hook)
+
     def hook_code(self, callback: TraceHookCalback, user_data: Any = None, begin: int = 1, end: int = 0) -> HookRet:
         """Intercept assembly instructions before they get executed.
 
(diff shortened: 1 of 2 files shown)


xwings:
Hi,

Welcome to Qiling Framework. I guess for this case we also need a test to make sure everything runs properly. (Current version of Unicorn is still broken and we are still fixing it)

Paul2803k:
> Hi,
> 
> Welcome to Qiling Framework. I guess for this case we also need a test to make sure everything runs properly. (Current version of Unicorn is still broken and we are still fixing it)

Hey! Thanks for replying! I don't mind adding a test or two.  Should I expand this test to also include the new hook? https://github.com/qilingframework/qiling/blob/9a78d186c97d6ff42d7df31155dda2cd9e1a7fe3/tests/test_evm.py#L34 This seems to me one of the few places where hooks are tested.

elicn:
This is indeed an interesting idea, but let me suggest another view to this.
What if that could be more generic (say, a decorator) and used to define any kind of hook?

Consider something similar to this:
```python
@toggle_hook(on=0x1000, off=0x2000)
def my_hook(ql: Qiling, *args):
   ...
```

Paul2803k:
> This is indeed an interesting idea, but let me suggest another view to this. What if that could be more generic (say, a decorator) and used to define any kind of hook?
> 
> Consider something similar to this:
> 
> ```python
> @toggle_hook(on=0x1000, off=0x2000)
> def my_hook(ql: Qiling, *args):
>    ...
> ```

I like the idea! Should I just give it a try or do you have already some requirements or some idea on how it should interact with the other types of hooks?

elicn:
I didn't think it through, maybe a code hook is always required (otherwise, how can we tell when to toggle it on or off?)

Please go ahead and implement what you think is useful. Consider placing that in the "extensions" folder as this is still experimental.

Paul2803k:
> I didn't think it through, maybe a code hook is always required (otherwise, how can we tell when to toggle it on or off?)
> 
> Please go ahead and implement what you think is useful. Consider placing that in the "extensions" folder as this is still experimental.

Hey! Sorry for the late reply, it has been a busy month. It is not very clear to me how I would be able to move the implementation to the "extensions" folder. The hooks are declared functions of the class `QlCoreHooks`  which keeps track of the registered hooks and provides internal methods to register new ones. If I had to move everything outside of it I would have to pass an instance of the `Qiling` class, to have access to the internal `_ql_hook` method. The whole thing would look something like this, so slightly different from the existing hook methods. 

`
class HookSwitch(Hook):
    def __init__(self, callback, user_data=None, begin: int = 1, end: int = 0) -> None:
        super().__init__(callback=callback, user_data=user_data, begin=begin, end=end)
        self.switch = False

    def bound_check(self, pc: int, size: int = 1) -> bool:
        if self.begin == pc and not self.switch:
            self.switch = True
        if self.end == pc and self.switch:
            self.switch = False
        return self.switch


def ql_hook_switch(
    ql: Qiling, callback: Callable, user_data: Any = None, begin: int = 1, end: int = 0
) -> HookRet:
    hook = HookSwitch(callback=callback, user_data=user_data, begin=begin, end=end)
    ql._ql_hook(UC_HOOK_CODE, hook)

    return HookRet(ql=ql, hook_type=UC_HOOK_CODE, hook_obj=hook)


def hook_switch(
    ql: Qiling,
    callback: TraceHookCalback,
    user_data: Any = None,
    begin: int = 1,
    end: int = 0,
) -> HookRet:
    return ql_hook_switch(
        ql=ql, callback=callback, user_data=user_data, begin=begin, end=end
    )
`
If you have better recommendations pls let me know :)

xwings:
Since this is a new hook, i still recomment we need to add in a new test too.

Paul2803k:
> Since this is a new hook, i still recomment we need to add in a new test too.

Will add one soon! Any advice on where to place it? Maybe around where the other hooks are tested?

xwings:
Hi, 

somewhere around https://github.com/qilingframework/qiling/blob/master/tests/test_elf.py will work

PR #1526, opened 2025-01-09
Fix: gdb cannot parse some regs value when debugging ARM64

the-soloist:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

edit: test with gdb 12.1
image

gdb-multiarch is unable to correctly retrieve the values of the cspr, fpsr, and fpcr registers when debugging programs on the ARM64 architecture because Qiling has not defined them.

image

After modification, it can be correctly parsed.

image

```py
import sys
import qiling
import struct
from pathlib import Path
from qiling import Qiling
from qiling.debugger.qdb import QlQdb
from qiling.os.mapper import QlFsMappedObject


def main():
    ql = Qiling(["./rootfs/bin/arm64_hello"],
                rootfs="./rootfs",
                verbose=QL_VERBOSE.DEBUG)
    ql.debugger = True
    ql.run()


if __name__ == "__main__":
    main()
```

- https://community.arm.com/support-forums/f/infrastructure-solutions-forum/48875/armv8-why-does-not-use-current-program-status-register-cpsr


xwings:
There is a conlict. appricate if you can update it.

the-soloist:
Conflicts already resolved.

elicn:
These changes are not aligned with latest Unicorn.
Since it 2.1.x release Unicorn has dropped various registers and also changed the ways some of them are accessed.

the-soloist:
Edit: Sorry, I didn't notice "deprecated".

elicn:
I gave it a deeper look and since CP regs are no longer accessible through `arch.regs`, the change is not quite straightforward. CP regs are no longer supported in gdb, and we'll need to fix that.

wtdcode:
These registers seem not deprecated to access directly? They are not cp registers.

Ref: https://github.com/unicorn-engine/unicorn/blob/64c72267aed15e8a47dec7b9d9df6d9e36459757/qemu/target/arm/unicorn_aarch64.c#L262

elicn:
> These registers seem not deprecated to access directly? They are not cp registers.

Well.. I am following Unicorn's Python bindings constants and this is quite confusing TBH. [FPSR](https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Registers/FPSR--Floating-point-Status-Register?lang=en) and [FPCR](https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Registers/FPCR--Floating-point-Control-Register?lang=en) are considered special registers and if you follow the Unicorn logic of changing the way special regs are accessed, it makes sense to consider them too.

Besides that, the Python const file is not consistent with Unicorn's C implementation; for example, PSTATE is considered [deprecated](https://github.com/unicorn-engine/unicorn/blob/df75effba35b64a7442be42c2d4fd39707ee20a2/bindings/python/unicorn/arm64_const.py#L282) on the Python const file, but [still appears](https://github.com/unicorn-engine/unicorn/blob/df75effba35b64a7442be42c2d4fd39707ee20a2/qemu/target/arm/unicorn_aarch64.c#L238) on the C file you linked. Same goes for TTBR0_EL1 and TTBR1_EL1 and a few others.

Will appreciate a more consistent approach that will not lead me to provide silly answers on github (:
Let me know how you would like to consider FPSR, FPCR and PSTATE.

wtdcode:
> > These registers seem not deprecated to access directly? They are not cp registers.
> 
> Well.. I am following Unicorn's Python bindings constants and this is quite confusing TBH. [FPSR](https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Registers/FPSR--Floating-point-Status-Register?lang=en) and [FPCR](https://developer.arm.com/documentation/ddi0601/2024-12/AArch64-Registers/FPCR--Floating-point-Control-Register?lang=en) are considered special registers and if you follow the Unicorn logic of changing the way special regs are accessed, it makes sense to consider them too.

Oh.... You are correct. They should be accessed via `UC_ARM64_REG_CP_REG` indeed.

> 
> Besides that, the Python const file is not consistent with Unicorn's C implementation; for example, PSTATE is considered [deprecated](https://github.com/unicorn-engine/unicorn/blob/df75effba35b64a7442be42c2d4fd39707ee20a2/bindings/python/unicorn/arm64_const.py#L282) on the Python const file, but [still appears](https://github.com/unicorn-engine/unicorn/blob/df75effba35b64a7442be42c2d4fd39707ee20a2/qemu/target/arm/unicorn_aarch64.c#L238) on the C file you linked. Same goes for TTBR0_EL1 and TTBR1_EL1 and a few others.
> 
> Will appreciate a more consistent approach that will not lead me to provide silly answers on github (: Let me know how you would like to consider FPSR, FPCR and PSTATE.

Generally your answer is correct and I just forget the semantic here (though implemented by myself =/). 

Regarding these "deprecated" registers, they won't be removed in any future releases and no warning will be printed as well. I think I should mark these registers as alias instead of deprecated to reflect these.

qiling/arch/arm64_const.py
```diff
@@ -68,6 +68,9 @@
     "pc": UC_ARM64_REG_PC,
     "lr": UC_ARM64_REG_LR,
     "cpacr_el1": UC_ARM64_REG_CPACR_EL1,
+    "tpidr_el0": UC_ARM64_REG_TPIDR_EL0,
```
elicn: Remove "tpidr_el0" and "cpsr" entries

elicn (changes requested):
Per @wtdcode, FPSR, FPCR and PSTATE are valid registers so after removing the obsolete ones, we should be good to go.

xwings (approved):

elicn (approved):

PR #1531, opened 2025-01-27
Implement kernel ARM emulation

McCaulay:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

These fixes add support for emulating a ARM kernel object by implementing `R_ARM_CALL` and `R_ARM_JUMP24`. Additionally, the `__get_symbol` function now has caching in order to speed up loading ARM kernel objects as elftools `get_symbol_by_name` function is slow when iterating each time.

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -21,6 +21,7 @@ from qiling import Qiling
 from qiling.const import QL_ARCH, QL_ENDIAN, QL_OS
 from qiling.exception import QlErrorELFFormat, QlMemoryMappedError
 from qiling.loader.loader import QlLoader, Image
+from qiling.os.memory import QlMemoryHeap
 from qiling.os.linux.function_hook import FunctionHook
 from qiling.os.linux.syscall_nums import SYSCALL_NR
 from qiling.os.linux.kernel_api.hook import *
@@ -57,7 +58,7 @@ class AUXV(IntEnum):
 
 # start area memory for API hooking
 # we will reserve 0x1000 bytes for this (which contains multiple slots of 4/8 bytes, each for one api)
-API_HOOK_MEM = 0x1000000
+API_HOOK_MEM = 0x2000000
 
 # memory for syscall table
 SYSCALL_MEM = API_HOOK_MEM + 0x1000
@@ -87,6 +88,9 @@ class QlLoaderELF(QlLoader):
         stack_size = self.profile.getint('stack_size')
         self.ql.mem.map(stack_address, stack_size, info='[stack]')
 
+        # Setup heap
+        self.ql.os.heap = QlMemoryHeap(self.ql, 0x3000000, 0x3000000 + 0x1000000)
+
         self.path = self.ql.path
 
         with open(self.path, 'rb') as infile:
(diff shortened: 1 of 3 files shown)


xwings:
Hi,

there is a conflict in elf.py. You might want to take a look ?

qiling/loader/elf.py
```diff
@@ -57,7 +58,7 @@ class AUXV(IntEnum):
 
 # start area memory for API hooking
 # we will reserve 0x1000 bytes for this (which contains multiple slots of 4/8 bytes, each for one api)
-API_HOOK_MEM = 0x1000000
+API_HOOK_MEM = 0x2000000
```
elicn, line 61: Any reason for the change?

qiling/loader/elf.py
```diff
@@ -87,6 +88,9 @@ def run(self):
         stack_size = self.profile.getint('stack_size')
         self.ql.mem.map(stack_address, stack_size, info='[stack]')
 
+        # Setup heap
+        self.ql.os.heap = QlMemoryHeap(self.ql, 0x3000000, 0x3000000 + 0x1000000)
```
elicn, line 92: Heap base and size should be set based on the value configured in the profile file. Using hardcoded values is not a good idea.

qiling/loader/elf.py
```diff
@@ -97,7 +101,7 @@ def run(self):
 
         # is it a driver?
         if elftype == 'ET_REL':
-            self.load_driver(elffile, stack_address + stack_size, loadbase=0x8000000)
+            self.load_driver(elffile, stack_address + stack_size, loadbase=0x1000000)
```
elicn, line 104: Any reason for this change here?
Keep in mind this module serves all ELF files on all architectures, not only ARM.

qiling/loader/elf.py
```diff
@@ -414,10 +418,19 @@ def lkm_get_init(self, elffile: ELFFile) -> int:
         raise QlErrorELFFormat('invalid module: symbol init_module not found')
 
     def lkm_dynlinker(self, elffile: ELFFile, mem_start: int) -> Mapping[str, int]:
+        self._symbol_name_map = None
         def __get_symbol(name: str) -> Optional[Symbol]:
             _symtab = elffile.get_section_by_name('.symtab')
-            _sym = _symtab.get_symbol_by_name(name)
 
+            # Cache
+            if self._symbol_name_map == None:
```
elicn, line 426: This could be simplified with an internal function decorated with `@cache` (available from Python 3.9) or `@lru_cache` (available earlier).

```python
from functools import cache

@cache
def __get_cached_symbol(name: str) -> int:
    # access symtab symbol here and return the result
    ...
```

Also, I am not sure that `_symbol_name_map` should be assigned to `self`, but stay local.

qiling/os/linux/kernel_api/hook.py
```diff
@@ -25,6 +25,11 @@ def hook_kernel_api(ql: Qiling, address: int, size):
         if api_func:
             try:
                 api_func(ql, address, api_name)
+
+                # Restore PC
+                if ql.arch.type == QL_ARCH.ARM:
```
elicn, line 30: I think function calls are automatically unwinded.
Have you checked this is?

elicn (changes requested):

PR #1537, opened 2025-03-03
Fixing UNIX sockets error

elicn:
The current temporary path under which UNIX sockets are created gets too long (more than 107 chars) which causes tests to fail with an "AF_UNIX path too long" error. This is fixed by specifying `$TMPDIR` to be a shorter path.


elicn:
This is causing some weird unexpected errors. Closing.

PR #1552, opened 2025-03-24
Various fixes and improvements for Windows emulation

JFarAur:

### Coding convention?

### Extra tests?

- [X] Tests will be added after some discussion and review.

### Changelog?

### Target branch?

### One last thing

-----

### Summary
In this PR, the following is addressed for Windows emulation on x86 and x86_64:
* Fixed several issues preventing C++ runtime initialization.
* Programs using the C++ runtime and standard input/output streams are emulated with better correctness.
* Improved support for several C++ runtime features, including `typeid` and `dynamic_cast`.
* Added support for forwarded exports to the PE loader.
  - This fixes many Windows API functions. For example, interlocked SList operations (`InterlockedPushEntrySList`, etc) are now functioning properly.
  - Hooks on forwarded functions cannot be easily "bypassed" by calls to a forwarding target. For example, calls to `RtlAllocateHeap` and `HeapAlloc` will trigger the same Qiling hook.

The following is also addressed, on x86_64 only:
* Improved support for software user-mode exceptions, including C++ exceptions. Programs making use of try.. catch are emulated with better correctness.

### Details
**(Original 24 March)**
Previously, Qiling hooked some functions in msvcrt.dll (`_initterm`, `_initterm_e` and `__acrt_iob_func`) so that they would return right away. This worked fine for programs using the C runtime library, as the startup code was not necessary when running in emulation. However, this caused some problems for C++ programs, namely programs using C++ runtime DLLs, for example msvcp140.dll. During my testing, a program printing to std::cout or std::err using operator<< was unable to print to the terminal. This was because of some global variables which were necessary for standard streams to function properly, yet were uninitialized because the CRT startup code did not fully complete due to the hooks.

Leaving these functions unhooked, and implementing `HeapReAlloc` and `_realloc_base` allowed my test C++ program to run without crashing. Qiling loads the C++ runtime library without crashing, and the program is able to print using the standard input/output streams. However, removing the CRT startup hooks breaks msvcrt.dll DllMain, which crashes when a program that uses msvcrt loads the module.

After some analysis, I have found that the error during CRT startup in programs using msvcrt.dll when Qiling CRT startup hooks are removed is due to a part of the C runtime startup code which obtains a pointer from `RtlPcToFileHeader` which is invalid.

After implementing a hook for `RtlPcToFileHeader` in Qiling, the issue is resolved. Both msvcrt and msvcp140 DllMain load without errors, and my test C++ program can successfully print to stdout and stderr. C and C++ programs, even those compiled with recent compilers, initialize CRT successfully and can print to the terminal.

With these additional winapi implementations, the stub hooks for `_initterm`, `_initterm_e` and `__acrt_iob_func` are no longer needed as a stopgap solution, and the library versions of these functions can run without problems.

**(Update 27 March)**
Currently, Qiling has limited support for software exceptions. Windows programs using SEH or C++ language features such as try.. catch are not emulated correctly.

After some analysis, I have found that the main obstacle in emulating software exception handling in Qiling is in several functions in ntdll which make use of various global structures, caches, or loader data, which are in an invalid state during emulation because Qiling does not fully emulate Windows kernel initialization or the loader process.

The most important functions in question are `RtlLookupFunctionEntry` and `RtlLookupFunctionTable`. These functions are used for looking up compiler-generated data used during exception handling, including function locations and stack unwinding instructions.

After implementing hooks for these functions, many of the functions involved with exception handling, including `RtlRaiseException`, `RtlDispatchException`, and stack unwinding functions such as `RtlVirtualUnwind` and `RtlUnwindEx`, are actually emulated correctly. Handler routines are correctly located and executed.

However, there is another obstacle. After the exception is handled and control is returned to the dispatcher, `RtlUnwindEx` calls `RtlRestoreContext`. In the case of C++ exceptions, this function calls `RcFrameConsolidation` which recursively consolidates stack frames before finally restoring the new context with an IRETQ instruction.

Due to the way Qiling sets up the GDT on x86_64, the IRETQ instruction at this point resulted in CPU faults. After making some adjustments to the GDT setup on x86_64, the context switch occurs as expected. All of the tests (at least those I have access to) are still passing, so it seems this new GDT setup does not break anything for other platforms. I believe it is technically more correct on x86_64, also.

Regardless, in my tests, some simple programs which make use of C++ exceptions are now correctly emulated.


JFarAur:
Some more information, the `_realloc_base` hook with proper error checking is necessary because realloc is sometimes used by the CRT with base=NULL, in which case it should behave like a regular malloc. If `_realloc_base` is not hooked then these allocations will fail.

```
[FlsGetValue(dwFlsIndex = 0x2) = 0x0]
[SetLastError(dwErrCode = 0)]
[EnterCriticalSection(lpCriticalSection = 0x1806f0e90) = 0x0]
[_realloc_base(block = 0, size = 0x100) = 0x5000071fc]
[memset(dest = 0x5000071fc, c = 0, count = 0x100) = 0x5000071fc]
[_free_base(address = 0)]
[LeaveCriticalSection(lpCriticalSection = 0x1806f0e90) = 0x0]
[api __acrt_iob_func (ucrtbase) is not implemented]
```

JFarAur:
Regarding `RtlPcToFileHeader`, this function is very important because it is used for various things, for example, when setting up SEH during CRT init, in CxxThrowException, and also in the implementations of type ID and dynamic_cast. By implementing it in Qiling, I would expect C++ programs to work *a lot* better now.

JFarAur:
I would like to elaborate on my previous comments. After doing some more investigation, I can provide very detailed information about the problems that were occurring previously, and why these changes are beneficial.

First, the `_initterm` function's job is essentially to loop over an array of function pointers and execute them. For msvcrt.dll, it appears that the only function executed is responsible for initializing the static member `std::shared_ptr<__ExceptionPtr> m_badAllocExceptionPtr` in `__ExceptionPtr`:

```
shared_ptr<__ExceptionPtr> __ExceptionPtr::_InitBadAllocException()
{
    std::bad_alloc _Except;
    return __ExceptionPtr::_CopyException(&_Except, static_cast<const ThrowInfo*>(__GetExceptionInfo(_Except)), false);
}
```

An `std::bad_alloc` object is constructed on the stack, then `__ExceptionPtr::_CopyException` is called to turn this object, together with a `ThrowInfo*` obtained from `__GetExceptionInfo`, into an `std::shared_ptr<__ExceptionPtr>`.

`__GetExceptionInfo` seems to be a compiler builtin function. In IDA we can see that this becomes a pointer to a static ThrowInfo struct corresponding to the `std::bad_alloc` class.

```
loc_110117DA0:
call    ??0exception@@QEAA@AEBQEBDH@Z ; exception::exception(char const * const &,int)
lea     rdi, ??_7bad_alloc@std@@6B@ ; const std::bad_alloc::`vftable'
mov     [rsp+58h+var_28], rdi
xor     r9d, r9d
lea     r8, _TI2?AVbad_alloc@std@@ ; throw info for 'class std::bad_alloc'
lea     rdx, [rsp+58h+var_28]
mov     rcx, rbx
```

Inside `__ExceptionPtr::_CopyException` is the following code:

```
#if _EH_RELATIVE_OFFSETS && !defined(_M_CEE_PURE)
        PVOID ThrowImageBase = RtlPcToFileHeader((PVOID)pTI, &ThrowImageBase);
        PER_PTHROWIB(pExcept) = ThrowImageBase;
#endif
```

We can see that `RtlPcToFileHeader` is used to obtain the image base of the module where the ThrowInfo struct is located. This is used to initialize the ImageBase field of the ExceptionRecord.

Previously, without the `_initterm` stub hooks, this code would run, but ImageBase would be populated with an invalid pointer, which would result in an error at a later point. Below is output from a program that loads msvcrt.dll showing the use of `RtlPcToFileHeader` and error shortly thereafter.

```
[!]     api _initterm (msvcrt) is not implemented
[!]     api ??0exception@@QEAA@AEBQEBDH@Z (msvcrt) is not implemented
[!]     api RtlPcToFileHeader (ntdll) is not implemented
[!]     api ZwQueryVirtualMemory (ntdll) is not implemented
[=]     malloc(size = 0xa0) = 0x500009842
[=]     memset(dest = 0x500009882, c = 0, count = 0x58) = 0x500009882
[x]     Error encountered while running msvcrt.dll DllMain, bailing
[=]     Done loading msvcrt.dll
```

The ntdll implementation of `RtlPcToFileHeader` looks up the address in the inverted function tables, and calls `ZwQueryVirtualMemory`. However this syscall is not implemented in Qiling, and `RtlPcToFileHeader` will return NULL.

With a correct implementation of `RtlPcToFileHeader`, the CRT startup code can run without problems:

```
[!]     api _initterm (msvcrt) is not implemented
[!]     api ??0exception@@QEAA@AEBQEBDH@Z (msvcrt) is not implemented
[=]     RtlPcToFileHeader(PcValue = 0x110186998, BaseOfImage = 0x80000001cea8) = 0x110100000
[=]     malloc(size = 0xa0) = 0x500009842
[=]     memset(dest = 0x500009882, c = 0, count = 0x58) = 0x500009882
[=]     EncodePointer(Ptr = 0x110186998) = 0x110186998
...
[=]     atexit(func = 0x110175680) = 0x0
[=]     Returned from msvcrt.dll DllMain
[=]     Done loading msvcrt.dll
```

This also means that the stub hooks for `_initterm` and `_initterm_e` to prevent the crash are unnecessary.

For msvcp140.dll, the situation with `_initterm` is more complex. The array given to `_initterm` is larger, and contains dynamic initializers of standard input/output streams such as cout, cerr, and cin. Skipping `_initterm` means that these initializers do not run, meaning input/output streams are never initialized when programs attempt to use them, resulting in errors, most often null pointer exceptions. This is why programs using the C++ runtime library and standard input/output streams were unstable when running in Qiling.

However, with fixed `RtlPcToFileHeader` and removal of the `_initterm` and `_initterm_e` stubs, the C++ runtime library can correctly call its dynamic initializers, including for the standard input/output streams.

Note that `RtlPcToFileHeader` is also used to fill in the ExceptionRecord in `_CxxThrowException`, which is used any time a C++ exception is thrown anywhere:

```
#if _EH_RELATIVE_OFFSETS
        PVOID ThrowImageBase = RtlPcToFileHeader((PVOID)pTI, &ThrowImageBase);
        ThisException.params.pThrowImageBase = ThrowImageBase;
#endif
```

And `RtlPcToFileHeader` is also used in several other important functions in the Microsoft C++ runtime implementation, including `_RTDynamicCast` and `_RTtypeid`. Implementing the `RtlPcToFileHeader` hook allows these functions to work properly. (Well, in the case of CxxThrowException, at least until RtlRaiseException is called, as Qiling does not have support for SEH right now)

JFarAur:
Update: I have since built more on this work, currently I have added some preliminary support for user-mode C++ runtime exceptions to work properly on 64-bit Windows.

I turned this PR into a draft for now. I will update it with more changes soon.

JFarAur:
There is an issue currently where C++ exceptions are still not quite working because of the RaiseException hook. If the RaiseException hook is removed then C++ exceptions are emulated correctly, however I am pretty sure this is incompatible with the previous code for setting unhandled exception filters, which I did not want to break.

Likewise, typeid and dynamic_cast are still not 100% there, but they're very close to working properly, I think these can be fixed with maybe another little hook or two.

I will try to figure these out later.

JFarAur:
After the latest changes, `typeid` and `dynamic_cast` are now functioning correctly.

Some of the tests were failing on x86. It turned out that this was because user32 DllMain was proceeding for longer than before since some APIs are now functioning properly, but now it was crashing in a different manner that made the tests fail. I added user32 to the DllMain blacklist because it was expected to fail in the tests before anyways, and allowing it to run at all was causing more issues than was worth to fix.

JFarAur:
To resolve the conflict with the existing unhandled exception filter code, a new hook for `ZwRaiseException` was added which forwards second-chance exceptions to the registered unhandled exception filter. The hook for `RaiseException` can then be safely removed, as the native functions involved with first-chance exception dispatching are now functioning mostly correctly.

The result of this work is now that C++ exceptions and try.. catch are functioning more or less correctly.

qiling/arch/x86_utils.py
```diff
@@ -60,6 +61,11 @@ def __init__(self, ql: Qiling, base=QL_X86_GDT_ADDR, limit=QL_X86_GDT_LIMIT, num
         # setup GDT by writing to GDTR
         ql.arch.regs.write(UC_X86_REG_GDTR, (0, base, limit, 0x0))
 
+        if ql.arch.type == QL_ARCH.X8664:
```
elicn: `QL_ARCH` is an enum, better use `is` instead of `==`.
Also, this could have been simply:
```python
self.is_long_mode = ql.arch.type is QL_ARCH.X8664
```
or
```python
self.is_long_mode = ql.arch.bits == 64
```

qiling/arch/x86_utils.py
```diff
@@ -93,7 +99,18 @@ def make_selector(idx: int, rpl: int) -> int:
         return (idx << 3) | QL_X86_SEGSEL_TI_GDT | rpl
 
     def register_gdt_segment(self, index: int, seg_base: int, seg_limit: int, access: int) -> int:
-        flags = QL_X86_F_OPSIZE_32
+        is_code = access & QL_X86_A_CODE
+
+        if is_code and self.is_long_mode:
```
elicn, line 101: It looks like `is_long_mode` is used only here. We could spare the self member and define it here ad-hoc just like `is_code` above.

qiling/loader/pe.py
```diff
@@ -29,6 +29,16 @@
     from logging import Logger
     from qiling import Qiling
 
+class ForwardedExport:
```
elicn: Looks like the instance content of this class never changes. Better make it a `NamedTuple`.

qiling/loader/pe.py
```diff
@@ -79,6 +89,16 @@ class Process:
     export_symbols: MutableMapping[int, Dict[str, Any]]
     libcache: Optional[QlPeCache]
 
+    # maps image base to RVA of its function table
+    function_table_lookup: MutableMapping[int, int]
```
elicn: Generalized annotations are good for function arguments, but return values and members should be annotated by the most specific way. In this case I guess this is `Dict[int, int]`.

qiling/loader/pe.py
```diff
@@ -79,6 +89,16 @@ class Process:
...
+
+    # maps image base to its list of function table entries
+    function_tables: MutableMapping[int, list]
+
+    # List of exports which have been forwarded from
+    # one DLL to another.
+    forwarded_exports: list[ForwardedExport]
```
elicn: `list` cannot be used for annotations on earlier Python 3 versions. Should be replaced with `List`.

qiling/loader/pe.py
```diff
@@ -105,6 +125,110 @@ def __get_path_elements(self, name: str) -> Tuple[str, str]:
...
+            if hasattr(pe, 'DIRECTORY_ENTRY_EXCEPTION'):
+                exception_dir = pe.OPTIONAL_HEADER.DATA_DIRECTORY[
+                    pefile.DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_EXCEPTION']
+                ]
+                
+                self.function_table_lookup[image_base] = exception_dir.VirtualAddress
+
+                runtime_function_list = []
+
+                for _, exception_entry in enumerate(pe.DIRECTORY_ENTRY_EXCEPTION, start=1):
```
elicn: Not sure why you use `enumerate` only to discard the index it provides. This loop could have been just:
```python
runtime_function_list = list(exception_entry)
```

qiling/loader/pe.py
```diff
@@ -105,6 +125,110 @@ def __get_path_elements(self, name: str) -> Tuple[str, str]:
...
+            control_pc: The program counter.
+
+        Returns:
+            A tuple (index, runtime_function)
+        """
+        function_table = self.function_tables[base_addr]
+
+        # Initiate a search of the function table for a RUNTIME_FUNCTION
+        # entry such that the provided PC falls within its start and end range.
+        for i, runtime_function in enumerate(function_table):
```
elicn: Consider turning this loop into:
```python
return next(((i, rtfunc) for i, rtfunc in enumerate(function_table) if control_pc - base_addr in range(rtfunc.struct.BeginAddress, rtfunc.struct.EndAddress)), (None, None))
```

qiling/loader/pe.py
```diff
@@ -281,8 +435,8 @@ def call_dll_entrypoint(self, dll: pefile.PE, dll_base: int, dll_len: int, dll_n
         # the blacklist may be revisited from time to time to see if any of the file
         # can be safely unlisted.
         blacklist = {
-            32 : ('gdi32.dll',),
-            64 : ('gdi32.dll',)
+            32 : ('gdi32.dll','user32.dll',),
```
elicn, line 434: Why blacklist user32.dll?

qiling/loader/pe.py
```diff
@@ -105,6 +125,110 @@ def __get_path_elements(self, name: str) -> Tuple[str, str]:
...
+                ]
+                
+                self.function_table_lookup[image_base] = exception_dir.VirtualAddress
+
+                runtime_function_list = []
+
+                for _, exception_entry in enumerate(pe.DIRECTORY_ENTRY_EXCEPTION, start=1):
+                    runtime_function_list.append(exception_entry)
+
+                if self.function_tables.get(image_base) is None:
```
elicn: Consider this instead:
```python
if image_base not in self.function_tables:
```

qiling/os/windows/dlls/kernel32/errhandlingapi.py
```diff
@@ -44,15 +44,6 @@ def hook_GetLastError(ql: Qiling, address: int, params):
 def hook_SetLastError(ql: Qiling, address: int, params):
     ql.os.last_error = params['dwErrCode']
 
-# LONG UnhandledExceptionFilter(
-#   _EXCEPTION_POINTERS *ExceptionInfo
-# );
-@winsdkapi(cc=STDCALL, params={
-    'ExceptionInfo' : POINTER
-})
-def hook_UnhandledExceptionFilter(ql: Qiling, address: int, params):
```
elicn, line 53: Though function definitions like this one may appear useless, they actually serve a purpose of emitting the function name along with its arguments when called. If the reason for removing this was to let it execute the real implementation (instead of doing nothing and returning 1), then just make it a pass-through stub by adding `passthru=True` to the `winsdkapi` decorator. This was get the best of two worlds: we get the real implementation to execute, while we get notified about it being called and the effective arguments used to call it.

qiling/os/windows/dlls/kernel32/winbase.py
```diff
@@ -159,24 +159,6 @@ def hook__lwrite(ql: Qiling, address: int, params):
 def hook_FatalExit(ql: Qiling, address: int, params):
     ql.emu_stop()
 
-# PVOID EncodePointer(
-#  _In_ PVOID Ptr
-# );
-@winsdkapi(cc=STDCALL, params={
-    'Ptr' : PVOID
-})
-def hook_EncodePointer(ql: Qiling, address: int, params):
```
elicn, line 168: What is the reason for removing those two functions? Were they incorrectly implemented / interfered with the real implementation?

qiling/os/windows/dlls/msvcrt.py
```diff
@@ -498,6 +469,42 @@ def hook_malloc(ql: Qiling, address: int, params):
 
     return ql.os.heap.alloc(size)
 
+def __realloc(ql: Qiling, address: int, params):
```
elicn: Isn't this pretty much the same as heap realloc?
Consider refactoring them both in a way they could share common code.

qiling/os/windows/dlls/ntdll.py
```diff
@@ -68,6 +70,27 @@ def _QueryInformationProcess(ql: Qiling, address: int, params):
 
         res_data = bytes(pci_obj)
 
+    
+    elif flag == ProcessCookie:
+        hCurrentProcess = {
```
elicn: If you need an unsigned representation of a `-1` value, you can have it with:
```python
hCurrentProcess = ql.pack(-1)
```
If this is about getting the `MAX_INT` for a given architecture size, then consider:
```python
hCurrentProcess = (1 << ql.arch.bits) - 1
```

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+    'PcValue'    : PVOID,
+    'BaseOfImage': PVOID
+})
+def hook_RtlPcToFileHeader(ql: Qiling, address: int, params):
+    pc = params["PcValue"]
+    base_of_image_ptr = params["BaseOfImage"]
+
+    containing_image = ql.loader.find_containing_image(pc)
+
+    if containing_image:
```
elicn, line 507: Consider the following instead:
```python
base_addr = containing_image.base if containing_image else 0
```

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+
+    # TODO: Make use of the history table to optimize this function.
+    # Alternatively, we could add caching to the loader, seeing as the
+    # loader is responsible for lookups in the function table.
+
+    # For simplicity, we are going to ignore the history table.
+    # history_table_ptr = params["HistoryTable"]
+
+    # This function should not be getting called on x86.
+    if ql.arch.type != QL_ARCH.X8664:
```
elicn: Judging by the error message it would make more sense to write this as:
```python
if ql.arch.type is QL_ARCH.X86:
```
If only x86-64 is allowed (bottom line is the same, but different way of looking at it) then leave it like this.

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+    # copy it there, and proceed.
+    if image_base_ptr != 0:
+        ql.mem.write_ptr(image_base_ptr, base_addr)
+
+    # Get the base address of the function table.
+    function_table_addr = base_addr + ql.loader.function_table_lookup[base_addr]
+
+    # Look up the RUNTIME_FUNCTION entry; we are interested in the index in the table
+    # so that we can compute the address.
+    runtime_function_idx, runtime_function = ql.loader.lookup_function_entry(base_addr, control_pc)
```
elicn, line 559: It looks like `lookup_function_entry` don't really need to return the function, but only its index. Maybe modify that function to return a `Optional[int]` where an integer means index and `None` means "not found"?

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+#     IN PVOID ControlPc,
+#     OUT PVOID *ImageBase,
+#     OUT PULONG SizeOfTable
+# );
+@winsdkapi(cc=STDCALL, params={
+    'ControlPc': PVOID,
+    'ImageBase': PVOID,
+    'SizeOfTable': PVOID
+})
+def hook_RtlLookupFunctionTable(ql: Qiling, address: int, params):
```
elicn, line 580: I have a deja-vu with this function implementation.. is it possible to avoid code duplication by refactoring them for better code sharing?

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+    # In Windows, an unhandled exception triggers the
+    # top-level unhandled exception filter, after which the process
+    # is terminated and error reporting services are called.
+    # Regardless of whether an unhandled exception filter is present,
+    # the process terminates with the same error code that was raised.
+
+    # Our strategy for this hook is to forward second-chance exceptions
+    # to the registered unhandled exception filter, if one exists.
+
+    if first_chance:
```
elicn: Any reason we are not supporting first-chance..?
Is it by-design, or we just lack the implementation for it?

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+    exception_pointers_ptr = ql.os.heap.alloc(epointers_struct.sizeof())
+
+    with epointers_struct.ref(ql.mem, exception_pointers_ptr) as epointers_obj:
+        epointers_obj.ExceptionRecord = exception_ptr
+        epointers_obj.ContextRecord = context_ptr
+
+    exception_filter = handle.obj
+    ql.log.debug(f'[ZwRaiseException] Resuming execution at the top-level exception filter at 0x{exception_filter:08X}.')
+
+    # Hack: We are going to fake that the caller of ZwRaiseException
```
elicn, line 691: Can you please explain why we need this hack?

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+
+    handle = ql.os.handle_manager.search("TopLevelExceptionHandler")
+
+    if handle is None:
+        ql.log.debug(f'[ZwRaiseException] No top-level exception filter was found.')
+        ql.log.info(f'The process exited with code 0x{exception_code:08X}.')
+
+        ql.os.exit_code = exception_code
+        
+        ql.emu_stop()
```
elicn, line 673: `emu_stop` will stop the emulation but the function still continues.
Do we need to return right away, or proceed to the rest of this function (like it does currently)?

qiling/os/windows/dlls/ntdll.py
```diff
@@ -452,4 +475,238 @@ def hook_wcsstr(ql: Qiling, address: int, params):
...
+
+        ql.emu_stop()
+
+    ql.hook_address(__post_exception_filter, ret_addr)
+
+    exception_filter_args = [(POINTER, exception_pointers_ptr)]
+
+    # Resume execution at the registered unhandled exception filter.
+    # If a program is using a custom unhandled exception filter as an anti-debugging
+    # trick, then the exception filter might not return.
```
elicn, line 715: Can we provide a way for users to intercept that so they can work around that anti-debugging trick?

elicn (changes requested):
This is a great contribution, thank you!
Really appreciating the time you took to comment and clarify what the code is doing and why.
Please see my comments and questions.

elicn:
Would you be able to create test cases for the exception handling?
(e.g. exception caught, exception caught through a chain, unhandled exception, etc.)

JFarAur:
> Would you be able to create test cases for the exception handling? (e.g. exception caught, exception caught through a chain, unhandled exception, etc.)

I've been working on some test cases, I will add them soon :)

Thanks for looking it over, I will incorporate your feedback and try to answer your questions.

elicn:
Just note `passthru` means the hook implementation is _ignored_ and it used only for its declaration.

qiling/loader/pe.py
JFarAur: I clarified a little in one of my comments on the PR, but to summarize, user32 DllMain was crashing even before my changes, but Qiling has some code that tries to eat exceptions in DllMain and continue execution anyway. So, examples and tests were still working properly.

However after my changes, user32 DllMain proceeds for longer (since more APIs are working now), but it crashes later in the process in a way that the other recovery code doesn't really handle. So adding it to the blacklist allows x86 to work again. Basically, Qiling was already counting on user32 DllMain to fail, the logic was just somewhere else.

Getting user32 DllMain to work properly would also be a lot of work which is outside the scope/goals of this PR, since it calls a bunch of functions to connect to CSR.

JFarAur:
> Just note `passthru` means the hook implementation is _ignored_ and it used only for its declaration.

That doesn't seem to be the case in my tests, it seems like Qiling just uses passthru as a flag to not do rewinding after the hook code finishes. For the `ZwRaiseException` hook, this allows me to set RIP to the exception filter without Qiling interfering, otherwise Qiling rewinds to the caller of `ZwRaiseException` and execution doesn't continue at the exception filter.

qiling/os/windows/dlls/kernel32/winbase.py
JFarAur: Well, the value these functions use to crypt pointers comes from `NtQueryInformationProcess` with the undocumented `ProcessCookie` information class. One thing I noticed is that in ntdll, the process cookie is sometimes used to crypt things in an inline fashion, *without* going through Encode/DecodePointer.

During testing I had some strange crashes which I believe were due to mismatches, maybe a pointer encoded somewhere with one implementation, then Windows trying to decode it somewhere else inline, bypassing the Qiling hook, also using an invalid process cookie. This was causing some issues, but adding the ProcessCookie case to `NtQueryInformationProcess` allows us to run the native functions without problems. Also, it ensures that stuff only ever gets crypted with the same logic.

qiling/os/windows/dlls/ntdll.py
JFarAur: Yeah, this function was mostly written before I had settled on an approach for exception dispatching, at the time I wasn't really sure that getting the native implementations of `RtlDispatchException`, `RtlVirtualUnwind` to work was viable. They returned function entries in case I had to reimplement stack unwinding from scratch in Qiling, but this turned out to be unnecessary, so maybe the prototypes can be rewritten.

qiling/os/windows/dlls/ntdll.py
JFarAur: Well, first-chance user-mode software exceptions are already pretty much supported for 64-bit. `ZwRaiseException` is actually not called anywhere in the process for such exceptions.

For C++ exceptions, the sequence of events is:

`throw` happens somewhere
-> `_CxxThrowException` is called
-> `kernelbase.RaiseException`
-> `ntdll.RtlRaiseException`
-> `ntdll.RtlDispatchException` is called directly, and locates and invokes handlers.
-> `ntdll.RtlUnwindEx` does stack unwinding, and restores context to where it should be after the handler routine is done.

In this case, `ZwRaiseException` is never called, and there is never any transition to kernel mode.

However, if the dispatcher fails to find a handler, then `ZwRaiseException` is called. At this point there is a short transition to kernel-mode, then the kernel dispatcher passes control to `KiUserExceptionDispatcher` and `RtlRaiseException` is called again. The exception data is the same as before, except that now we have a second-chance exception and the dispatcher will terminate the process after the unhandled exception filter is called.

I am not really sure in what situations `ZwRaiseException` would be called to raise first-chance exceptions, I think the most likely scenario would be a Windows driver that intentionally raises an exception in some error situation. To support this properly we would have to handle kernel-mode exception dispatching which is outside the scope of this PR, I wanted to focus mostly on user-mode here.

qiling/os/windows/dlls/ntdll.py
JFarAur: My last comment clarifies the process a bit, but normally `ZwRaiseException` is a wrapper for the "raise exception" syscall on Windows. It transitions to the kernel-mode dispatcher. If we have an uncaught software exception, then we can safely assume that the kernel-mode dispatcher would have passed control to the user-mode dispatcher, and eventually end up in the unhandled exception filter. However, since we have an uncaught exception, Windows is going to terminate the process regardless after calling the unhandled exception filter.

From what I can tell, Qiling's unhandled exception code before just invoked the unhandled exception filter on any call to RaiseException, so basically Qiling was sort of doing this same hack before, but for all exceptions, and it skipped even more of the process because it didn't pass first-chance exceptions to native code. Qiling's hook also returned after calling the exception filter. However, on Windows a second-chance exception will always result in an "abnormal termination" and the process terminates with an error code.

The hack mainly allows us two things:
* To skip the kernel-mode transition and pretend that we came from the dispatcher, and
* to retake control after the unhandled exception filter executes, at which point we can simulate the "abnormal" termination.

qiling/os/windows/dlls/ntdll.py
JFarAur: I'm not really sure what you mean by intercept. With this implementation the program should already behave more like it's running on real Windows in this scenario, a debugger would break on this exception and maybe even terminate the program.

Maybe it would make sense to count calls to SetUnhandledExceptionFilter, if there is more than one we could print a warning that the program is likely setting its own exception filter.

qiling/loader/pe.py
JFarAur: I'm skeptical of the `if ... in range(...)`, I think that's more inefficient than just checking the interval.

qiling/os/windows/dlls/ntdll.py
JFarAur: Good catch, that was not intentional, it should return here.

qiling/os/windows/dlls/ntdll.py
JFarAur: Actually, to follow up on this, it seems that what I wrote is not entirely correct. On 32-bit Windows, `ZwRaiseException` is used for exceptions raised in user mode as well. If the native `ZwRaiseException` is allowed to run, then it will definitely result in an error, because all it does is use a syscall. Personally, I thought it will be nicer for users if they see a "this isn't implemented" message, rather than some unicorn execution errors.

Right now, Qiling doesn't really have much infrastructure for Windows syscalls. And Windows syscalls are a lot more challenging than for example Linux, because the syscall numbers change every version of the OS, sometimes they even change between builds.

Also, 32-bit Windows has very different implementations of lots of exception-related stuff, not just the actual dispatching. For example, stack frames are managed on the stack rather than being in nice compiler-generated structures in the data section. So it would be a lot of work to get this working, and for the sake of the PR scope not getting too large, I mainly wanted to focus on 64-bit. In the future maybe I will work on the general case though.

qiling/os/windows/dlls/ntdll.py
JFarAur: I think exception data may be present on ARM too so it makes the most sense to write the way you did.

qiling/arch/x86_utils.py
JFarAur: ql is not saved in self, so it's not available in that method, I opted for just saving a boolean instead.

qiling/os/windows/dlls/kernel32/errhandlingapi.py
JFarAur: Right, this is one thing I was confused about when I first started using Qiling. Seeing the warnings about unimplemented functions, though this just meant that native versions were being called.

I feel like it may even make sense to rewrite the message, so it says something like `Calling native function` instead. And maybe a nice future project could be putting a bunch of function prototypes in a data file that gets loaded, so that we can see the arguments etc for all the known APIs. And maybe have flags for "known working" APIs, so those don't get shown with warning icons.

In general my approach has been: I just need to identify the absolute most vital APIs that other APIs depend on, then a bunch of unnecessary hooks can be removed and the emulation will have better stability. Especially after these changes, there are a lot of working native APIs. They're working because Qiling covers all their dependencies, and I feel like it doesn't make sense to show warning icons for all these APIs, but also it might be unmaintainable to write passthru hooks for every single one too.

JFarAur:
Set of tests has been added, they depend on some binaries which I added to the rootfs: https://github.com/qilingframework/rootfs/pull/34

Tests on x86:
- C++ hello world with `std::cout`
- Type id, virtual functions, dynamic_cast

Tests on x86_64:
- C++ hello world with `std::cout`
- Type id, virtual functions, dynamic_cast
- "Soft" SEH with RaiseException
- C++ exceptions with try..catch
- Unhandled exception (user filter present)
- Unhandled exception (user filter not present)

elicn:
> That doesn't seem to be the case in my tests, it seems like Qiling just uses passthru as a flag to not do rewinding after the hook code finishes.

Tagging a hook as a `passthru` means it is there only to show the call details, and there should not be an alternate implementation. Rather it continues to the real implementation (that is, doesn't patch the program counter so it keeps going to where it meant to go). Having a hook implementation tagged as `passthru` is kind of a hack - just FYI.

Maybe worth documenting that is your desired behavior, so it won't fail when the code changes in the future.

elicn:
> Set of tests has been added, they depend on some binaries which I added to the rootfs: [qilingframework/rootfs#34](https://github.com/qilingframework/rootfs/pull/34)

Thanks! We'll need to author test cases for them.
Please be sure to include also the source files and Makefiles in the appropriate directories.

JFarAur:
> > Set of tests has been added, they depend on some binaries which I added to the rootfs: [qilingframework/rootfs#34](https://github.com/qilingframework/rootfs/pull/34)
> 
> Thanks! We'll need to author test cases for them. Please be sure to include also the source files and Makefiles in the appropriate directories.

Test cases added here: 7dd9fcdc74aba6ac8e190ba71799d46881600cf1
Sources added here: 8f92c73b4076b5c6cc5da64f3e53a966bae37fcf

xwings:
Hi @sakura57 thanks for the contribution and welcome to Qiling.

This looks amazing. its a huge PR and i might need sometime to take a look. 

as of now i will go ahead to approve the PR in rootfs and enable the CI.

JFarAur:
> Hi @sakura57 thanks for the contribution and welcome to Qiling.
> 
> This looks amazing. its a huge PR and i might need sometime to take a look.
> 
> as of now i will go ahead to approve the PR in rootfs and enable the CI.

Thanks for the reply, I am looking forward to your feedback. :)

I think I can see why the tests are failing also. There is a problem related to native `HeapValidate` being called. And there is a slight problem with unhandled exceptions on 32-bit, therefore al khaser is failing. But I think it is fixable without too much work, I will have a look later.

qiling/loader/pe.py
```diff
@@ -29,15 +30,10 @@
...
-class ForwardedExport:
-    def __init__(self,
-                 source_dll: str, source_ordinal: str, source_symbol: str,
-                 target_dll: str, target_symbol: str):
-        self.source_dll = source_dll
-        self.source_ordinal = source_ordinal
-        self.source_symbol = source_symbol
-        self.target_dll = target_dll
-        self.target_symbol = target_symbol
+ForwardedExport = namedtuple('ForwardedExport', [
```
elicn: Sorry to be picking on this..
This is an obsolete method of defining a named tuple. Your class should extend `NamedTuple` (see how `QlPeCacheEntry` does that), and then you specify all members as class members. This is a more elegant way to declare the class, and also you don't lose the typing annotation.

Something around these lines:
```python
class ForwardedExport(NamedTuple):
    source_dll: str
    source_ordinal: str
    source_symbol: str
    target_dll: str
    target_symbol: str
```

qiling/os/windows/dlls/kernel32/errhandlingapi.py
elicn: Thanks for the explanation. I agree that we should revise the message, but it is important to keep the users informed about what is being called and how, regardless of whether it is a native or emulated functionality. Pass-through stubs are the way to announce calls to native implementation.

JFarAur:
After some changes yesterday, wannacry is working properly, and al-khaser is starting. I know why al-khaser is failing, it won't really be a big problem to fix. There is an issue with the clipboard test though.

The clipboard test binary was compiled in Debug mode, and uses debug versions of the C/C++ runtime DLLs. These include a lot of additional error checking. For example, there are debug versions of all the memory functions which allocate a little extra space at the beginning and end of all memory buffers, and fill them with special values. If any changes are detected, then this is taken to mean the heap is corrupted, and the CRT terminates with assertion failures. This is the main reason why the clipboard test is not working right now.

I think a solution for this could be to allow all the native CRT memory functions to run. These depend ultimately on the kernelbase Heap* functions. As long as those implementations are robust, then all the error checking logic *should* work properly in theory.

The test is still failing though after I tried these changes. I think there are some different possibilities, either the CRT is actually doing its job correctly, and this is actually revealing a buffer overrun occurring somewhere in Qiling hooks. Or, somewhere Qiling is passing allocated memory to the program, which the CRT debug libs are expecting to have the debug header/footer, but it doesn't.

Though, I do think it is slightly strange to include a debug build in the test suites, since they are not suitable for production, even for malware. Not everyone is guaranteed to have the -d versions of the CRT DLLs on their system, they are usually included with Visual Studio. This is another reason the test could ultimately be failing, debug binaries should be really run with the same exact builds of the debug libs they were linked against. @xwings was this intentional to include a debug binary in the tests?

elicn:
At the end of the day this is an emulation framework, so we strive to _emulate_ as much as we can rather than _simulate_. If you notice a working API implementation, we can safely remove the hook -- or better, make it a pass-through stub.

As for the debug versions, we need to support them too. There is no good reason for us to exclude them. We can try to work around problematic APIs, or admit we cannot guarantee a flawless emulation, but we need to support it.

elicn:
To ease your debugging work, be sure to use the `trace` module (example [here](https://github.com/qilingframework/qiling/blob/dev/examples/extensions/trace/trace.py))

I am attaching here a small script that helped me debug al-khaser back then, by controlling the trace output. File extension changed to allow upload.
[run-al-khaser.py.txt](https://github.com/user-attachments/files/19564176/run-al-khaser.py.txt)

JFarAur:
My feeling was correct, the CRT assertions were detecting a real buffer overrun. It turned out to be from the LCMapString* functions, this has been addressed in a3bccf5bc7d5816c60b6c628127320f2a7ec2560.

I have cleaned up much of the code for msvcrt, keeping in mind to make hooks passthru when native APIs can be used instead of removing them, and allowing the native CRT memory functions to run. This is important for the debug CRT and its additional heap error checking to work correctly.

With these changes the tests are passing again, on my end at least.

For al-khaser, I reverted some of the original hook for `RaiseException`, on x86 only the unhandled exception filter is called as before. So basically on x86 the situation with exceptions is the same, however on x86_64 the native code for exception dispatching is used and software exceptions are mostly working correctly with the new changes.

qiling/os/windows/dlls/kernel32/winnls.py
```diff
@@ -80,18 +80,19 @@ def hook_IsValidCodePage(ql: Qiling, address: int, params):
...
 
-    enc = "utf-16le" if wstring else "utf-8"
-    res = f'{lpSrcStr}\x00'
+    char_size = 2 if wstring else 1
+    byte_count = cchSrc * char_size
 
     if cchDest and lpDestStr:
-        # TODO maybe do some other check, for now is working
-        ql.mem.write(lpDestStr, res.encode(enc))
+        source_bytes = ql.mem.read(lpSrcStr, byte_count)
```
elicn, line 92: Was the issue a buffer overflow? (the string is longer than the buffer size)
Maybe this should be taken care of by the caller, and if it crashes - we need to emulate this and not prevent it?

If so, why not just trimming the string to `lpDestStr`? Also, the new implementation doesn't do that and make the same assumption of `lpSrcStr` <= `lpDestStr`.

Unless I am missing something, I think the former implementation (using `LPCWSTR` and `LPCCSTR` args) was cleaner - just needed a small patch to make sure source is not larger than destination.

JFarAur: > Unless I am missing something, I think the former implementation (using `LPCWSTR` and `LPCCSTR` args) was cleaner - just needed a small patch to make sure source is not larger than destination.

The problem was not that Qiling was permitting a buffer overrun to occur if the function was used incorrectly, the problem was that Qiling itself was overrunning the buffer even when the caller was using the function correctly.

The typical sequence of events before was:
1. Some Windows library calls LCMapStringW, passing a source buffer with cchSrc=0x100 and cchDest=0, essentially asking what size buffer it needs to hold the mapped string.
2. Qiling LCMapString appends a terminating null and tells the Windows library that it needs a buffer that can hold 0x101 characters.
3. The Windows library allocates a buffer to hold 0x101 characters. If the program is using the debug CRT, `_malloc_dbg` allocates a slightly larger area of memory, adding a header before the buffer and guard bytes past the end.
4. The Windows library calls LCMapStringW again, passing the same source buffer with 0x100 characters, and destination buffer with cchDest=0x101.
5. Qiling LCMapString appends a terminating null, then calls encode() which appends *another* terminating null. Qiling then writes 0x102 characters, the original 0x100 plus two terminating nulls, to lpDestStr, which was only supposed to hold 0x101. This overwrites two of the guard bytes past the end of the buffer.
6. The debug CRT detects that memory past the end of the buffer was written, concludes that the heap is corrupted, and terminates on an assertion failure.

It is perhaps misleading to think of LCMapString as mapping a null-terminated string, it is more useful to think of it as mapping an array of characters. Null characters are not considered as terminating the string in the native implementation, rather they are ignored, unless cchSrc is negative. So it is not really necessary to append additional null characters, and the data should be viewed as an array of characters rather than a string.

On real Windows it would have been like this:

1. LCMapStringW(lpSrcStr=source buffer, cchSrc=0x100, lpDestStr=NULL, cchDest=0x0) returns 0x100
2. Windows library allocates a buffer with size 0x100
3. LCMapStringW(lpSrcStr=source buffer, cchSrc=0x100, lpDestStr=new buffer, cchDest=0x100) returns 0x100

and any null characters are ignored while mapping. The function always iterates over the entire source buffer, and null is mapped to itself.

If the supplied buffer size is too small, then according to MSDN, the function will return and set a last error code. This could be added for better emulation realism, but Windows itself does not really ever use this function in a way where this could happen, so it might not be strictly necessary.

xwings:
@elicn are we good to merge ?

elicn:
@sakura57, is this PR ready for merge, or you are still working on it?

JFarAur:
> @sakura57, is this PR ready for merge, or you are still working on it?

@elicn Ready, I think it's in a good state right now.

it has been getting some "battle testing" the past few weeks as I'm using this branch in my particular use case, haven't noted any problems since the last fixes.

elicn (approved):

elicn:
@xwings we are good to go.

xwings (approved):

PR #1558, opened 2025-04-11
epoll support

libumem:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

### Comments

This PR is intended to provide initial support for the `epoll_create`, `epoll_create`, `epoll_ctl`, and `epoll_wait` syscalls using the `select` library; similar to how `poll` is done already. The binaries for testing have also been PR'd [here](https://github.com/qilingframework/rootfs/pull/35); but I have since changed the names slightly after checking out the `examples/src` folder.


xwings:
Hi,

Thanks for the PR and welcome to Qiling Framework.

Please take a look at the error. it seems lik here is some error

libumem:
@xwings 

Ok, I disabled the unit tests I had provided, as they won't yet work until my examples are accepted upstream in `rootfs`; let alone the issues I outlined in my PR description.

xwings:
let me accept that pr

xwings:
you can enable the test. i updated the CI test and the merged rootfs

xwings:
We still have tons of issue witl multithread. 

not sure how to fix it and it seems like we dont have a betteer way too.

@wtdcode @elicn  you guys might want to take a look at this PR

libumem:
> We still have tons of issue witl multithread.

FWIW, one problem I've pointed out is with a single test case -- and only under unit testing, not the underlying Qiling emulation of multithreaded code. The example binaries are single-threaded, as is the epoll implementation I have so far.

qiling/os/posix/const.py
```diff
@@ -1064,3 +1064,23 @@ class qnx_mmap_prot_flags(QlPrettyFlag):
 SHMDT       = 22
 SHMGET      = 23
 SHMCTL      = 24
+
+# epoll syscall
+EPOLL_OPS = {0x001: "EPOLL_CTL_ADD", 0x002: "EPOLL_CTL_DEL", 0x003: "EPOLL_CTL_MOD"}
```
elicn: Please keep dict items on separate lines.

qiling/os/posix/const.py
```diff
@@ -1064,3 +1064,23 @@ class qnx_mmap_prot_flags(QlPrettyFlag):
...
+
+EPOLLIN = 0x001
```
elicn: Please keep consistency: either use hardcoded values or bit shifting for all values, but not both. Also please add a comment with the URL from which those values were taken from.

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+
+class QlEpollObj:
+    def __init__(self, epoll_object):
+        self._epoll_object = epoll_object
+        self._fds = {}  # key: fd, value: eventmask
+        # keep track of which fds have what eventmasks,
+        # since this isn't directly supported in select.epoll
+
+    @property
+    def get_epoll_instance(self):
```
elicn: No need to specify "get_" in an attribute getter. Same goes for next methods specifying "get_" and "set_".

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+        return self._epoll_object
+
+    @property
+    def get_eventmask(self, fd):
+        return self._fds[fd]
+
+    @property
+    def get_fds(self):
+        if len(self._fds.keys()) == 0:
```
elicn: Why not just:
```python
return list(self._fds.keys())
```
The empty case seems redundant.

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+            return []
+        return list(self._fds.keys())
+
+    def set_eventmask(self, fd, newmask):
```
elicn: Please use type annotations where possible.

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+        self._fds.pop(fd)
+        self._epoll_object.unregister(fd)
+
+    def close(self):
+        self.get_epoll_instance.close()
+
+    def is_present(self, fd):
+        if fd not in self.get_fds:
+            return 0
+        return 1
```
elicn: Unnecessary code complication; use:
```python3
def is_present(self, fd: int) -> bool:
    return fd in self.get_fds
```

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+
+'''
```
elicn: This should be a comment, not a dangling string.

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+        return 1
+    new_epolls_list = []
+    flag = 0
+    for ent in list(epolls_list):
+        watched = ent.get_fds
+        for w in watched:
+            if isinstance(ql_fd_list[w], QlEpollObj):
+                flag = 1
+                new_epolls_list.append(ql_fd_list[w])
+        if flag:
```
elicn: `flag` is not needed here. We can test for `new_epolls_list` instead:
```python
if new_epolls_list:
```

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+            check_epoll_depth(ql_fd_list, new_epolls_list, depth + 1)
+    return 0
+
+'''
+Modify an existing epoll
+man 7 epoll for more details
+'''
+def ql_syscall_epoll_ctl(ql: Qiling, epfd: int, op: int, fd: int, event: POINTER):
+    # Basic sanity checks first
+    if event != 0:
```
elicn: The entire if-else statement could be replaced with:
```python
# events list is uint32_t. however event may be null, for example, when deleting a fd from interest list
ql_event = event and ql.mem.read_ptr(event, 4)
```

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+        ql_event = (
+            0  # event can be null, for example, when deleting a fd from interest list
+        )
+    ql_op = ""
+    epoll_obj = -1
+    try:
+        epoll_parent_obj = ql.os.fd[epfd]
+        epoll_obj = epoll_parent_obj.get_epoll_instance
+    except KeyError as k:
+        ql.log.debug("Unable to grab epoll object, something wrong with ql.os.fd!")
```
elicn: We should not warn about emulated software errors (i.e. when the emulated code calls a function with illegal parameters), only about errors or unexpected scenarios of Qiling itself (e.g. something was supposed to be found, but we can't find it).

If this is the former, we should remove the log message and stick with the returned error value.

qiling/os/posix/syscall/epoll.py
```diff
@@ -0,0 +1,251 @@
...
+    try:
+        epoll_parent_obj = ql.os.fd[epfd]
+        epoll_obj = epoll_parent_obj.get_epoll_instance
+        if not isinstance(epoll_parent_obj, QlEpollObj):
+            return EINVAL
+    except KeyError:
+        ql.log.debug(f"FD {epfd} doesn't appear to be a valid epoll file descriptor")
+        return EBADF
+    try:
+        ql_event = ql.unpack(ql.mem.read(epoll_events, ql.arch.pointersize))
```
elicn: Use `ql.mem.read_ptr` instead.

tests/test_elf.py
```diff
@@ -769,7 +770,48 @@ def test_elf_linux_x8664_path_traversion(self):
...
+        # TODO: https://github.com/qilingframework/rootfs/pull/35 must be merged
+        # Source for onestraw server: https://github.com/onestraw/epoll-example
+        # with a slight change to exit after the first request
+        def client():
+            time.sleep(3) # give time for the server to listen
+            s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            dest = ("127.0.0.1", 8000)
+            try:
+                s.connect(dest)
+            except Exception as e:
```
elicn: No need to catch that exception only to fail the test, it would happen anyway.

elicn (changes requested):
Hi there,
Thanks for your contribution. See my review comments inline.
Also please add comments to explain the tests, what is happening on each stage and what is the expected outcome. Comments are key to make the tests usable; otherwise nobody will know what is actually tested and how to react when they fail.

libumem:
I appreciate the feedback @elicn 

I'm running into a couple of issues with the tests I have written, both in `test_elf.py`

1: When trying to feed my `epoll_0` example input via stdin, similar to the examples in the qiling docs; the input doesn't seem to make it through. If I am not doing anything to modify stdin; I can run the binary under Qiling normally, as it will echo my input or exit on receiving "stop", which is what it's supposed to do. I've disabled the test case for now.

2: When running the test `ELFTest.test_elf_linux_x8664_epoll_server` by itself, it works fine. But when I run the whole test suite,  with `PYTHONPATH=</path/to/this/pr/> python3 test_elf.py`; I get a "PermissionError: [Errno 1] Operation not permitted" error. I haven't been able to isolate the cause of this, as my server doesn't do anything that ought to require any sort special privileges outside of opening port 8000. If the port was an issue, I would expect the test to also fail when run standalone.

Would you please take a look at these problems?

elicn:
I've patched some of the code earlier today and posted a PR to your repo. Please review the changes I made to see if they make sense. There were a few snippets in which I could not figure out what should be the expected behavior; look for my inline comments starting with "elicn".

elicn:
Note I've posted another PR to your branch.
Please review the changes there to see if they align with your intent.

zkw13:
looks great. how's everything going now?

qiling/os/posix/syscall/epoll.py
```diff
@@ -154,15 +154,14 @@ def ql_syscall_epoll_ctl(ql: Qiling, epfd: int, op: int, fd: int, event: int):
             return -EINVAL
 
         event_ptr = ql.mem.read_ptr(event)
-        events = ql.mem.read_ptr(event_ptr, 4)
```
elicn: I am not sure this is correct. The 4th argument of `epoll_ctl` is `struct epoll_event *_Nullable event`, which is a pointer to a structure. The functions called later on need the `events` data, which is the first 32-bit field in that structure (i.e. `*epoll_event` and not `epoll_event`, which is the structure address)

qiling/os/posix/syscall/epoll.py
```diff
@@ -226,7 +224,7 @@ def ql_syscall_epoll_wait(ql: Qiling, epfd: int, epoll_events: int, maxevents: i
 
         # FIXME: the data packed after events should be the one passed on epoll_ctl
         # for that specific fd. currently this does not align with the spec
-        data = ql.pack32(events) + ql.pack64(fd)
```
elicn: According to [Linux kernel code](https://elixir.bootlin.com/linux/v6.14.4/source/include/uapi/linux/eventpoll.h#L83) `events` is a form of an `int`, and `data` is of type `__u64`, so data is always 64-bit in size. The code uses a macro that conditionally expands to alignment based on whether this is a 32 or 64 bit arch. In any case the size of the structure is 12 bytes

elicn (changes requested):
I think these fixes are wrong. Please see my comments.

qiling/os/posix/syscall/epoll.py
libumem: I first removed this in an attempt to fix invalid memory access issues this was causing. Working on it.

libumem: Should be taken care of now.

elicn:
You should roll back your last 2 commits to align back with the code I pushed.

libumem:
@elicn Thanks for your patience, I've had some stuff come up IRL that's delayed my response. 

> The functions called later on need the events data, which is the first 32-bit field in that structure (i.e. *epoll_event and not epoll_event, which is the structure address)

That's what I'm doing now in `ql_syscall_epoll_ctl`
```
    events = 0
    if event:
        events = ql.mem.read_ptr(event,4)
```

which should match what I think your intent is. This also matches the expected behavior in the test cases I have; although those aren't passing yet for unrelated reasons.

libumem:
Looking into the permission denied errors I'm getting when running the whole test suite here

elicn:
The commits you pushed after my PR are introducing wrong 'fixes'.
You modified the pointer handling with a wrong implementation, and replaced a useful comment with a useless one. There is no need to fix what not needs fixing.

libumem:
Rolled back to `b724f4cf63de7a2fd88239b12c3eaca9a8122c80`
which fires a `UC_ERR_READ_UNMAPPED` error in the `test_elf_linux_x8664_epoll_server` test case. 

Ran in PDB, which shows what I think is the issue -- this is what I was trying to deal with in the next commit
```
-> event_ptr = ql.mem.read_ptr(event)
(Pdb) list
151                 return -EEXIST
152  
153             if not event:
154                 return -EINVAL
155  
156 B->         event_ptr = ql.mem.read_ptr(event)
157             events = ql.mem.read_ptr(event_ptr, 4)
158  
159             # EPOLLEXCLUSIVE was specified in event and fd refers to an epoll instance
160             if isinstance(fd_obj, QlEpollObj) and (op & EPOLLEXCLUSIVE):
161                 return -EINVAL

(Pdb) hex(event)
'0x7ffffffffab8'
(Pdb) n
> <removed>/epoll.py(157)ql_syscall_epoll_ctl()
-> events = ql.mem.read_ptr(event_ptr, 4)
(Pdb) hex(event_ptr)
'0x380000005' -> # invoking events = ql.mem.read_ptr(event_ptr, 4) will cause an access violation
```

elicn:
Got it.
Undo these two commits so I can rebase my new work on top of yours:
`e4242ca741116c66fdde120e003b608559ef0b92`
`f9497bc7d6998c02bef5dd4cddf18e9132b24c7a`

libumem:
@elicn Done

elicn:
I've created a new PR.
Please let me know if you have any questions.

elicn:
The test errors suggest that you relocate the `x8664_onestraw_server` to the `bin` directory under the rootfs you are using.

libumem:
Commented out the test case, will dig deeper with what's going on in unittest.

libumem:
@elicn Up to you if you want to merge as is, or I can convert to draft PR.

The blocker was/is that the code works outside of unittest, but gets "Operation not permitted" errors when running the ELFTest group.

libumem:
@elicn Thoughts?

elicn:
I prefer avoiding incorporating changes that cannot be tested.
Is there anything we can do to have the test work?

libumem:
> Is there anything we can do to have the test work?

@elicn 
Unfortunately, no AFAICT. I have been stumped so far in trying to determine why the tests I wrote fail when run inside of unittest, but work when invoked separately. If you want to hold this PR until I or someone else can find+fix the root cause, that's fine. Or I could log a message that fires on using these syscalls which warns no tests are available and it should be treated as experimental until this is handled.

All up to you.

libumem:
@elicn Haven't looked at this in a while, but I finally figured out what was the issue. 
Now have one unit test.

qiling/os/posix/syscall/epoll.py
```diff
@@ -118,7 +118,7 @@ def ql_syscall_epoll_ctl(ql: Qiling, epfd: int, op: int, fd: int, event: int):
     if op not in (EPOLL_CTL_ADD, EPOLL_CTL_DEL, EPOLL_CTL_MOD):
         return -EINVAL
 
-    if epfd == fd:
+    if epfd == fd or fd == 0xffffffff: # latter condition was seen in testing, but should not happen in the real world
```
elicn: If it happens on testing, then this is a case we should take into account.
Please provide better context of what happens here and how it should be handled. Usually there is a reason `fd` would be set to `-1`, can you tell that is it here?

tests/test_elf.py
```diff
@@ -217,13 +217,13 @@ def test_elf_linux_x8664_static(self):
         ql = Qiling(["../examples/rootfs/x8664_linux/bin/x8664_hello_static"], "../examples/rootfs/x8664_linux", verbose=QL_VERBOSE.DEBUG)
         ql.run()
         del ql
-
+    #@unittest.skip('Experiment to avoid FD issue')
```
elicn: Guess this should be removed and replaced with a blank line.

tests/test_elf.py
```diff
@@ -217,13 +217,13 @@ def test_elf_linux_x8664_static(self):
...
     def test_elf_linux_x86(self):
         filename = 'test.qlog'
 
         ql = Qiling(["../examples/rootfs/x86_linux/bin/x86_hello"], "../examples/rootfs/x86_linux", verbose=QL_VERBOSE.DEBUG, log_devices=[filename])
         ql.run()
-
+        ql._log_file_fd.handlers[0].close() # prevent FD leak that causes downstream issues
```
elicn, line 226: This is very unusual.
Care to explain what this is about?

tests/test_elf.py
```diff
@@ -789,8 +789,7 @@ def test_elf_linux_x8664_epoll_simple(self):
         ql.run()
 
         self.assertIn(b'echo\n', ql.os.stdout.read())
-        del ql
-    @unittest.skip("See comment in https://github.com/qilingframework/qiling/pull/1558")
+        del ql
```
elicn, line 793: Please bring back the blank lines, here and down below where they got [accidently?] removed.

elicn (changes requested):
Good to hear. Providing a brief explanation could help us look into issues around this in the future.
Please refer to my comments in this review.

tests/test_elf.py
libumem: Done

tests/test_elf.py
libumem: Done

qiling/os/posix/syscall/epoll.py
libumem: Removed, this isn't necessary anymore. I wrote this in the process of creating the fix elsewhere

tests/test_elf.py
libumem: When running this test case, there was a dangling FD pointing to `ql.log` on the disk after `os.remove` got invoked. This FD ended up being added to the epoll args, and not the legitimate server socket as I observed outside the test suite.

So here, I just force a closure to keep the FD list in a consistent state.

libumem:
@elicn Are you ok with this?

elicn:
There are some still questionable parts in the code, but overall it looks fine.

Please create a different PR for `rootfs` repo to include your test binaries there. That PR must be merged before this one can, or it will try to test non-existing binaries.

libumem:
@elicn 

> Please create a different PR for rootfs repo to include your test binaries there

That was already merged [here](https://github.com/qilingframework/rootfs/pull/35)

elicn:
@xwings, this one is ready for an additional review + merge.

libumem:
@xwings Just bumping this

xwings (approved):

xwings:
Sorry for the delay. merged.

PR #1562, opened 2025-04-17
disable getauxval HWCAP value for ARM64, which indicates Atomics supp…

rliebig:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This change fixes the ARM64 elf loader to not indicate that atomics instructions are supported. The current QEMU version by Unicorn does not support these atomics call, therefore we should also not indicate to processes that this is a good idea. 

getauxval values are determined by the [kernel documentation](https://www.kernel.org/doc/html/v5.3/arm64/elf_hwcaps.html) like this. Therefore, we need to ensure that `HWCAP_ATOMICS` is not set.

Mostly the processes do check for support judging from these values like this (example from LIBC). The variable `_aarch64_have_lse_atomics` is controlled by getauxval.

```c
char __fastcall _aarch64_cas1_acq_rel(int a1, unsigned __int8 a2, atomic_uchar *a3)
{
  int v3; // w16

  if ( _aarch64_have_lse_atomics )
  {
    atomic_compare_exchange_strong(a3, (unsigned __int8 *)&a1, a2);
  }
  else
  {
    v3 = (unsigned __int8)a1;
    do
      a1 = __ldaxr((unsigned __int8 *)a3);
    while ( a1 == v3 && __stlxr(a2, (unsigned __int8 *)a3) );
  }
  return a1;
}
```

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -330,7 +330,7 @@ class QlLoaderELF(QlLoader):
         hwcap_values = {
             (QL_ARCH.ARM,   QL_ENDIAN.EL, 32): 0x001fb8d7,
             (QL_ARCH.ARM,   QL_ENDIAN.EB, 32): 0xd7b81f00,
-            (QL_ARCH.ARM64, QL_ENDIAN.EL, 64): 0x078bfbfd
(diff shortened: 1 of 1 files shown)


wtdcode (approved):
LGTM

elicn (approved):

PR #1577, opened 2025-07-06
QDB and BLOB fixes

### LINKED ISSUE #1568: Qdb error when using QlLoaderBlob
**Describe the bug**
When trying to use Qdb with a baremetal binary emulation that uses QlLoaderBlob, an AttributeError raises.

**Sample Code**
I am working on the dev branch.
My setup is the following one :
I emulate a bare metal binary running on ARM cortex A7 processor. For the global emulation setup I followed this qiling/examples/hello_arm_uboot.py and defined a cortex_a.ql file to setup the memory as required by the blob loader.
```python
with open("./baremetal_binary", "rb") as f:
        binary = f.read()

ql = Qiling(code=binary[0x00000000:],archtype=QL_ARCH.ARM, ostype=QL_OS.BLOB , verbose=QL_VERBOSE.DISABLED, cputype=ARM_CPU_MODEL.ARM_CORTEX_A7, profile="cortex_a.ql")

[...]

ql.debugger= "qdb"
ql.run()
```

**Expected behavior**
No error.

**Screenshots**
```
  File "/test_qilin/qiling/examples/mcu/fuzzing_test/cortex_A7/code_coverage/dev/general_script.py", line 60, in <module>
    ql.run()
  File "/test_qilin/qilingenv/lib/python3.12/site-packages/qiling/core.py", line 585, in run
    debugger = debugger(self)
               ^^^^^^^^^^^^^^
  File "/test_qilin/qilingenv/lib/python3.12/site-packages/qiling/debugger/qdb/qdb.py", line 73, in __init__
    self.dbg_hook(list(filter(lambda d: int(d, 0) != self.ql.loader.entry_point, init_hook)))
  File "/test_qilin/qilingenv/lib/python3.12/site-packages/qiling/debugger/qdb/qdb.py", line 105, in dbg_hook
    elif self.ql.loader.entry_point:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'QlLoaderBLOB' object has no attribute 'entry_point'
```

**Additional context**
The error is normal, indeed the QlLoaderBlob object doesn't have this attribute.
I found these lines in the qdb.py file (line 140) which seems to be here to handle the case but they are after different other access to self.ql.loader.entry_point atrribute previous them :
```python 
  if self.ql.os.type is QL_OS.BLOB:
            self.ql.loader.entry_point = self.ql.loader.load_address

   elif init_hook:
            for each_hook in init_hook:
                self.do_breakpoint(each_hook)
```
Maybe a fix could be to change the place of these lines.

elicn:
Highlights:
 - BLOB entry point and load address decoupled. Now it is possible to define a load address different than the entry point
 - BLOB example and test explained
 - Fixed qdb crash on disassembling instructions on allocated memory boundaries
 - Fixed gdb registers reference for ARM Cortex-M

Fixes #1568, #1580

### DIFF
--- a/qiling/debugger/gdb/xmlregs.py
+++ b/qiling/debugger/gdb/xmlregs.py
@@ -13,14 +13,21 @@ from qiling.arch.arm_const import (
     reg_map_q as arm_regs_q,
     reg_map_s as arm_regs_s
 )
+
+from qiling.arch.cortex_m_const import (
+    reg_map as conretx_m_regs
+)
+
 from qiling.arch.arm64_const import (
     reg_map as arm64_regs,
     reg_map_v as arm64_regs_v,
     reg_map_fp as arm64_reg_map_fp
 )
+
 from qiling.arch.mips_const import (
     reg_map as mips_regs_gpr
 )
+
 from qiling.arch.x86_const import (
     reg_map_32 as x86_regs_32,
     reg_map_64 as x86_regs_64,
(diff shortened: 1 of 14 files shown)


qiling/debugger/gdb/xmlregs.py
antcpl, line 1: I assume ```conretx_m_regs``` was supposed to be ```cortex_m_regs```

antcpl, line 1: Don't want to be insistent but I'm afraid there still is a typo : ```coretx_m_regs``` instead of ```cortex_m_regs```

elicn: omg

PR #1583, opened 2025-08-12
Raw binary blob support

technikelly:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

## Overview

This PR introduces a small code change that allows users to specify BLOB_RAW in the profile, instructing the Qiling loader not to map any memory segments. This functionality is particularly useful for emulating small code snippets or bare-metal code without an operating system, offering greater customization and flexibility. Users can now define all memory mappings directly within their Qiling scripts, tailoring the environment precisely to their use case. 

With this change, it becomes easier to port existing Unicorn scripts to Qiling and take advantage of additional features Qiling provides, such as debugging support and code coverage generation. Users can simply replace Unicorn API calls with equivalent Qiling API calls to benefit from these enhancements.

## Notes

The added unit tests are currently disabled because the test binary has not been merged to rootfs. The added example code will also not work until the test binary has been merged into rootfs. Corresponding rootfs PR - https://github.com/qilingframework/rootfs/pull/36

### DIFF
--- a/examples/hello_arm_blob_raw.py
+++ b/examples/hello_arm_blob_raw.py
@@ -1,0 +1,81 @@
+##############################################################################
+# Added example for raw binary blob  
+#    Kelly Patterson - Cisco Talos
+#         Copyright (C) 2025 Cisco Systems Inc
+##############################################################################
+from qiling import Qiling
+from qiling.const import QL_ARCH, QL_OS, QL_VERBOSE
+from qiling.extensions.coverage import utils as cov_utils
+
+BASE_ADDRESS = 0x10000000
+CHECKSUM_FUNC_ADDR = BASE_ADDRESS + 0x8
+END_ADDRESS = 0x100000ba
+DATA_ADDR = 0xa0000000 # Arbitrary address for data
+STACK_ADDR = 0xb0000000 # Arbitrary address for stack
+
+# Python implementation of the checksum function being emulated
+def checksum_function(input_data_buffer: bytes):
+    expected_checksum_python = 0
+    input_data_len = len(input_data_buffer)
+    if input_data_len >= 1 and input_data_buffer[0] == 0xDE: # MAGIC_VALUE_1
+        for i in range(min(input_data_len, 4)):
+            expected_checksum_python += input_data_buffer[i]
+        expected_checksum_python += 0x10
+    elif input_data_len >= 2 and input_data_buffer[1] == 0xAD: # MAGIC_VALUE_2
+        for i in range(input_data_len):
+            expected_checksum_python ^= input_data_buffer[i]
+        expected_checksum_python += 0x20
+    else:
+        for i in range(input_data_len):
+            expected_checksum_python += input_data_buffer[i]
+    expected_checksum_python &= 0xFF # Ensure it's a single byte
+
+def unmapped_handler(ql, type, addr, size, value):
+
+    print(f"Unmapped Memory R/W, trying to access {hex(size)} bytes at {hex(addr)} from {hex(ql.arch.regs.pc)}")
+
+def emulate_checksum_function(input_data_buffer: bytes):
+    print(f"\n--- Testing with input: {input_data_buffer.hex()} ---")
+
+    ql = Qiling(archtype=QL_ARCH.ARM, ostype=QL_OS.BLOB, profile="blob_raw.ql", verbose=QL_VERBOSE.DEBUG, thumb=True)
+
+    input_data_len = len(input_data_buffer)
+
+    # Map memory for the binary, data and stack
+    ql.mem.map(BASE_ADDRESS, 0x10000)
+    ql.mem.map(STACK_ADDR, 0x2000)
+    ql.mem.map(DATA_ADDR, ql.mem.align_up(input_data_len + 0x100)) # Map enough space for data
+
+    # Write the binary into memory
+    ql.mem.write(BASE_ADDRESS, open("rootfs/blob/example_raw.bin", "rb").read())
+
+    # Write input data
+    ql.mem.write(DATA_ADDR, input_data_buffer)
+
+    # Set up the stack pointer
+    ql.arch.regs.sp = STACK_ADDR + 0x2000 - 4
+    # Set up argument registers
+    ql.arch.regs.r0 = DATA_ADDR
+    ql.arch.regs.r1 = input_data_len
+
+    # Set the program counter to the function's entry point
+    ql.arch.regs.pc = CHECKSUM_FUNC_ADDR
+
+    # Set the return address (LR) to a dummy address.
+    ql.arch.regs.lr = 0xbebebebe
+
+    ql.hook_mem_unmapped(unmapped_handler)
+    #ql.debugger="gdb:127.0.0.1:9999"
+
+    # Start emulation
+    print(f"Starting emulation at PC: {hex(ql.arch.regs.pc)}")
+    try:
+        ql.run(begin=CHECKSUM_FUNC_ADDR, end=END_ADDRESS)
+    except Exception as e:
+        print(f"Emulation error: {e}")
+
+    print(f"Emulated checksum: {hex(ql.arch.regs.r0)}")
+
+if __name__ == "__main__":
+    data = b"\x01\x02\x03\x04\x05"  # Example input data
+    emulate_checksum_function(data)
--- a/examples/hello_arm_blob_raw.py
+++ b/examples/hello_arm_blob_raw.py
@@ -6,6 +6,7 @@
 from qiling import Qiling
 from qiling.const import QL_ARCH, QL_OS, QL_VERBOSE
 from qiling.extensions.coverage import utils as cov_utils
+from qiling.loader.loader import Image
 
 BASE_ADDRESS = 0x10000000
 CHECKSUM_FUNC_ADDR = BASE_ADDRESS + 0x8
@@ -29,6 +30,7 @@ def checksum_function(input_data_buffer: bytes):
         for i in range(input_data_len):
             expected_checksum_python += input_data_buffer[i]
     expected_checksum_python &= 0xFF # Ensure it's a single byte
+    return expected_checksum_python
 
 def unmapped_handler(ql, type, addr, size, value):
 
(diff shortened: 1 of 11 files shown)


elicn:
Hi @technikelly, welcome and thank you for your contribution.

Though the motivation behind this PR does resonate with me, I believe it introduces a special use case that could be covered by the existing Blob loader with minimal changes. While providing a convenient method for their users, special use cases tend to serve a very focused or limited purpose and introduce more complexity to the framework as it needs to support these cases on various components.

This is what I've picked up in my review:
 - The new approach uses `binary_name` key in profile, which is inconsistent with Qiling usage model. Blob is assumed to be provided directly as `bytes`, that is actually more "Unicorn-like"
 - Profile has a `binary_size` key which appears to be redundant, as file size is inferred when it is being read or simply by the length of code `bytes`
 - The existing Blob loader could be modified as follows to provide an equivalent functionality:
   - Treat the `heap_size` key as optional and configure the heap only if it is available. This way only the bare minimum memory is allocated, avoiding other Qiling "noise"
   - Everything else remains intact, while the user script can monkeypatch details as needed (e.g.: patch module name in loader, unmap unnecessary memory areas, etc.)

Also, please note that we cannot accept proprietary code with copyrights.
Please consider this feedback and let me know if you have any concerns.

[BTW - you may want to consider [udbserver](https://github.com/bet4it/udbserver) as a simpler method to debug Unicorn scripts]

technikelly:
Hi @elicn, thank you for getting back to me so quickly. I appreciate your thorough explanation of your reasoning and I fully agree that this could be achieved with a much smaller code change to qiling and monkeypatching the user script. 

I'll go ahead and make the change you suggested - of `heap_size` being configurable within the blob loader.

Regarding the copyrights, the code is not proprietary. Cisco requires a copyright notice on open source contributions, but the code ultimately falls under the license of the overall project which in this case is GPLv2. With that additional information is it still an issue to have copyright notices included with contributions to qiling? 

Thanks for the tip about udbserver - it definitely is simpler, but having code coverage in addition to debug is really nice.

xwings:
> Hi @elicn, thank you for getting back to me so quickly. I appreciate your thorough explanation of your reasoning and I fully agree that this could be achieved with a much smaller code change to qiling and monkeypatching the user script. 
> 
> 
> 
> I'll go ahead and make the change you suggested - of `heap_size` being configurable within the blob loader.
> 
> 
> 
> Regarding the copyrights, the code is not proprietary. Cisco requires a copyright notice on open source contributions, but the code ultimately falls under the license of the overall project which in this case is GPLv2. With that additional information is it still an issue to have copyright notices included with contributions to qiling? 
> 
> 
> 
> Thanks for the tip about udbserver - it definitely is simpler, but having code coverage in addition to debug is really nice.

Hi,

Thank you for PR and also welcome to Qiling Framework.

We do have the "tradition" to include author (PR) during the early days. Along the way we found out the list will get longer and longer in the specific file.

Later, we decided update changelog and put the names in sparate for easier maintainance. I hope you understand where we come from.

examples/hello_arm_blob_raw.py
```diff
@@ -0,0 +1,92 @@
+##############################################################################
```
elicn, line 1: Examples are an essential resource for beginners to learn how a certain function of Qiling works. For an example to be effective, it has to describe clearly what is does and how it does it, otherwise the example does not add any value to the reader.

This example file seems to be the very specific case you have been working on, so you'll need to ask yourself what is in there for other Qiling users. If you think it is valuable, kindly revise the example file and add comments and documentation.

examples/hello_arm_blob_raw.py
```diff
@@ -0,0 +1,92 @@
...
+        for i in range(input_data_len):
+            expected_checksum_python ^= input_data_buffer[i]
+        expected_checksum_python += 0x20
+    else:
+        for i in range(input_data_len):
+            expected_checksum_python += input_data_buffer[i]
+    expected_checksum_python &= 0xFF # Ensure it's a single byte
+    return expected_checksum_python
+
+def unmapped_handler(ql, type, addr, size, value):
```
elicn: Please use annotations where applicable.
Annotations have to be compliant to Python 3.8

examples/hello_arm_blob_raw.py
```diff
@@ -0,0 +1,92 @@
...
+
+    print(f"Unmapped Memory R/W, trying to access {hex(size)} bytes at {hex(addr)} from {hex(ql.arch.regs.pc)}")
```
elicn: While still possible, `hex` is considered obsolete. Could replace this with:
```python
print(f"Unmapped Memory R/W, trying to access {size:d} bytes at {addr:#010x} from {ql.arch.regs.pc:#010x}")
```
FYI, ARM has an additional `effective_pc` method that may be used to retrieve the 'real' PC value, taking thumb mode (odd address) into consideration - just in case you need it.

examples/hello_arm_blob_raw.py
```diff
@@ -0,0 +1,92 @@
...
+    with open("rootfs/blob/example_raw.bin", "rb") as f:
+        raw_code = f.read()
+
+    ql = Qiling(code=raw_code, archtype=QL_ARCH.ARM, ostype=QL_OS.BLOB, profile="blob_raw.ql", verbose=QL_VERBOSE.DEBUG, thumb=True)
+
+    # monkeypatch - Correcting the loader image name, used for coverage collection
+    # Remove all images with name 'blob_code' that were created by the blob loader
+    ql.loader.images = [img for img in ql.loader.images if img.path != 'blob_code']
+    # Add image back with correct info
+    ql.loader.images.append(Image(ql.loader.load_address, ql.loader.load_address + ql.os.code_ram_size, 'example_raw.bin'))
```
elicn: Better avoid hardcoding filenames

qiling/loader/blob.py
```diff
@@ -31,8 +36,12 @@ def run(self):
         # FIXME: heap starts above end of ram??
         # FIXME: heap should be allocated by OS, not loader
         heap_base = code_ends
-        heap_size = int(self.ql.os.profile.get("CODE", "heap_size"), 16)
-        self.ql.os.heap = QlMemoryHeap(self.ql, heap_base, heap_base + heap_size)
+        # if heap_size is defined, create the heap
+        try:
```
elicn: Better use this opportunity to move this to Blob OS, as the `FIXME` comment above suggests.
I would also suggest reading `heap_base` along with `heap_size` to keep consistency with other use cases.

Also note that we have a better method to extract integer values from config:
```python
self.ql.os.profile.getint("CODE", "heap_size")
```
It already takes care of whatever representation it is there (not assuming hex specifically).

Putting all of this together, we can remove the two `FIXME` comments above.

tests/test_blob.py
```diff
@@ -82,6 +86,101 @@ def partial_run_init(ql: Qiling):
...
+                    expected_checksum ^= input_data_buffer[i]
+                expected_checksum += 0x20
+            else:
+                for i in range(input_data_len):
+                    expected_checksum += input_data_buffer[i]
+
+            return expected_checksum & 0xFF
+
+        # Test cases with descriptions
+        test_cases = {
```
elicn: Though I understand where this is coming from [de-duplicating test code], it is a `unittest` bad practice to bundle multiple test cases in one method: if one of these tests fails, everything else is also considered as a failure.

Please create dedicated `test_*` methods for each test case (not sure that we need all, though, after all we test Qiling functionality, not the binary) and extract the ql prep code to a method all test cases can call.

elicn (changes requested):
@technikelly, thanks for the fixes.
Please see my review comments inline.

I believe we would need to get rid of those repeating copyright notices, as @xwings suggested we can include your name and your employer's name in the contributors list.

technikelly:
Thank you @elicn and @xwings for explaining about the issue with copyright notices. I was able to get an exception, so I've removed them. I've also implemented the changes requested by @elicn (thank you for the review!). As it stands now, the changes in this PR consist of:

* moving the heap creation to the blob OS module, and making that heap creation optional based on the existence profile values `heap_address` and `heap_size`
* Test code 
* Example showcasing the modifications needed to enable code coverage

As stated in the comments, let me know if you would prefer not to include an example with this PR. 

I also wanted to point out that I modified the two instances of uboot_bin.ql so that they contain `heap_address` so that the code change wouldn't break the uboot example or test

tests/test_blob.py
technikelly: Sure thing, thanks for pointing that out

qiling/loader/blob.py
technikelly: oh good to know, I do see that used in other loaders now that you mention it. But good point, I did address those FIXME's in https://github.com/qilingframework/qiling/pull/1583/commits/5ec4a3b57a0cdf6e638026e04ab94cc60b1aabb5 as a result I modified uboot_bin.ql to contain a value for `heap_address`

examples/hello_arm_blob_raw.py
technikelly: addressed in - https://github.com/qilingframework/qiling/pull/1583/commits/5ec4a3b57a0cdf6e638026e04ab94cc60b1aabb5

examples/hello_arm_blob_raw.py
technikelly: good to know, thanks!

examples/hello_arm_blob_raw.py
technikelly: apologies, addressed in https://github.com/qilingframework/qiling/pull/1583/commits/5ec4a3b57a0cdf6e638026e04ab94cc60b1aabb5

examples/hello_arm_blob_raw.py
technikelly: Good point, thank you for that feedback. This example code is a generic fabricated example with the goal of being simple and  having multiple code paths that could be explored and therefore highlighted with code coverage. I believe a beginner would not know that they could emulate a small code snippet in this way. Although, if you feel that the example isn't beneficial, it is ultimately your decision. Let me know if you would prefer me to remove it altogether.

elicn:
Everything looks good to me. Thanks @technikelly.
We'll merge the changes to rootfs first to allow the tests here to pass.

elicn (approved):

technikelly:
Great, thank you @elicn! I just enabled the test and updated rootfs.

elicn:
@xwings, pending your review.

xwings (approved):
Looks good to me.

xwings:
Merged and thanks! Again, welcome to Qiling Framework

PR #1600, opened 2025-11-04
PPC debugger XML support 

0xMirasio:
Hi
I added basic xml support for PPC arch for gdb debugger
Also i update a little drcov coverage : added some constructor functions in the drcov class, that allow python set() to work (I have some cases where i have a lot of drcov trace for the same binary, this help a lot to filter and set unique bbs)

(sorry for duplicates got issues with git) 

- [ X] This PR only contains minor fixes.

### Coding convention?

- [ X] The new code conforms to Qiling Framework naming convention.
- [ X] The imports are arranged properly.
- [ X] Essential comments are added.
- [ X] The reference of the new code is pointed out.

### Extra tests?

- [ X] Tests will be added after some discussion and review.

### Changelog?

- [ X] Changelog will be updated after some proper review.

### Target branch?

- [X ] The target branch is dev branch.

### One last thing

- [ X] I have read the [contribution guide](https://docs.qiling.io/en/latest/contribution/)

-----


elicn:
Thanks for contributing to the project.
Could you please clarify what is the addition in `drcov`? I don't see any `set` that could benefit from making `bb_entry` hashable. Also, this addition appears to discard the class `c_types.Structure` nature by defining those methods (especially `__init__` which actually uses a different set of members name)

0xMirasio:
Ah sorry didn't see that `__init__`  has override c_types.Structure. Should i remove the hashable functionnality ? 

I use it to hash bb_entry , with that i can filter most of duplicate entry. 
One example is for fuzzing for example, when you have 300+ corpus outputs, you can use hashable bb_entry to filter duplicates one and generate a single coverage file. 

```
bb_gbl = set()
for corpus in corpus_list:
    bb_gbl |= set(test_corpus(corpus)) 
    
dump_coverage("out.cov", ql, list(bb_gbl)) 
```

If not wanted or problematic, i can remove it from PR. Or maybe we have a simpler alternative that's doesn't override c_types.Structure init?

0xMirasio:
yea thought about it and i should use:

```
def __init__(self, start, size, module_id=0):
    super().__init__(start, size, module_id) 
    self.start = start
    self.size = size
    self.module_id = module_id
```

We still use original c_type.Structure constructor + we can use hashable bbEntry

elicn:
Thanks for the clarification.

Though I am not sure I understand why this has to be part of the project.. As far as I understand, this can be part of your own private script that de-duplicates covered basic blocks. In case you would like to incorporate the de-duplication code into the project as well, then it will make more sense to include this change.

BTW - even in this case the `__init__` method is redundant, it is already implied.

Please let me know if I misunderstood this.

0xMirasio:
ye i understand. Do you want me to remove it ?

elicn:
If this addition (that is, the methods that helps turning it into hashable) is there only to serve your own scripts, then it shouldn't be there. In other words, either we introduce a feature that uses it and all Qiling users can benefit from, or we drop the changes.

elicn (approved):

xwings (approved):

xwings:
Thanks!

PR #1601, opened 2025-11-25
Fix elf stack, pe stack and pe sections memory protection

xlatbx59:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

When running some pe binaries and having them throwing unhandled exceptions, I noticed that all all the protections are RWX by default.
That could be used to detect qiling from a malware perspective and here's how it can be done:
```cpp
#include <iostream>
#include <windows.h>

static void qiling_detected()       //Runs if the stack is executable which isn't normal for modern winodws systems
{
    printf("[-]Will terminate because qiling was detected...");
    ExitProcess(0);
}
static LONG doing_evil_stuff(_EXCEPTION_POINTERS* ExceptionInfo)    //Runs if the stack is not executable which is the norm for modern windows systems
{
    printf("[+]No qiling, doing evil stuff");
    return 0;
}


int main()
{
    DWORD old_mem = 0;
    byte some_code[3] = { 0 };          //call [rcx]
    some_code[0] = 0xff;                //ret
    some_code[1] = 0xd1;
    some_code[2] = 0xc3;
    //The call rcx, ret, is executed on the stack which isn't supposed to happen
    void(__fastcall * shellcode)(void*) = (void(__fastcall *)(void*)) (void*)some_code;

    AddVectoredExceptionHandler(1, doing_evil_stuff);
    shellcode(&qiling_detected);

    return 0;
}
```

### DIFF
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -88,7 +88,7 @@ class QlLoaderELF(QlLoader):
         stack_address = self.profile.getint('stack_address')
         stack_size = self.profile.getint('stack_size')
         top_of_stack = stack_address + stack_size
-        self.ql.mem.map(stack_address, stack_size, info='[stack]')
+        self.ql.mem.map(stack_address, stack_size, UC_PROT_READ | UC_PROT_WRITE, info='[stack]')
 
         self.path = self.ql.path
 
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -92,13 +92,9 @@ class QlLoaderELF(QlLoader):
         elffile = ELFFile(fstream)
         elftype = elffile['e_type']
 
-        stack_perm = UC_PROT_NONE
-        for seg in elffile.iter_segments('PT_GNU_STACK'):
-            stack_perm = QlLoaderELF.seg_perm_to_uc_prot(seg['p_flags'])
-
-        QlLoaderELF.seg_perm_to_uc_prot(stack_perm)
-
         # setup program stack
+        stack_seg = elffile.iter_segments('PT_GNU_STACK')
+        stack_perm = QlLoaderELF.seg_perm_to_uc_prot(next(stack_seg)['p_flags'])
         stack_address = self.profile.getint('stack_address')
         stack_size = self.profile.getint('stack_size')
         top_of_stack = stack_address + stack_size
--- a/qiling/loader/elf.py
+++ b/qiling/loader/elf.py
@@ -84,12 +84,6 @@ class QlLoaderELF(QlLoader):
 
         self.profile = self.ql.os.profile[f'OS{self.ql.arch.bits}']
 
-        # setup program stack
-        stack_address = self.profile.getint('stack_address')
-        stack_size = self.profile.getint('stack_size')
-        top_of_stack = stack_address + stack_size
-        self.ql.mem.map(stack_address, stack_size, UC_PROT_READ | UC_PROT_WRITE, info='[stack]')
-
         self.path = self.ql.path
 
         with open(self.path, 'rb') as infile:
@@ -97,6 +91,18 @@ class QlLoaderELF(QlLoader):
 
         elffile = ELFFile(fstream)
         elftype = elffile['e_type']
+
+        stack_perm = UC_PROT_NONE
+        for seg in elffile.iter_segments('PT_GNU_STACK'):
+            stack_perm = QlLoaderELF.seg_perm_to_uc_prot(seg['p_flags'])
+
+        QlLoaderELF.seg_perm_to_uc_prot(stack_perm)
+
+        # setup program stack
+        stack_address = self.profile.getint('stack_address')
+        stack_size = self.profile.getint('stack_size')
+        top_of_stack = stack_address + stack_size
+        self.ql.mem.map(stack_address, stack_size, stack_perm, info='[stack]')
 
         # is it a driver?
         if elftype == 'ET_REL':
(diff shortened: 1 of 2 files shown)


elicn:
Hi there, welcome and thanks for the contribution.

Setting stack pages as non-executable is also inaccurate since stack pages may be either executable or not based on how the file was compiled. Stack attributes may be found in the `GNU_STACK` segment in ELF files, I guess there is a similar place for PE, I just don't remember where off hand.

As for mapping PE sections with their respective attributes, this was already implemented for UEFI PE loader [here](https://github.com/qilingframework/qiling/blob/fe462b8313aa46d84cac5a0f7b207382fb25a37b/qiling/loader/pe_uefi.py#L78), I just forgot to port it to the Windows PE loader...

Do you think you could extract the correct stack attributes and use them?

xlatbx59:
I'm on it

qiling/loader/elf.py
```diff
@@ -98,6 +92,18 @@ def run(self):
         elffile = ELFFile(fstream)
         elftype = elffile['e_type']
 
+        stack_perm = UC_PROT_NONE
+        for seg in elffile.iter_segments('PT_GNU_STACK'):
+            stack_perm = QlLoaderELF.seg_perm_to_uc_prot(seg['p_flags'])
+
+        QlLoaderELF.seg_perm_to_uc_prot(stack_perm)
```
elicn: Why iterating through segments where there is just one (or none)?
It would make more sense to get the segment by name and then extract its flags.

Also, in case segment is found, `stack_perm` already holds UC Protection value, not sure what happens on line 99 (converting again? where the returned value goes?)

xlatbx59:
Tried porting your function to QlLoaderPE, changed some stuff but couldn't manage to add sections infos for each one of them, plus when executing some pe binaries(qiling\\examples\\rootfs\\x8664_windows\\bin\\x8664_clipboard_test.exe, to be exact), I get this error message when changing the memory protection of a section:
```
Cannot change mapinfo at 0x140011000-0x140019000
```

qiling/loader/pe.py
```diff
@@ -902,7 +934,7 @@ def load(self, pe: Optional[pefile.PE]):
             self.ql.log.info(f'Loading {self.path} to {image_base:#x}')
             self.ql.log.info(f'PE entry point at {self.entry_point:#x}')
 
-            self.ql.mem.map(image_base, image_size, info=f'{image_name}')
+            _map_sections_(self.ql, pe, image_base, image_size, image_name)
```
elicn: As suggested on the UEFI PE loader, where this code was taken from, some PE files might not be suitable to be mapped like that. Setting permissions by page is doable only when the file segments are page-aligned; if they are not, one page may contain multiple segments with different permissions, we have to map the file as a whole (all RWX). We should maintain the same behavior here.

qiling/loader/pe.py
```diff
@@ -932,7 +964,6 @@ def load(self, pe: Optional[pefile.PE]):
             pe.parse_data_directories()
 
             # done manipulating pe file; write its contents into memory
-            self.ql.mem.write(image_base, bytes(pe.get_memory_mapped_image()))
```
elicn, line 935: I believe file contents cannot be written till this point, mainly because of the `pe.parse_data_directories` 2 lines above that still manipulates the file contents.

elicn (changes requested):
Thanks for making the adjustments.
Please see my comments.

qiling/loader/pe.py
xlatbx59: so, should I just leave it as it is with rwx for every section?

elicn: No. The point is that executables could be either compatible with this kind of mapping or not based on whether their sections are page-aligned [or not]. If they are not compatible, the only way is to map them as a whole with RWX permissions. Follow the UEFI PE loader to see how it works.

xlatbx59:
Hoping this is good now

qiling/loader/elf.py
```diff
@@ -92,13 +92,9 @@ def run(self):
...
 
-        stack_perm = UC_PROT_NONE
-        for seg in elffile.iter_segments('PT_GNU_STACK'):
-            stack_perm = QlLoaderELF.seg_perm_to_uc_prot(seg['p_flags'])
-
-        QlLoaderELF.seg_perm_to_uc_prot(stack_perm)
-
         # setup program stack
+        stack_seg = elffile.iter_segments('PT_GNU_STACK')
+        stack_perm = QlLoaderELF.seg_perm_to_uc_prot(next(stack_seg)['p_flags'])
```
elicn, line 97: What if there is no such segment in the file? The code should be prepared to handle this situation (that is, default to RWX).

qiling/loader/pe.py
```diff
@@ -382,7 +382,8 @@ def load_dll(self, name: str, is_driver: bool = False) -> int:
         dll_len = image_size
 
         self.dll_size += dll_len
-        _map_sections_(self.ql, dll, dll_base, dll_len, dll_name)
+        _map_pe_(self.ql, dll, dll_base, dll_len, dll_name)
+        self.ql.mem.write(dll_base, bytes(data))
```
elicn, line 386: That is not going to work... Mapping a PE file section be section does not necessarily form a continuous block in memory, so we cannot just `mem.write` the file as a whole.

As far I can see, there should be two methods: one that maps _and_ writes the PE file section by section, and one that maps _and_ writes the PE file as a whole. Just like it is on the UEFI PE loader.

If this is too much to handle, it is OK. We can stay with the stack permission adjustment and leave the PE mapping by sections to another time.

elicn (changes requested):

PR #1611, opened 2026-04-19
timestamp modification system call support

libumem:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

This PR adds support for the syscalls:

* `utime`
* `futimesat`
* `utimes`
* `utimensat`

and adds source code for an example to be merged in the rootfs repo.


libumem:
Converting to draft until https://github.com/qilingframework/rootfs/pull/37 is handled

xwings:
Two issues need fixes before merge:

1. `times` is nullable for `utime(2)`, `utimes(2)`, and `utimensat(2)`. Linux treats a NULL pointer as "set atime/mtime to now" subject to the normal permission rules. This implementation currently either dereferences `0` in `do_utime()` or returns positive `EACCES` from `ql_syscall_utimensat()`, which is not a syscall error under Qiling return conventions. Please handle `times == 0` by setting both timestamps to now and return negative errno values on failure.

2. `futimesat(2)` takes `struct timeval times[2]`, while `utimensat(2)` takes `struct timespec times[2]`. The current `ql_syscall_futimesat()` delegates to the `utimensat` path, so the second field is parsed as nanoseconds instead of microseconds. Please use a timeval-aware path like `utimes()` and multiply `tv_usec` by 1000 before calling `os.utime(ns=...)`.

libumem:
@xwings Finally got some free time to work on this again. Please take another look. 

I've had to comment out a bunch of test cases (mostly arm64) that were having errors unrelated to anything I am touching in the codebase.

xwings:
Hi,

Will you be able to fix the conflict ?

libumem:
@xwings Done

PR #1630, opened 2026-06-21
examples/shellcode_run.py: skip the Windows stages when system DLLs are absent

### LINKED ISSUE #1629: examples/shellcode_run.py crashes: unhandled CPU exceptions (ARM64/ARM/MIPS) and missing-DLL handling (Windows)
**Describe the bug**

Running `examples/shellcode_run.py` does not complete. It crashes on the very first stage and never reaches the later ones. There are two distinct problems.

**1. Unhandled CPU exception on ARM64 / ARM / MIPS (`QlErrorCoreHook`)**

Since `execve` was patched to return `-1` for paths outside the rootfs (commit "Patch POSIX execve"), the rootfs-less shellcodes in this example no longer stop at `execve`. Execution falls through into the trailing data (the sockaddr struct and the `/bin/sh` string), which decodes as an illegal instruction:

- ARM64 / ARM: undefined instruction → exception `1` (`EXCP_UDEF`)
- MIPS: reserved instruction → exception `20` (`EXCP_RI`)

The Linux OS layer only hooks the syscall trap (exception `2` for ARM/ARM64, `17` for MIPS), so these exceptions reach the unhandled-interrupt dispatcher and abort the whole run:

```
qiling.exception.QlErrorCoreHook: _hook_intr_cb : not handled
```

A real kernel would deliver `SIGILL` and terminate the process, rather than crashing the framework.

**2. Windows stages crash with `UC_ERR_READ_UNMAPPED` when system DLLs are absent**

The two Windows stages need genuine Windows system DLLs (collected via `examples/scripts/dllscollector.bat`), which are not shipped. On a stock checkout the loader logs `Could not find DLL file: .../Windows/System32/ntdll.dll` but then continues and dies with an unhandled `unicorn.UcError: Invalid memory read (UC_ERR_READ_UNMAPPED)` instead of failing with a clear, actionable message.

**Sample Code**
```python
# examples/shellcode_run.py, first stage
from qiling import Qiling
from qiling.const import QL_ARCH, QL_OS, QL_VERBOSE

ARM64_LIN = bytes.fromhex(
    '420002ca210080d2400080d2c81880d2010000d4e60300aa01020010020280d2'
    '681980d2010000d4410080d2420002cae00306aa080380d2010000d4210400f1'
    '65ffff54e0000010420002ca210001caa81b80d2010000d4020004d27f000001'
    '2f62696e2f736800')

ql = Qiling(code=ARM64_LIN, archtype=QL_ARCH.ARM64, ostype=QL_OS.LINUX, verbose=QL_VERBOSE.DEBUG)
ql.run()   # -> QlErrorCoreHook: _hook_intr_cb : not handled
```

**Expected behavior**

`examples/shellcode_run.py` should run to completion. Illegal-instruction exceptions on the Linux archs should be treated as a fatal signal that ends the emulated process cleanly (as a real kernel would), and the Windows stages should be skipped with a helpful message when the required DLLs are not present.

**Additional context**

- Qiling: 1.4.10 (reproduced on current `master`, `da210f07`)
- Unicorn: 2.1.3
- Python: 3.12.3
- Host: Linux

I have fixes for all of the above and will open a PR shortly:
- hook the undefined/reserved-instruction exceptions on ARM/ARM64/MIPS Linux and end emulation cleanly (SIGILL semantics);
- detect missing Windows DLLs in the example and skip those stages gracefully;
- plus a MIPS32 big-endian shellcode test.

---
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

retrocpugeek:
Split as requested — this PR is now only the `examples/shellcode_run.py` change. The framework-side illegal-instruction handling moved to #1661.

`shellcode_run.py` runs a Windows x86 and x86-64 stage that need genuine Windows system DLLs. Those cannot be redistributed, so they are not shipped with Qiling and must be collected from a licensed Windows host with `examples/scripts/dllscollector.bat`.

Without them the example does not fail gracefully — it crashes on an unmapped read partway through. This probes for `Windows/System32/ntdll.dll` under each Windows rootfs and, when it is missing, prints a message pointing at the collector script instead of running that stage.

### DIFF
--- a/examples/shellcode_run.py
+++ b/examples/shellcode_run.py
@@ -3,11 +3,22 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
+import os
 import sys
 
 sys.path.append("..")
 from qiling import Qiling
 from qiling.const import QL_ARCH, QL_OS, QL_VERBOSE
+
+
+def windows_rootfs_ready(rootfs: str) -> bool:
+    # The Windows examples need genuine Windows system DLLs, which cannot be
+    # redistributed and are therefore not shipped with Qiling. They must be
+    # collected from a licensed Windows host using the helper script at
+    # examples/scripts/dllscollector.bat. Probe for ntdll.dll so we can skip
+    # these stages with a helpful message instead of crashing on an unmapped
+    # read when the DLLs are absent.
+    return os.path.isfile(os.path.join(rootfs, 'Windows', 'System32', 'ntdll.dll'))
 
 X86_LIN = bytes.fromhex('31c050682f2f7368682f62696e89e3505389e1b00bcd80')
 X8664_LIN = bytes.fromhex('31c048bbd19d9691d08c97ff48f7db53545f995257545eb03b0f05')
@@ -87,12 +98,20 @@ if __name__ == "__main__":
     ql.run()
 
     print("\nWindows x86 Shellcode")
-    ql = Qiling(code=X86_WIN, archtype=QL_ARCH.X86, ostype=QL_OS.WINDOWS, rootfs=r'rootfs/x86_windows')
-    ql.run()
+    if windows_rootfs_ready(r'rootfs/x86_windows'):
+        ql = Qiling(code=X86_WIN, archtype=QL_ARCH.X86, ostype=QL_OS.WINDOWS, rootfs=r'rootfs/x86_windows')
+        ql.run()
+    else:
+        print("  [skipped] Windows system DLLs not found under rootfs/x86_windows/Windows/System32.")
+        print("  Collect them from a licensed Windows host with examples/scripts/dllscollector.bat.")
 
     print("\nWindows x86-64 Shellcode")
-    ql = Qiling(code=X8664_WIN, archtype=QL_ARCH.X8664, ostype=QL_OS.WINDOWS, rootfs=r'rootfs/x8664_windows')
-    ql.run()
+    if windows_rootfs_ready(r'rootfs/x8664_windows'):
+        ql = Qiling(code=X8664_WIN, archtype=QL_ARCH.X8664, ostype=QL_OS.WINDOWS, rootfs=r'rootfs/x8664_windows')
+        ql.run()
+    else:
+        print("  [skipped] Windows system DLLs not found under rootfs/x8664_windows/Windows/System32.")
+        print("  Collect them from a licensed Windows host with examples/scripts/dllscollector.bat.")
 
     # FIXME: freebsd sockets are currently broken.
     #


xwings:
Hey,

Welcome to QIling Framework and thanks for the PR.

PR looks good to me. Lets wait for second approval

xwings (approved):

retrocpugeek:
> Hey,
> 
> Welcome to QIling Framework and thanks for the PR.
> 
> PR looks good to me. Lets wait for second approval

Hi, thanks for the welcome. Love the project.
I've been working towards getting mips64be support for a project I'm working on.
It's almost there - although it's dependent on a couple of issues/PRs in unicorn - I've created https://github.com/qilingframework/qiling/issues/1645 to track them.

qiling/os/linux/linux.py
```diff
@@ -137,6 +142,24 @@ def setup_procfs(self):
...
+        # the image.
+        signame = {
+            EXCP.UDEF:      'SIGILL',   # ARM / ARM64 undefined instruction
+            MIPS_EXCP.RI:   'SIGILL',   # MIPS reserved (illegal) instruction
+        }.get(intno, f'exception {intno:#x}')
+
+        pc = ql.arch.regs.arch_pc
+
+        ql.log.debug(f'CPU raised {signame} at {pc:#x}; terminating emulated process')
+        ql.stop()
```
xwings: One issue must be resolved before this can merge; it has its own section below.

## Issue 1 — major in `qiling/os/linux/linux.py:161`

```
  158 |         pc = ql.arch.regs.arch_pc
  159 |
  160 |         ql.log.debug(f'CPU raised {signame} at {pc:#x}; terminating emulated process')
> 161 |         ql.stop()
  162 |
  163 |     def register_function_after_load(self, function):
  164 |         if function not in self.function_after_load_list:
```

**Problem.** Calling ql.stop() does not terminate the enclosing Linux run when a handled CPU fault occurs during single-threaded ELF interpreter execution. On ARM, ARM64, or MIPS, with distinct interpreter and executable entry points and no explicit begin override, it stops only the current emulation segment. QlOsLinux.run() then proceeds through patches and after-load callbacks to executable startup at lines 201–207, despite the handler reporting fatal termination. Previously, absent a custom handler, the unhandled interrupt propagated as QlErrorCoreHook and prevented that continuation.

**How to change it.** Track fatal CPU termination separately from ordinary emulation-segment completion: reset a flag at Linux run entry, set it before stopping in this handler, and return immediately after interpreter emulation when the flag is set, before patches, callbacks, or executable startup. Do not use the STOPPED emulation state as the distinction, since normal segment completion also sets it.

Suggested change — a sketch to adapt, not a patch to apply:

```python
# At the beginning of QlOsLinux.run():
self._fatal_cpu_exception = False

# In hook_cpu_exception(), replace the standalone ql.stop():
self._fatal_cpu_exception = True
ql.stop()

# In run(), guard continuation immediately after interpreter emulation:
self.ql.emu_start(self.ql.loader.entry_point, entry_address, self.ql.timeout)
if self._fatal_cpu_exception:
    return

self.ql.do_lib_patch()
```

Your changes address the documented shellcode demonstration with actionable Windows prerequisite messages, named exception constants, and a big-endian MIPS case. The panel found the scope and placement justified, while recognizing that example-local syscall interception remains a smaller alternative to changing Linux fault handling.

xwings:
Hi,

Welcome to QIling Framework. 

Except the ql.stop(), it would be better if we split into two differnt PR for two differnt code.

retrocpugeek:
Ok, using this PR for the 'skip if windows DLLs absent' part

xwings:
Hi,

This PR seems like makes no different. if the rootfs does not exits. it will not run anyway.

retrocpugeek:
I think we're talking about two different cases here

There's 

**A/ rootfs directory absent**

user will get
`QlErrorFileNotFound: Target rootfs not found: "..."`
which is a clear error. No fix needed as I see it.

**B/ rootfs directory present, but no Windows/System32/*.dll**

user will get
`UcError: Invalid memory read (UC_ERR_READ_UNMAPPED)`
with no hint that DLLs are the problem. This is what the PR fixes.

PR #1631, opened 2026-06-21
Add MIPS64 (n64 ABI) Linux support

retrocpugeek:
## Summary

Adds `QL_ARCH.MIPS64` so Qiling can emulate MIPS64 Linux user-space in both big-
and little-endian, built out from raw shellcode up to full dynamically-linked
glibc ELF binaries. The native backends already support MIPS64 (Unicorn
`UC_MODE_MIPS64`, Capstone/Keystone `MIPS64`); this wires it through the framework
and fixes the big-endian / n64 correctness issues that only surface once real
glibc binaries run.

## Not mergeable yet — two external dependencies

1. **qilingframework/rootfs#38** ("Add MIPS64 (n64 ABI) Linux test binaries") adds
   the `mips64_linux/`, `mips64el_linux/` and `mips64_linux_buildroot/` trees.
   Without it the MIPS64 `test_elf.py` cases fail with `QlErrorFileNotFound`.
2. **Upstream unicorn vtlb fixes** (unicorn-engine/unicorn#2346 and #2347, both
   still open). MIPS64 needs `UC_TLB_VIRTUAL` to reach the n64 load address above
   `useg`, which exposes two unicorn bugs. On stock unicorn 2.1.3 the tests that
   run a real glibc binary fail; the shellcode, `stat64` and `getdents` tests pass.
   Tracked in #1645.

This PR is ready for review but should not merge until a unicorn release carries
both fixes and the pin in `pyproject.toml` moves.

## What's included

- **`QlArchMIPS64`** (`qiling/arch/mips64.py`) — 64-bit arch class reusing the
  shared MIPS register map; defaults to a MIPS64R2-class CPU model.
- **n64 calling convention & syscall ABI** — first eight args in `a0`–`a7`
  (the tail four are the physical registers `$8`–`$11`) with no o32 shadow space;
  n64 syscall table (base 5000) from `asm/unistd_n64.h`; MIPS64 routed through the
  MIPS arch-specific syscall constants.
- **ELF loading** — `EM_MIPS` + `ELFCLASS64` detection and n64 doubleword stack
  alignment; static, PIE, and non-PIE (`ET_EXEC`) dynamically-linked big-endian
  MIPS64 binaries.
- **Big-endian / n64 syscall correctness fixes** — statx and stat64 struct byte
  order on BE guests; legacy `getdents` and `getdents64` record alignment for the
  n64 unaligned walk; MIPS `SO_*` socket option values.
- **clone3 + multithreading** support, with an example and test.
- **gdb-stub support** — MIPS64 register description (`debugger/gdb/xml/mips64/`)
  and `QL_ARCH.MIPS64` wiring in the gdb debugger, so `ql.debugger = "gdb:..."`
  works for MIPS64 targets (verified with `gdb-multiarch`: registers, memory,
  disassembly, single-step, breakpoints, continue, backtraces). Registers a
  target description declares but the backend can't supply are reported as zero
  rather than `<unavailable>`, so stricter clients (e.g. Ghidra's `ghidragdb`)
  can attach. The stub also honours an async interrupt (ctrl-c / `\x03`): the
  run hook polls the client socket during `emu_start`, so a free-running guest
  (an idle/event loop) can be paused from the client rather than only stopping
  at a breakpoint or exit.
- **Rootfs & tests** — static and dynamic MIPS64 hello binaries, a Buildroot glibc
  tree split into its own rootfs, and shellcode / ELF / statx / multithread
  regression tests.

## Commits

**Core arch & n64 ABI**
- Add MIPS64 (n64 ABI) Linux support — `QlArchMIPS64`, n64 calling convention +
  syscall table, ELF machine detection, doubleword stack alignment
- Route MIPS64 through the MIPS arch-specific syscall constants
- Support loading MIPS64 ELF binaries
- Fix MIPS socket option (SO_*) values

**Examples & features**
- Add `examples/shellcode_mips64_run.py`
- Add MIPS64 customapi examples and function-hook support
- Add clone3 syscall and MIPS64 multithreading example/test

**Big-endian / n64 correctness**
- Fix statx byte order on big-endian guests
- Wire up stat64 for MIPS64
- Align getdents64 records to fix unaligned access on MIPS
- Align legacy getdents records (fixes n64 unaligned walk)
- Add MIPS64 EB statx regression test

**Rootfs & test binaries**
- Bump rootfs submodule to MIPS64 static glibc hello binaries
- Bump rootfs submodule for MIPS64 dynamic hello binaries
- tests: add MIPS64 BE Buildroot dynamic-load regression + rootfs split

**Debugging**
- Add MIPS64 gdb-stub support (register XML + `QL_ARCH.MIPS64` in the gdb debugger)
- gdb: report 0 for unavailable registers so strict clients can attach
- gdb: support async interrupt (ctrl-c) to break into a running guest

## Testing

The MIPS64 paths are covered by the existing suites: `test_shellcode.py`
(`test_linux_mips64eb` / `test_linux_mips64el`), `test_elf.py` (MIPS64 EB static,
Buildroot dynamic, statx, stat64, getdents byte-order/alignment),
`test_elf_multithread.py` (clone3 multithreading), and `test_debugger.py`
(`test_gdbdebug_mips64`).

The `test_elf.py` and `test_elf_multithread.py` MIPS64 cases need the binaries from
rootfs#38; the shellcode tests are self-contained and pass without it.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


retrocpugeek:
### Follow-up note: reconcile with #1630

This PR and #1630 both modify the MIPS branch in `qiling/os/linux/linux.py`, so whichever merges second will need a small reconciliation.

#1630 introduces a shared `hook_cpu_exception` handler (faithful `SIGILL` on illegal instructions) and a `MIPS_EXCP` enum, and registers them for MIPS32. To keep this PR self-contained against `dev`, MIPS64 here mirrors dev's current MIPS32 convention — it hooks the syscall exception with the literal number `17` and does **not** register the RI illegal-instruction handler.

Once both land, MIPS64 should also be wired to `hook_cpu_exception` on `MIPS_EXCP.RI` (and switch `17` → `MIPS_EXCP.SYSCALL`), so it terminates cleanly on illegal instructions the same way MIPS32 will. The MIPS64 shellcode test here (`write` + `exit_group`) doesn't exercise that path, which is why it's safe to defer.

xwings:
Hi, 

Welcome to QIling Framework and thanks for the PR.
Please take a look at the conflit and we need to also have a differnt PR in rootfs.

retrocpugeek:
Rebased onto `dev`, conflicts resolved.

Rootfs PR is open: qilingframework/rootfs#38. CI here will fail at submodule checkout until that one merges, since the submodule pointer isn't reachable upstream yet.

retrocpugeek:
Reconciled now that #1661 has merged: rebased onto `dev`, and MIPS64 joins the MIPS32 branch in `linux.py` — `MIPS_EXCP.SYSCALL` for the syscall trap plus `hook_cpu_exception` on `MIPS_EXCP.RI`. The follow-up note above is no longer outstanding.

PR #1632, opened 2026-06-23
Fix capstone 6 compatibility for ARM64 and RISC-V

kripticni:
Capstone 6 renamed several architecture constants in its Python bindings:
- CS_ARCH_ARM64 -> CS_ARCH_AARCH64 (old name removed)
- CS_MODE_RISCVC -> CS_MODE_RISCV_C (old name still aliased in v6, but future
  versions may drop it)

This PR:
- Adds try/except fallback in arm64.py for the ARM64 / AARCH64 rename
- Adds forward-compatible try/except in riscv.py and riscv64.py for RISCVC /
  RISCV_C
- Relaxes the capstone dependency from ^5 to >=5,<7 to allow capstone 6.x

Target branch: dev

Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ keywords = [
 
 [tool.poetry.dependencies]
 python = "^3.10"
-capstone = "^5"
+capstone = ">=5,<7"
 unicorn = "2.1.3"
 pefile = ">=2022.5.30"
 python-registry = "^1.3.1"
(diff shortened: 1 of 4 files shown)


xwings:
Hi

Welcome to QIling Framework. Thanks for the PR

Made some comments. please do let me know what do you think

kripticni:
@xwings , thanks for taking a look. I checked the files diff but dont see any inline review comments on my end. Could you double check that your review was submitted? Sometimes GitHub saves them as pending and they dont show up for the contributor.

xwings:
sorry, my bad. 

but for capstone we should just move to Capstone 6 and we dont have to do backward compatabalities. So, the try catch exception is not needed.

PR #1634, opened 2026-06-24
Fix statx byte order on big-endian guests

### LINKED ISSUE #1633: statx returns a byte-swapped struct on big-endian guests (directories misdetected)
# `statx` returns a byte-swapped struct on big-endian guests (directories misdetected)

## Describe the bug

`ql_syscall_statx` (`qiling/os/posix/syscall/stat.py`) always serializes the result with the little-endian `Statx32` / `Statx64` structs (`ctypes.Structure`), regardless of the guest's endianness:

```python
if ql.arch.bits == 32:
    Statx = Statx32
else:
    Statx = Statx64
```

Every *other* stat-family struct in this file already has an explicit big-endian variant (`LinuxMips64EBStat`, `LinuxARMEBStat`, …) selected via `ql.arch.endian`. `statx` is the only one missing it, so on a big-endian guest (MIPS/MIPS64 EB, ARM EB, PPC, …) **every field of `struct statx` is byte-swapped**.

The most visible consequence is `stx_mode`: for a directory the type bits land in the wrong byte, so `S_ISDIR` is false on a big-endian guest. A directory is reported as a non-directory, and e.g. `busybox ls /` prints just the path name instead of listing the directory. Because modern glibc routes `stat`/`fstatat` through `statx`, this affects ordinary directory operations, not only direct `statx` callers.

## Repro (pure Python, no guest binary)

This serializes the exact struct Qiling emits and shows it reads back wrong on a big-endian guest:

```python
import os
from qiling.os.posix.syscall.stat import Statx64

st_mode = os.stat('/').st_mode                       # a directory: 0o40755
raw = bytes(Statx64(stx_mode=st_mode))
off = Statx64.stx_mode.offset
field = raw[off:off+2]

print('host / st_mode      = 0o%o  (S_IFDIR=%s)' % (st_mode, (st_mode & 0o170000) == 0o040000))
print('stx_mode bytes      = %s  (offset %d)' % (field.hex(), off))
le = int.from_bytes(field, 'little')
be = int.from_bytes(field, 'big')
print('LE guest reads mode = 0x%04x -> type 0o%o' % (le, le & 0o170000))
print('BE guest reads mode = 0x%04x -> type 0o%o' % (be, be & 0o170000))
```

Output:

```
host / st_mode      = 0o40755  (S_IFDIR=True)
stx_mode bytes      = ed41  (offset 28)
LE guest reads mode = 0x41ed -> type 0o40000     # S_IFDIR  -> directory
BE guest reads mode = 0xed41 -> type 0o160000    # not S_IFDIR -> "not a directory"
```

`S_IFDIR` is `0o40000`. The struct is written little-endian (`ed 41`), so a little-endian guest reads `0x41ed` (correct) while a big-endian guest reads `0xed41` → `0o160000`, which is not a directory type.

## Expected behavior

`statx` should serialize the struct in the guest's byte order, so `stx_mode` (and every other field) reads back correctly. `S_ISDIR` should be true for a directory on big-endian guests, exactly as it already is on little-endian ones.

## End-to-end confirmation

A static MIPS64 **big-endian** binary that calls `statx(AT_FDCWD, "/", 0, STATX_BASIC_STATS, &stx)` and prints `DIR` iff `S_ISDIR(stx.stx_mode)` (source: `examples/src/linux/statx.c`):

- **Before the fix:** `NOTDIR mode=176501`  (the byte-swapped mode)
- **After the fix:** `DIR`

## Why it was never caught

`statx` is only emitted by recent glibc; the big-endian binaries in the test rootfs are older/static and use `stat`/`fstat`/`lstat`, which already have big-endian struct variants. And no existing test inspects a directory's *mode* via `statx` on a big-endian guest. (Same bug class as the already-fixed #521, where `bind` unpacked `sin_family` with the wrong endianness on MIPS MSB.)

## Environment

- Qiling: `dev` (reproduced through current HEAD); the Python repro above reproduces on any host.
- Affected guests: any big-endian Linux target (MIPS/MIPS64 EB, ARM EB, PPC, …).
- Not arch-specific in Qiling code — it's the shared `statx` syscall handler.

## Proposed fix

Add big-endian variants `Statx32EB` / `Statx64EB` (and `StatxTimestamp32EB` / `StatxTimestamp64EB`) as `ctypes.BigEndianStructure`, reusing the little-endian field layouts, and select them in `ql_syscall_statx` when `ql.arch.endian == QL_ENDIAN.EB`. The little-endian path is byte-for-byte unchanged.

Implemented on `retrocpugeek:feature/mips64be`:
- Fix: `Fix statx byte order on big-endian guests` (commit `7550983b`)
- Regression test: `test_elf.ELFTest.test_elf_linux_mips64eb_statx` (asserts `statx` reports `/` as `DIR`) + `examples/src/linux/statx.c`

Happy to open a PR against `dev`.

retrocpugeek:
## What

Fixes `statx` returning a byte-swapped `struct statx` on big-endian guests.

`ql_syscall_statx` always serialized the result with the little-endian `Statx32` / `Statx64` (`ctypes.Structure`), regardless of guest endianness. Every other stat-family struct in `stat.py` already has a big-endian variant selected via `ql.arch.endian`; `statx` was the only one missing it. On a big-endian guest this byte-swaps every field — most visibly `stx_mode`, which loses its `S_IFDIR` bit, so a directory looks like a plain file (e.g. `busybox ls /` prints the path name instead of listing it). Modern glibc routes `stat`/`fstatat` through `statx`, so this affects ordinary directory operations, not just direct `statx` callers.

Closes #1633.

## How

- Add `Statx32EB` / `Statx64EB` (and `StatxTimestamp32EB` / `StatxTimestamp64EB`) as `ctypes.BigEndianStructure`, reusing the little-endian field layouts.
- Select the big-endian variants in `ql_syscall_statx` when `ql.arch.endian == QL_ENDIAN.EB`.
- The little-endian path is byte-for-byte unchanged.

## Test

`test_elf.ELFTest.test_linux_statx_bigendian` drives `ql_syscall_statx` on a big-endian MIPS context and checks that `stx_mode`, read back in the guest's byte order, still reports a directory. It is self-contained (no rootfs binary, runs on stock unicorn) and fails on `dev` (`AssertionError`), passes with this fix.

## Checklist

- [X] I have added enough tests for this PR.

### DIFF
--- a/qiling/os/posix/syscall/stat.py
+++ b/qiling/os/posix/syscall/stat.py
@@ -1413,17 +1413,45 @@ class Statx64(ctypes.Structure):
 
     _pack_ = 4
 
+# Big-endian counterparts of the statx structs. The kernel statx layout is the
+# same on every architecture, so the big-endian variants reuse the little-endian
+# field lists verbatim and only change the ctypes base class (and the nested
+# timestamp type, which must itself be big-endian). Without these, statx() byte-
+# swaps every field on a big-endian guest (e.g. MIPS/MIPS64 EB), so stx_mode
+# loses its S_IFDIR bit and tools like `ls` treat directories as plain files.
+class StatxTimestamp32EB(ctypes.BigEndianStructure):
+    _fields_ = StatxTimestamp32._fields_
+
+class StatxTimestamp64EB(ctypes.BigEndianStructure):
+    _fields_ = StatxTimestamp64._fields_
+
+def _statx_fields_eb(fields):
+    swap = {StatxTimestamp32: StatxTimestamp32EB, StatxTimestamp64: StatxTimestamp64EB}
+    return [(name, swap.get(ftype, ftype)) for (name, ftype) in fields]
+
+class Statx32EB(ctypes.BigEndianStructure):
+    _fields_ = _statx_fields_eb(Statx32._fields_)
+    _pack_ = 8
+
+class Statx64EB(ctypes.BigEndianStructure):
+    _fields_ = _statx_fields_eb(Statx64._fields_)
+    _pack_ = 4
+
 # int statx(int dirfd, const char *restrict pathname, int flags,
 #                  unsigned int mask, struct statx *restrict statxbuf);
 def ql_syscall_statx(ql: Qiling, dirfd: int, path: int, flags: int, mask: int, buf_ptr: int):
+    is_eb = ql.arch.endian == QL_ENDIAN.EB
+
     def statx_convert_timestamp(tv_sec, tv_nsec):
         tv_sec  = struct.unpack('i', struct.pack('f', tv_sec))[0]
         tv_nsec = struct.unpack('i', struct.pack('f', tv_nsec))[0]
 
         if ql.arch.bits == 32:
-            return StatxTimestamp32(tv_sec=tv_sec, tv_nsec=tv_nsec)
+            Timestamp = StatxTimestamp32EB if is_eb else StatxTimestamp32
         else:
-            return StatxTimestamp64(tv_sec=tv_sec, tv_nsec=tv_nsec)
+            Timestamp = StatxTimestamp64EB if is_eb else StatxTimestamp64
+
+        return Timestamp(tv_sec=tv_sec, tv_nsec=tv_nsec)
 
 
     def major(dev):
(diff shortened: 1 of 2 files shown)


xwings:
Hi,

Will you be able to fix the conflict ?

retrocpugeek:
Rebased onto `dev`, conflict resolved. CI is green.

xwings:
thanks!

PR #1425, opened 2023-12-15
Bugfix: unaligned access in MIPS getdents64

ljucovic:
### Which kind of PR do you create?

- [X] This PR only contains minor fixes.
- [ ] This PR contains major feature update.
- [ ] This PR introduces a new function/api for Qiling Framework.

### Coding convention?

- [X] The new code conforms to Qiling Framework naming convention.
- [ ] The imports are arranged properly.
- [X] Essential comments are added.
- [ ] The reference of the new code is pointed out.

### Extra tests?

- [X] No extra tests are needed for this PR.
- [ ] I have added enough tests for this PR.
- [ ] Tests will be added after some discussion and review.

### Changelog?

- [X] This PR doesn't need to update Changelog.
- [ ] Changelog will be updated after some proper review.
- [ ] Changelog has been updated in my PR.

### Target branch?

- [X] The target branch is dev branch.

### One last thing

- [X] I have read the [contribution guide](https://docs.qiling.io/en/latest/contribution/)

-----

PR #1423, opened 2023-12-14
Revert "Bugfix: unaligned access in MIPS getdents64"

xwings:
Reverts qilingframework/qiling#1417

seems like there is a issue

PR #1419, opened 2023-12-13
Bugfix: unaligned access in MIPS getdents64

tats-gd:
### Which kind of PR do you create?

- [X] This PR only contains minor fixes.
- [ ] This PR contains major feature update.
- [ ] This PR introduces a new function/api for Qiling Framework.

### Coding convention?

- [X] The new code conforms to Qiling Framework naming convention.
- [ ] The imports are arranged properly.
- [X] Essential comments are added.
- [ ] The reference of the new code is pointed out.

### Extra tests?

- [X] No extra tests are needed for this PR.
- [ ] I have added enough tests for this PR.
- [ ] Tests will be added after some discussion and review.

### Changelog?

- [X] This PR doesn't need to update Changelog.
- [ ] Changelog will be updated after some proper review.
- [ ] Changelog has been updated in my PR.

### Target branch?

- [ ] The target branch is dev branch.

### One last thing

- [X] I have read the [contribution guide](https://docs.qiling.io/en/latest/contribution/)

-----

PR #1636, opened 2026-06-24
Align getdents and getdents64 records to fix unaligned access on MIPS

### LINKED ISSUE #1635: getdents / getdents64 return unaligned dirent records (unaligned-load fault on MIPS)
# `getdents` / `getdents64` return unaligned dirent records (unaligned-load fault on MIPS)

## Describe the bug

`__getdents_common` (`qiling/os/posix/syscall/unistd.py`) computes each record length as

```python
d_reclen = n + n + 2 + len(d_name) + 1
```

without rounding it up. The kernel rounds each record up to the alignment of its leading `d_ino`, so the *next* record's `d_ino` stays aligned. Without that, records pack tightly and the next `d_ino` lands on a misaligned address. A strict-alignment guest (e.g. MIPS / MIPS64) then faults with an unaligned load (`UC_ERR_*_UNALIGNED`) while walking the buffer — e.g. `busybox ls` crashes right after the syscall. x86 tolerates the unaligned read, which is why it's invisible there.

This affects **both** dirent syscalls:

- **`getdents64`** — leading `u64 d_ino` (needs 8-byte alignment on every arch).
- **legacy `getdents`** — word-sized `d_ino`: 4 bytes on a 32-bit guest, **8 bytes on a 64-bit guest** (e.g. MIPS64 n64). Older glibc still issues this syscall, so MIPS64 directory listings fault here too.

## Repro (Python) — `getdents64`

```python
from qiling import Qiling
from qiling.const import QL_ARCH, QL_OS, QL_ENDIAN, QL_VERBOSE
from qiling.os.posix.syscall.fcntl import ql_syscall_open
from qiling.os.posix.syscall.unistd import ql_syscall_getdents64

ql = Qiling(code=b"\x00\x00\x00\x00", archtype=QL_ARCH.MIPS, ostype=QL_OS.LINUX,
            endian=QL_ENDIAN.EB, rootfs="examples/rootfs/mips32_linux", verbose=QL_VERBOSE.OFF)

base = 0x100000
ql.mem.map(base, 0x4000)
ql.mem.write(base, b"/\x00")
fd = ql_syscall_open(ql, base, 0, 0)
n = ql_syscall_getdents64(ql, fd, base + 0x100, 0x2000)

buf = bytes(ql.mem.read(base + 0x100, n))
off = 0
while off < n:
    print("record at offset", off, "aligned" if off % 8 == 0 else "*** MISALIGNED ***")
    off += int.from_bytes(buf[off + 16:off + 18], "big")  # d_reclen (u16 @ 16)
```

Output (without the fix):

```
record at offset 0 aligned
record at offset 21 *** MISALIGNED ***      # ".." starts at 21 -> its u64 d_ino is unaligned
record at offset 43 *** MISALIGNED ***
...
```

The legacy `getdents` path has the identical defect (swap `ql_syscall_getdents64` → `ql_syscall_getdents`; `d_reclen` is a u16 at offset `2 * pointersize`). On MIPS the guest's load of the misaligned `d_ino` raises `UC_ERR_*_UNALIGNED`.

## Expected behavior

Each record should be padded so the next record's leading `d_ino` is aligned, matching the kernel's `ALIGN(reclen, sizeof(long))`. All record offsets land on aligned boundaries and a strict-alignment guest can walk the buffer.

## Why it was never caught

x86 tolerates unaligned loads. The existing getdents test binaries (`test_elf_linux_x8664_getdents`, `test_elf_linux_x86_getdents64`) drive the **legacy** `getdents` path but only assert directory *names*, never the record alignment / `d_type`, so the defect is invisible in CI.

## Prior art (please read before reviewing)

This bug was fixed once and reverted:

- #1419 merged `d_reclen = (d_reclen + n) & ~(n - 1)`, then #1423 reverted it ("seems like there is an issue"). #1425 re-submits the **same** formula and is still open.

That blanket approach has two problems the fix here avoids:

1. `(d_reclen + n)` over-pads; the canonical round-up is `(d_reclen + (n - 1)) & ~(n - 1)`.
2. It rounded the **legacy `getdents`** record unconditionally **without relocating `d_type`**. The legacy `linux_dirent` stores `d_type` in the record's *last byte* (offset `d_reclen - 1`); padding the record moved the consumer's `d_type` read into the pad bytes — almost certainly the cause of the revert. (`getdents64` is safe because it puts `d_type` *before* `d_name`, so trailing pad is inert.)

A fix for both layouts is proposed in PR #1636 (supersedes #1425).

retrocpugeek:
## What

`__getdents_common` did not round dirent records up to the alignment of their leading `d_ino`, so the next record's `d_ino` could land unaligned. A strict-alignment guest (MIPS/MIPS64) then faults with `UC_ERR_*_UNALIGNED` while walking the buffer (e.g. `busybox ls` crashes right after the syscall). x86 tolerates the unaligned access, which is why it was never caught.

This affects **both** `getdents64` (`u64 d_ino`) and the legacy `getdents` (word-sized `d_ino` — 8 bytes on a 64-bit guest such as MIPS64 n64; older glibc still issues this syscall).

Closes #1635.

## How

Round each record up to the alignment of `d_ino`, matching the kernel's `ALIGN(reclen, sizeof(long))`:

```python
unaligned_reclen = n + n + 2 + len(d_name) + 1
d_reclen = (unaligned_reclen + (n - 1)) & ~(n - 1)
```

applied to **both** record layouts. The one subtlety is `d_type`:

- `getdents64` places `d_type` *before* `d_name`, so the trailing pad bytes are inert.
- legacy `getdents` stores `d_type` in the record's **last byte** (offset `d_reclen - 1`), so the alignment padding is inserted **between `d_name` and `d_type`** to keep `d_type` there.

### Relationship to #1419 / #1423 / #1425

This bug was fixed once (#1419) and reverted (#1423); #1425 re-submits `(d_reclen + n) & ~(n - 1)` and is still open. That blanket approach had two issues this PR fixes properly:

1. `(d_reclen + n)` over-pads — the canonical round-up is `(d_reclen + (n - 1))`.
2. It rounded the legacy record unconditionally **without relocating `d_type`**, so the consumer's `d_type` read (at `d_reclen - 1`) landed in the pad bytes — almost certainly the revert cause. This PR moves `d_type` to the new `d_reclen - 1`, so the legacy layout is correct after padding.

This supersedes #1425.

## Tests

- `test_elf.ELFTest.test_linux_getdents64_alignment` — drives `ql_syscall_getdents64` on a directory and asserts every `linux_dirent64` record starts 8-byte aligned.
- `test_elf.ELFTest.test_linux_getdents_alignment` — drives the legacy `ql_syscall_getdents` and asserts every record is aligned, tiles the buffer, and still carries `d_type` (`DT_DIR` for `.`/`..`) in its last byte.

Both are self-contained (no rootfs binary, run on stock unicorn); each fails on `dev` and passes with this fix.

## Checklist

- [X] I have added enough tests for this PR.

### DIFF
--- a/qiling/os/posix/syscall/unistd.py
+++ b/qiling/os/posix/syscall/unistd.py
@@ -926,7 +926,15 @@ def __getdents_common(ql: Qiling, fd: int, dirp: int, count: int, *, is_64: bool
             d_off = 0
             d_name = (result.name if isinstance(result, os.DirEntry) else result._str).encode() + b'\x00'
             d_type = _type_mapping(result)
-            d_reclen = n + n + 2 + len(d_name) + 1
+
+            # The kernel rounds each record up to the alignment of its leading
+            # (pointer-sized) d_ino, so the next record's d_ino stays aligned.
+            # Without this, strict-alignment guests (e.g. MIPS) fault with an
+            # unaligned load when walking the buffer; x86 tolerates it, which is
+            # why it was never caught. Matches the kernel's ALIGN(reclen,
+            # sizeof(long)) for both getdents and getdents64.
+            unaligned_reclen = n + n + 2 + len(d_name) + 1
+            d_reclen = (unaligned_reclen + (n - 1)) & ~(n - 1)
 
             # TODO: Dirty fix for X8664 MACOS 11.6 APFS
             # For some reason MACOS return int value is 64bit
@@ -936,18 +944,8 @@ def __getdents_common(ql: Qiling, fd: int, dirp: int, count: int, *, is_64: bool
                 packed_d_ino = (ql.pack64(d_ino), n)
 
             if is_64:
-                # The kernel rounds each linux_dirent64 record up to the
-                # alignment of its leading u64 d_ino, so the next record's d_ino
-                # stays aligned. Without this, strict-alignment guests (e.g.
-                # MIPS) fault with an unaligned load when walking the buffer;
-                # x86 tolerates it, which is why it was never caught. This is
-                # safe here because getdents64 places d_type *before* d_name, so
-                # the trailing pad bytes are inert. The legacy getdents layout
-                # below stores d_type at offset d_reclen-1, so it can't be
-                # padded the same way without relocating d_type -- that mistake
-                # is what got the earlier blanket fix (PR #1419) reverted.
-                d_reclen = (d_reclen + (n - 1)) & ~(n - 1)
-
+                # getdents64 places d_type *before* d_name, so the trailing pad
+                # bytes are inert.
                 fields = (
                     (ql.pack64(d_ino), n),
                     (ql.pack64(d_off), n),
@@ -956,11 +954,18 @@ def __getdents_common(ql: Qiling, fd: int, dirp: int, count: int, *, is_64: bool
                     (d_name, len(d_name))
                 )
             else:
+                # legacy linux_dirent stores d_type in the record's last byte
+                # (offset d_reclen-1), so the alignment padding goes *between*
+                # d_name and d_type to keep d_type there. padding without
+                # relocating d_type is what got the earlier blanket fix
+                # (PR #1419) reverted.
+                pad = d_reclen - unaligned_reclen
                 fields = (
                     packed_d_ino,
                     (ql.pack(d_off), n),
                     (ql.pack16(d_reclen), 2),
                     (d_name, len(d_name)),
+                    (b'\x00' * pad, pad),
                     (d_type, 1)
                 )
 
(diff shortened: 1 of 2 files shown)


ljucovic:
Thanks for the careful write-up and for picking this up. You're right on both points, (d_reclen + n) over-pads, and not relocating d_type in the legacy record is exactly what broke #1419 after merge; I never fully isolated that at the time, which is why #1425 carried the same formula forward.
Happy to close #1425 in favor of #1636. Your version handles both layouts correctly and has the regression tests this fix has needed from the start. Nice work.

xwings:
Agree with @ljucovic , thanks @retrocpugeek

PR #1639, opened 2026-06-25
Implement the clone3 syscall

### LINKED ISSUE #1638: clone3 is not implemented — thread creation fails on modern-glibc binaries
# `clone3` is not implemented — thread creation fails on modern-glibc binaries

## Describe the bug

`clone3` (the modern variant of `clone`, syscall 435 on most arches, 5435 on MIPS n64) has **no handler** in Qiling. It appears in the number→name maps (`qiling/os/linux/map_syscall.py`) but there is no `ql_syscall_clone3`:

```
$ grep -rn "def ql_syscall_clone" qiling/
qiling/os/posix/syscall/sched.py:13:def ql_syscall_clone(...)   # only the legacy clone
```

Recent glibc (e.g. the toolchains shipping with Ubuntu 24.04 / glibc 2.39) issues **`clone3`** from `pthread_create`, falling back to the legacy `clone` only if `clone3` returns `-ENOSYS`. So on any guest built against a recent glibc, `pthread_create` hits the unimplemented `clone3` and thread creation fails:

```
[!] 0x...: syscall ql_syscall_clone3 number = 0x153b(5435) not implemented
Create pthread error!
```

Crucially, the `clone3`→`clone` fallback never kicks in, because **Qiling does not return `-ENOSYS` for unimplemented syscalls** — `QlOsPosix` just logs a warning and leaves the return register untouched (`qiling/os/posix/posix.py`):

```python
self.ql.log.warning(f'... syscall {syscall_name} ... not implemented')
```

glibc only falls back on a literal `-ENOSYS`, so it sees a garbage return value, assumes `clone3` "worked" (or failed for a real reason), and gives up instead of retrying with `clone`.

## Repro

Any multithreaded binary built with a recent glibc reproduces it. With the legacy-`clone` test binaries currently in the rootfs (older glibc) the path is not exercised, which is why CI doesn't catch it. The symptom for a modern-glibc pthread binary is the warning above followed by `Create pthread error!` (or, if creation is allowed to proceed with a bogus return, a crash during thread setup).

## Expected behavior

`clone3` should create a thread just like `clone` does, so `pthread_create` works on guests built with a modern glibc.

## Proposed fix

Add `ql_syscall_clone3` that unpacks `struct clone_args` and delegates to the existing `ql_syscall_clone`. The non-obvious translations:

- **`child_stack = stack + stack_size`** — `clone3` passes the stack *base* plus a separate size; legacy `clone` wants the *highest* stack address.
- **`exit_signal`** is its own `struct clone_args` field in `clone3`; legacy `clone` packs it into the low `CSIGNAL` byte of `flags`.
- **x8664 pre-swap** — `ql_syscall_clone` swaps `newtls`<->`child_tidptr` to undo x8664's raw-syscall register order; since `clone3` hands over already-logical args, pre-swap on x8664 so it cancels out.

Implemented on `retrocpugeek:fix/clone3-syscall` (off `dev`):
- Fix: `Implement the clone3 syscall`
- Regression test: `test_elf_multithread.ELFTest.test_clone3_translates_to_clone` — drives `ql_syscall_clone3` directly and asserts the translation for both the generic path and the x8664 swap. Self-contained, runs on stock unicorn (no `clone3` binary needed).

Verified end-to-end separately: a real MIPS64 BE glibc pthread binary now creates and joins threads correctly with this handler in place.

Happy to open a PR against `dev`.

retrocpugeek:
## Summary

Implements the `clone3` syscall, which currently has no handler in Qiling — it only appears in the number→name maps. Recent glibc issues `clone3` (not the legacy `clone`) from `pthread_create`, so **thread creation fails on any guest built against a modern glibc**. And because Qiling doesn't return `-ENOSYS` for unimplemented syscalls, glibc's built-in `clone3`→`clone` fallback never fires.

Closes #1638.

## What it does

Adds `ql_syscall_clone3`, which unpacks `struct clone_args` and delegates to the existing `ql_syscall_clone`. Non-obvious translations:

- `child_stack = stack + stack_size` — `clone3` passes the stack *base* plus a separate size; legacy `clone` wants the *highest* stack address.
- `exit_signal` is its own `struct clone_args` field; legacy `clone` packs it into the low `CSIGNAL` byte of `flags`.
- x8664 pre-swap — `ql_syscall_clone` swaps `newtls`<->`child_tidptr` to undo x8664's raw-syscall register order; since `clone3` hands over already-logical args, we pre-swap on x8664 so it cancels out.

The legacy `clone` path is untouched.

## Testing

- New self-contained regression test `test_elf_multithread.ELFTest.test_clone3_translates_to_clone` drives `ql_syscall_clone3` directly and asserts the translated arguments for both the generic path and the x8664 swap. **Runs on stock unicorn** — no `clone3` binary required.
- Full `test_elf_multithread.py` suite passes on stock unicorn 2.1.3 (24 tests, 1 pre-existing skip).
- Verified end-to-end separately with a real MIPS64 BE glibc pthread binary: threads are created and joined correctly with this handler in place.

### DIFF
--- a/qiling/os/posix/syscall/sched.py
+++ b/qiling/os/posix/syscall/sched.py
@@ -130,6 +130,38 @@ def ql_syscall_clone(ql: Qiling, flags: int, child_stack: int, parent_tidptr: in
 
     return regreturn
 
+def ql_syscall_clone3(ql: Qiling, cl_args: int, size: int):
+    # clone3(struct clone_args *uargs, size_t size). Translate the struct into
+    # the legacy clone() argument set and delegate. Modern glibc (used by recent
+    # toolchains) issues clone3 from pthread_create; without this it falls through
+    # to the "not implemented" path and thread creation fails.
+    #
+    # struct clone_args (all fields __aligned_u64):
+    #   0:flags 8:pidfd 16:child_tid 24:parent_tid 32:exit_signal
+    #   40:stack 48:stack_size 56:tls 64:set_tid 72:set_tid_size 80:cgroup
+    flags       = ql.mem.read_ptr(cl_args + 0, 8)
+    child_tid   = ql.mem.read_ptr(cl_args + 16, 8)
+    parent_tid  = ql.mem.read_ptr(cl_args + 24, 8)
+    exit_signal = ql.mem.read_ptr(cl_args + 32, 8)
+    stack       = ql.mem.read_ptr(cl_args + 40, 8)
+    stack_size  = ql.mem.read_ptr(cl_args + 48, 8)
+    tls         = ql.mem.read_ptr(cl_args + 56, 8)
+
+    # legacy clone() takes the highest stack address; clone3 gives the base + size
+    child_stack = (stack + stack_size) if stack else 0
+
+    # clone3 keeps exit_signal in its own field; legacy clone packs it into the
+    # low CSIGNAL byte of flags
+    flags |= exit_signal & 0xff
+
+    # ql_syscall_clone swaps newtls<->child_tidptr for x8664 to undo that arch's
+    # raw-syscall register order. clone3 hands us already-logical args, so on
+    # x8664 we pre-swap to cancel that out.
+    if ql.arch.type == QL_ARCH.X8664:
+        return ql_syscall_clone(ql, flags, child_stack, parent_tid, child_tid, tls)
+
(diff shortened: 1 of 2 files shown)


w1tcher (approved):
LGTM

xwings (approved):

xwings:
Thanks!

PR #1643, opened 2026-07-02
Fix GetProcAddress/LdrGetProcedureAddress export resolution

H1ghC0ntr4sT:
## Summary

`GetProcAddress` and `LdrGetProcedureAddress` resolved a non-self module's function
only through `ql.loader.import_address_table[dll_name]`. That lookup fails for
functions the target DLL exports natively but that never made it into the table,
and `LdrGetProcedureAddress` additionally never encoded the looked-up name to
`bytes`, so a name-based lookup could never match a table key.

This adds a shared `resolve_export()` helper in `qiling/os/windows/utils.py` used by
both hooks:

1. Consult the existing `import_address_table` first, so every already-working
   resolution stays unchanged.
2. On a miss, parse the target DLL's real export directory with `pefile`, following
   forwarder exports (e.g. `kernel32.HeapAlloc` -> `ntdll.RtlAllocateHeap`) to the
   DLL that implements the function.

The name key is normalised to `bytes`, fixing the `LdrGetProcedureAddress` path.

## Verification

Verified on Windows against the real system DLLs, using an MSVC-built test binary
that resolves both natively-exported and forwarded functions.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

### DIFF
--- a/qiling/os/windows/utils.py
+++ b/qiling/os/windows/utils.py
@@ -3,12 +3,17 @@
 # Cross Platform and Multi Architecture Advanced Binary Emulation Framework
 #
 
-from typing import Iterable, Optional, Tuple, TypeVar
+import os
+
+from typing import Iterable, Optional, Tuple, TypeVar, Union
+
+import pefile
 
 from unicorn import UcError
 
 from qiling import Qiling
 from qiling.exception import QlErrorSyscallError
+from qiling.loader.loader import Image
 from qiling.os.const import POINTER
 from qiling.os.windows.fncc import STDCALL
 from qiling.os.windows.wdk_const import *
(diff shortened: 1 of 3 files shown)


qiling/os/windows/utils.py
```diff
@@ -29,6 +34,100 @@ def has_lib_ext(name: str) -> bool:
...
+def __export_from_dll_file(ql: Qiling, image: Image, name: Optional[bytes], ordinal: Optional[int], seen: set) -> Optional[int]:
+    """Resolve an export by parsing the target DLL's export directory from disk.
+    Forwarder exports (e.g. kernel32.HeapAlloc -> ntdll.RtlAllocateHeap) are
+    followed to the DLL that actually implements the function.
+    """
+
+    try:
+        pe = pefile.PE(image.path, fast_load=True)
+        pe.parse_data_directories(directories=[pefile.DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_EXPORT']])
+    except (pefile.PEFormatError, OSError):
```
xwings: One issue must be resolved before this can merge; it has its own section below.

```
  43 |     try:
  44 |         pe = pefile.PE(image.path, fast_load=True)
  45 |         pe.parse_data_directories(directories=[pefile.DIRECTORY_ENTRY['IMAGE_DIRECTORY_ENTRY_EXPORT']])
> 46 |     except (pefile.PEFormatError, OSError):
  47 |         # a malformed pe or a file that vanished after being mapped -> treat as not found
  48 |         return None
  49 |
```

**Problem.** Catching PEFormatError and every OSError without preserving the exception makes a failure to inspect an already-loaded DLL indistinguishable from an absent export. For example, a permission failure becomes STATUS_PROCEDURE_NOT_FOUND in LdrGetProcedureAddress, with no diagnostic explaining the actual cause. Existing loader paths log unavailable DLL files and unresolved forwarders.

**How to change it.** Keep the guest-visible failure return, but log the image path and caught exception before returning None; ordinary missing-export results can remain silent.

Suggested change — a sketch to adapt, not a patch to apply:

```python
except (pefile.PEFormatError, OSError) as err:
    ql.log.warning(f'Could not read exports from {image.path!r}: {err}')
    return None
```

The name normalization addresses the confirmed mismatch between LdrGetProcedureAddress's string input and the loader's byte-keyed export inventory. Sharing resolution between the two hooks also keeps their lookup behavior together without introducing another dependency.

One actionable issue remains below. The checklist separately records the uncertainty about the broader disk fallback's incremental benefit; that uncertainty is not a conclusion that the fallback is unnecessary.

PR #1644, opened 2026-07-02
Reset emu state after a failing DllMain

H1ghC0ntr4sT:
## Summary

When a DLL's `DllMain` raises a `UcError`, `emu_start()` propagates the exception
without clearing `ql._state` (it only sets `STOPPED` on the success path), so the
state is left at `STARTED`. `call_dll_entrypoint()` (`qiling/loader/pe.py`) catches
the error and restores registers but never resets the state. Because DLL loading
gates on `emu_state is not STARTED` before running any later DLL's `DllMain`, one
early failure silently disables `DllMain` for every DLL loaded afterwards — with no
log output.

This resets `ql._state` to `STOPPED` on the exception path, mirroring what
`emu_start()` does on success.

## Verification

Verified on Windows: with a `DllMain` that raises, DLLs loaded afterwards now have
their `DllMain` executed instead of being silently skipped.

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

### DIFF
--- a/qiling/loader/pe.py
+++ b/qiling/loader/pe.py
@@ -462,6 +462,10 @@ class Process:
             self.ql.log.error(f'Error encountered while running {dll_name} DllMain, bailing')
 
             self.ql.arch.regs.restore(regs_state)
+
+            # emu_start leaves the state at STARTED when it raises; reset it so a single
+            # failing DllMain does not gate every subsequently loaded DLL's entry point.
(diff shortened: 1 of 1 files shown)


w1tcher (approved):
LGTM

xwings (approved):

xwings:
Hi, 

Welcome to Qiling Framework.

Thanks for the PR.

Issue #1377, opened 2023-08-30
Remote GDB debugging X8664_Linux occurs warning: Invalid remote reply:


Leshyan:
***Describe the bug**
When qiling runs the x8664_linux program and enables default GDB remote debugging, two n/si instructions issued by GDB will trigger warning: Invalid remote reply, causing debugging to fail. But the breakpoint-based c instruction is normal.

**Sample Code**
```python
import sys
from qiling import *
from qiling.extensions import pipe
from qiling.const import QL_VERBOSE

def test_gdb(path, rootfs):
    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.DEBUG,log_file="./output")
    ql.debugger = True
    ql.run()  

if __name__ == "__main__":
    test_gdb(["../tool/qiling/examples/rootfs/x8664_linux/bin/pwn1"], "../tool/qiling/examples/rootfs/x8664_linux")

```

**Expected behavior**
Remote gdb can debug normally without restriction.

**Screenshots**
![CleanShot 2023-08-30 at 22 05 42](https://github.com/qilingframework/qiling/assets/89032676/fd5f06d2-ce77-4a23-a805-852c9bcf1acf)

**Additional context**
Part of the log. 
```
[+] 	getpkt ("qSupported:multiprocess+;swbreak+;hwbreak+;qRelocInsn+;fork-events+;vfork-events+;exec-events+;vContSupported+;QThreadEvents+;no-resumed+;memory-tagging+;xmlRegisters=i386");
[+] 	putpkt ("+");
[+] 	[sent ack]
[...]

Issue #1310, opened 2023-02-06
`Qiling.emu_stop()` does not stop the emulator when gdb debugger is attached


bstee615:
**Describe the bug**
Thank you for your work on this useful framework.

When executing this sample code, Qiling waits for gdb to connect. I attached gdb using `target remote 127.0.0.1:9999`, then used the `stepi` command repeatedly until the executable reached the entry of `foo`. When I use `stepi` on the entry to `foo`, then my script's function `print_asm` calls `ql.emu_stop()`, which should stop the emulator, but the emulator does not stop - it remains on the same instruction. Because of this, if instead I use the `step` command to step to the next statement, then gdb hang in an infinite loop.

**Sample Code**
Qiling script:
```python
from qiling import *
from qiling.const import QL_VERBOSE

def test_gdb(path, rootfs):
    ql = Qiling(path, rootfs, verbose=QL_VERBOSE.OFF)

    # Enable debugger to listen at localhost address, default port 9999
    ql.debugger = True

    ba = ql.loader.images[0].base

    foo_addr = int("0000000000001135", 16)  # start address of foo
    foo_len = int("0000000000000024", 16)  # length of foo

    def print_asm(ql, address, size):
        print("INSTRUCTION:", hex(address), size)
        if address == ba + foo_addr:
            print("REACHED TARGET INSTRUCTION", hex(ba+foo_addr), "END EMU")
            ql.emu_stop()

    ql.hook_code(print_asm, begin=ba + foo_addr, end=ba + foo_addr + foo_len)
    ql.run()

if __name__ == "__main__":
    test_gdb(["./test"], "/")
```

Log of qiling script output: [test_gdb.log](https://github.com/qilingframework/qiling/files/10669264/test_gdb.log)

**Expected behavior**
When the `ql.emu_stop()` line is executed, emulation will quit and gdb will detach.

**Additional context**
I ran this test case on the official Docker container `qilingframework/qiling:latest` with the latest code from the `dev` branch.

It should be possible to reproduce the bug with any executable by calling `Qiling.emu_stop()` during execution.
Here is the target program I used:
```c
#include <stdio.h>

void foo(int a) {
    printf("Hello, world %d!\n", a);
}

int main(int argc, char **argv)
{
    int a = 10;
    foo(a);
    return a;
}
```

Compiled executable (unzip it first): [test.zip](https://github.com/qilingframework/qiling/files/10669312/test.zip)

elicn:
Thanks for filing this.

[...]

PR #1653, opened 2026-08-07
gdb: report SIGTRAP (not SIGTERM) on single-step

### LINKED ISSUE #1538: Qiling gdbserver not handling stepi instruction
***Describe the bug**
When remote debugging a program running in Qiling, the Qiling gdbserver seems not to handle the single step instruction from gdb client. Every time I perform the si command in my gdb client, Qiling gdbserver send  SIGTERM causing my debugged program to terminate.

**Sample Code**
```python
 ql = Qiling(["./toto.elf"],
                archtype=QL_ARCH.CORTEX_M, ostype=QL_OS.MCU, env=stm32f103, verbose=QL_VERBOSE.DISASM)

    ql.hw.create('scb')
    ql.hw.create('gpioa')
    ql.hw.create('usart1')
    ql.hw.create('rcc')
    ql.hw.create('afio')
    ql.hw.create('exti')
    ql.debugger = True
    ql.hw.show_info()

    ql.run(count=1000000)

    print(ql.hw.usart1.recv())

```

**Expected behavior**
Qiling gdbserver not sending SIGTERM each time performing the si command when remote debugging.

**Dig in gdbserver Qiling implementation**
I digged in the gdbserver implementation and debug with the command printing offered by Qiling gdbserver and this function seems weird to me, file :  qiling/debugger/gdb/gdb.py

```python
def handle_s(subcmd: str) -> Reply:
  """Perform a single step.
  """
  
  self.gdb.resume_emu(steps=1)
  
  # if emulation has been stopped, signal program termination
# ===== these two lines seems weird to me because the condition is always True ===
# This is what send the SIGTERM all the time 
 if self.ql.emu_state is QL_STATE.STOPPED: 
   return f'S{SIGTERM:02x}'
# ====================================================================
  # otherwise, this is just single stepping
  return f'S{SIGTRAP:02x}'
```
Wonder if it's normal or it's a bug or this condition is useful in a particular case that is not related to my implementation. I've tested to comment these two lines and everything works perfectly fine after that. I think it also causes other bugs with breakpoint but not so sure and didn't have the time to verify this point.

**Additional context**
I worked with the dev branch.
Just a  really simple code, I work on cortex-m3 and this code works perfectly fine without debugging.
I use arm-none-eabi-gdb as gdb client (from the toolchain used to cross compile toto.elf) with my program and it has symbols debug.
I know that mcu's haven't been really tested with gdb and Qiling but I just crafted two xml files for cortex-m placed at : qiling/debugger/gdb/xml/cortex-m.

arm-m-profile.xml : 
```xml 
<?xml version="1.0"?>
<!-- Copyright (C) 2010-2020 Free Software Foundation, Inc.

     Copying and distribution of this file, with or without modification,
     are permitted in any medium without royalty provided the copyright
     notice and this notice are preserved.  -->

<!DOCTYPE feature SYSTEM "gdb-target.dtd">
<feature name="org.gnu.gdb.arm.m-profile">
  <reg name="r0" bitsize="32" regnum="0"/>
  <reg name="r1" bitsize="32"/>
  <reg name="r2" bitsize="32"/>
  <reg name="r3" bitsize="32"/>
  <reg name="r4" bitsize="32"/>
  <reg name="r5" bitsize="32"/>
  <reg name="r6" bitsize="32"/>
  <reg name="r7" bitsize="32"/>
  <reg name="r8" bitsize="32"/>
  <reg name="r9" bitsize="32"/>
  <reg name="r10" bitsize="32"/>
  <reg name="r11" bitsize="32"/>
  <reg name="r12" bitsize="32"/>
  <reg name="sp" bitsize="32" type="data_ptr"/>
  <reg name="lr" bitsize="32"/>
  <reg name="pc" bitsize="32" type="code_ptr"/>
  <reg name="xpsr" bitsize="32" regnum="25"/>
</feature>
```

target.xml : 
```xml 
<?xml version="1.0"?>
<!-- Copyright (C) 2009-2016 Free Software Foundation, Inc.

 *!Copying and distribution of this file, with or without modification,
 *!are permitted in any medium without royalty provided the copyright
 *!notice and this notice are preserved.  -->

<!DOCTYPE target SYSTEM "gdb-target.dtd">
<target xmlns:xi="http://www.w3.org/2001/XInclude">
    <architecture>arm</architecture>
    <xi:include href="arm-m-profile.xml"/>
</target>
```

retrocpugeek:
## Summary

Single-stepping over Qiling's gdb stub disconnects the client: after the first `s` (step-instruction) the stub sends a **SIGTERM** stop-reply, so the client believes the program died. This is reported in **#1538** and is the same root cause behind **#1377**.

`handle_s` decided the reply from `ql.emu_state`:

```python
self.gdb.resume_emu(steps=1)
if self.ql.emu_state is QL_STATE.STOPPED:   # always true after a step
    return f'S{SIGTERM:02x}'
return f'S{SIGTRAP:02x}'
```

But `Qiling.emu_start` **unconditionally** sets the state to `QL_STATE.STOPPED` once it has run the requested step count, so the guard is true on *every* step — the `SIGTRAP` branch is dead code. (Diagnosed in detail by @cbdm in #1377.)

## Fix

Give `handle_s` the same exit-vs-trap discrimination `handle_c` already uses: a step reports `SIGTRAP` unless it carried `pc` all the way to the emulation exit point, in which case the guest really has terminated and we reply `W{exit_code}`.

While here:
- wrap the step in the same `UcError` / `KeyboardInterrupt` handling as `handle_c`, so a fault *while stepping* maps to a signal (`SIGSEGV`/`SIGBUS`/`SIGILL`) instead of crashing the stub with an unhandled exception;
- hoist the duplicated uc-error→signal map to a module-level `UC_ERROR_SIGMAP` shared by `handle_c` and `handle_s`.

This supersedes the common "just comment out the two lines" workaround, which leaves the client unable to detect real termination ("bad exit if you step too much").

## Test

Adds `test_gdbdebug_stepi_reports_sigtrap`, which single-steps over the stub and asserts the stop-reply is `S05` (SIGTRAP). Against the unpatched stub it fails with `S0f` (SIGTERM); with the fix it passes. The existing debugger tests only *sent* packets and never read replies, so this gap went uncaught — the test adds a small reply-reading client.

Verified locally: full `test_debugger.py` suite passes (5 tests), and an end-to-end check with `gdb-multiarch` against a dynamically-linked AArch64 busybox now single-steps cleanly (no `SIGTERM, Terminated`) and reports a clean exit on `continue`.

Fixes #1538

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/qiling/debugger/gdb/gdb.py
+++ b/qiling/debugger/gdb/gdb.py
@@ -30,7 +30,7 @@ from unicorn.unicorn_const import (
 )
 
 from qiling import Qiling
-from qiling.const import QL_ARCH, QL_ENDIAN, QL_OS, QL_STATE
+from qiling.const import QL_ARCH, QL_ENDIAN, QL_OS
 from qiling.debugger import QlDebugger
 from qiling.debugger.gdb.xmlregs import QlGdbFeatures
 from qiling.debugger.gdb.utils import QlGdbUtils
@@ -55,6 +55,21 @@ SIGTERM = 15
 SIGCHLD = 16
 SIGCONT = 17
 SIGSTOP = 18
+
+# translate a unicorn cpu fault into the closest posix signal, shared by the
+# continue ('c') and single-step ('s') handlers when emulation faults
+UC_ERROR_SIGMAP = {
+    UC_ERR_READ_UNMAPPED   : SIGSEGV,
+    UC_ERR_WRITE_UNMAPPED  : SIGSEGV,
+    UC_ERR_FETCH_UNMAPPED  : SIGSEGV,
+    UC_ERR_WRITE_PROT      : SIGSEGV,
+    UC_ERR_READ_PROT       : SIGSEGV,
+    UC_ERR_FETCH_PROT      : SIGSEGV,
+    UC_ERR_READ_UNALIGNED  : SIGBUS,
+    UC_ERR_WRITE_UNALIGNED : SIGBUS,
+    UC_ERR_FETCH_UNALIGNED : SIGBUS,
+    UC_ERR_INSN_INVALID    : SIGILL
+}
 
 # common replies
 REPLY_ACK = b'+'
@@ -226,21 +241,8 @@ class QlGdb(QlDebugger):
             try:
                 self.gdb.resume_emu()
             except UcError as err:
-                sigmap = {
-                    UC_ERR_READ_UNMAPPED   : SIGSEGV,
-                    UC_ERR_WRITE_UNMAPPED  : SIGSEGV,
-                    UC_ERR_FETCH_UNMAPPED  : SIGSEGV,
-                    UC_ERR_WRITE_PROT      : SIGSEGV,
-                    UC_ERR_READ_PROT       : SIGSEGV,
-                    UC_ERR_FETCH_PROT      : SIGSEGV,
-                    UC_ERR_READ_UNALIGNED  : SIGBUS,
-                    UC_ERR_WRITE_UNALIGNED : SIGBUS,
-                    UC_ERR_FETCH_UNALIGNED : SIGBUS,
-                    UC_ERR_INSN_INVALID    : SIGILL
-                }
-
                 # determine signal from uc error; default to SIGTERM
-                reply = f'S{sigmap.get(err.errno, SIGTERM):02x}'
+                reply = f'S{UC_ERROR_SIGMAP.get(err.errno, SIGTERM):02x}'
 
             except KeyboardInterrupt:
                 # emulation was interrupted with ctrl+c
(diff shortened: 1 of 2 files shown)


retrocpugeek:
Superseded by #1654, which carries this commit unchanged plus two further fixes in the same stop-reply path: `vCont;C<sig>`/`S<sig>` actions with any signal (the packet #1377 actually fails on), and async-interrupt (ctrl-c) support so a free-running guest can be broken into. It also documents how this change relates to #1310, whose fix (#1322) introduced the `emu_state` check removed here.

Closing in favour of #1654.

PR #1663, opened 2026-09-10
Add AArch64 code-range blob example with UDF trapping

warsang:
Adds a small examples/hello_arm64_range_blob.py example. Used QL_ARCH.ARM64

(Did that to emulate the ARM stub of an ARM64EC PE file).

Tested locally on a small ARM binary blob:

| blob | expected | observed |
|---|---|---|
| 8x NOP | clean return | `RETURNED, steps=8` |
| `UDF #0` | single UDF fault | `UDF pc=0x20000000 imm=0x0 word=0xd4200000` |
| `UDF #0x1234` | immediate decoded | `imm=0x1234 word=0xd4224680` |
| `RET` with lr=0 | stop at unmapped | `UNMAPPED pc=0x0 addr=0x0 size=4` |
| `B .`, max-steps 100 | step cap | `STEP-CAP steps=100` |
| no args | usage + exit code | usage printed, exit 2 |


examples/hello_arm64_range_blob.py
```diff
@@ -0,0 +1,81 @@
...
+    from qiling.const import QL_ARCH, QL_OS, QL_VERBOSE
+
+    if len(sys.argv) < 3:
+        print("usage: hello_arm64_range_blob.py <blob> <load-addr-hex> [max-steps]")
+        return 2
+    code = Path(sys.argv[1]).read_bytes()
+    load_at = int(sys.argv[2], 16)
+    max_steps = int(sys.argv[3]) if len(sys.argv) > 3 else 4096
+
+    ql = Qiling(code=code, archtype=QL_ARCH.AARCH64, ostype=QL_OS.BLOB,
```
xwings: Suggested sketch (not executed):

```
ql = Qiling(code=code, archtype=QL_ARCH.ARM64, ostype=QL_OS.BLOB,
```

i think we are using QL_ARCH.ARM64

examples/hello_arm64_range_blob.py
```diff
@@ -0,0 +1,81 @@
...
+    ql.mem.map(load_at, (len(code) + 0xFFF) & ~0xFFF)
+    ql.mem.write(load_at, code)
+    ql.mem.map(STUB_ADDR, 0x1000)
+    ql.mem.write(STUB_ADDR, b"\x00\x00\x00\x00")  # UDF landing pad
+    ql.mem.map(STACK_TOP - 0x4000, 0x4000)
+
+    events: list = []
+    state = {"steps": 0}
+
+    def hook_code(ql_inner):
```
xwings: Suggested sketch (not executed):

```
def hook_code(ql_inner, address, size):
    state["steps"] += 1
```

examples/hello_arm64_range_blob.py
```diff
@@ -0,0 +1,81 @@
...
+        state["steps"] += 1
+        if state["steps"] >= max_steps:
+            events.append({"type": "STEP-CAP", "steps": state["steps"]})
+            ql_inner.emu_stop()
+
+    def hook_unmapped(ql_inner, access, addr, size, value):
+        events.append({"type": "UNMAPPED", "pc": hex(ql_inner.arch.regs.pc),
+                       "addr": hex(addr), "size": size})
+        return False
```
xwings: Returning False from a memory hook does not stop emulation: Qiling's _hook_mem_cb ignores the callback return value and unconditionally returns True (qiling/core_hooks.py:238-252), which tells Unicorn the unmapped access was handled and lets execution continue. Impact: the docstring (line 8) and the PR description claim 'unmapped accesses stop at the loader-table boundary', but execution continues past the fault, so the example's central behavior is not delivered and the recorded UNMAPPED event is not a stop condition.

**Fix:** Stop explicitly from inside the callback (ql_inner.emu_stop(), as used for stop conditions elsewhere, e.g. qiling/extensions/idaplugin/qilingida.py's hook handlers, and in line 50 of this file) and drop or correct the false 'stop' claim if a record-only event is intended.

<details>
<summary>Source and suggested change</summary>

```
  53 |         events.append({"type": "UNMAPPED", "pc": hex(ql_inner.arch.regs.pc),
  54 |                        "addr": hex(addr), "size": size})
> 55 |         return False
  56 |
  57 |     ql.hook_code(hook_code)
```

Suggested sketch (not executed):

```
def hook_unmapped(ql_inner, access, addr, size, value):
    events.append({...})
    ql_inner.emu_stop()
```

examples/hello_arm64_range_blob.py
warsang: Changed it to  QL_ARCH.ARM64

Issue #1665, opened 2026-09-18
Make python-fx an optional dependency


toby-bro:
> **Is your feature request related to a problem? Please describe.**
> A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

I use qiling in a python project to emulate binaries. It's an awesome project thanks a lot !

Nevertheless each time I build it, it imports loads of dependencies that I don't need and don't use, especially python-fx, that imports pillow, which 
- crouds my github security advisory with vulnerabilities in pillow, 
- limits the python versions my project can build for.

> **Describe the solution you'd like**
> Send us a pull request always help, or any pseudo code might be useful.

At the moment I do a manual override to say that I don't need python-fx in my pyproject.toml, but I would really like it if qiling could implement blocks of optionnal dependencies groups (such as this other RE project angr does : 
https://github.com/angr/angr/blob/9af373689f450717ec0d0b8be0daed2477a091ee/pyproject.toml#L81-L99

So as people using qiling as a dependency can only import the core of a qiling in their projects.

Thanks a lot

PR #1666, opened 2026-09-22
Make python-fx an optional "tui" extra

toby-bro:

### Coding convention?

### Extra tests?

### Changelog?

### Target branch?

### One last thing

-----

python-fx is used only by qltui.py, for the "interactive report" entry of the post-run report menu. It was a mandatory runtime dependency, so every `pip install qiling` also pulled pyfx and its tree -- 19 packages and 25 MB, none of which the framework itself imports.

The cost is not only size. python-fx requires `pillow<11,>=10.4`, so that cap propagates to every qiling consumer: they are held on a pillow line that no longer receives fixes, and pillow 10.4.0 publishes no cp314 wheels, so `pip install qiling` on Python 3.14 falls back to building pillow from source.

Move it to a `tui` extra and import it at the point of use, mirroring the existing optional-termcolor guard (qltui.py:13). `qltool qltui` keeps working unchanged without the extra; selecting "interactive report" prints how to install it instead of failing.

test_qltui_import already owns "the installed qltui imports cleanly"; it now also asserts pyfx is not imported at module load, which is the contract this change adds. It fails before the change and passes after.

Closes #1665

Written by Claude, validated by me nevertheless


toby-bro:
I did not open a PR for moving all the other TUI dependencies in this group as pyfx was the worst.
But if you prefer, I opened #1668 to do it (which can also be merged after this one is accepted).
As you want.

Anyways, thanks for the awesome project !

(By the way the [project](https://github.com/toby-bro/microtaint) in which I use it, is part of a paper at NDSS27 if you are interested in academic conferences, (the paper is not public yet)).

